PR #65, opened 2013-04-10
Change password reset token when reset link is followed

jystewart:
This means that a password reset token will be changed within seconds of being logged. That does mean that the links in emails will only work once, so copy changes should follow to explain that.


alext:
This will break things for any users who have virus-scanners on their email system that follow links in emails to check for malware etc.

We're having issues with some users at the moment that I believe is caused by this.  Can you remember the original requirements behind this change.  Without this the token gets cleared after the password has been reset, which should cover most cases...

/cc @jystewart

PR #298, opened 2015-01-02
Replace CanCan with Pundit

vinayvinay:
https://www.agileplannerapp.com/boards/173808/cards/8934

we've found [CanCan](https://github.com/ryanb/cancan) to be limiting when we've tried to model fine-grained permissions, or introduced roles. [here's a detailed answer](https://docs.google.com/a/digital.cabinet-office.gov.uk/document/d/1CQWE68phgVVmypDST7tYnPlGb8zifYKxBLvrNP3Tseg/edit?pli=1#) to: why we're making this change?

along-with migrating from CanCan to [Pundit](https://github.com/elabs/pundit) this request includes a few other changes needed to support this migration:
- removing `EventLogsController` and merging its actions into `UsersController`: https://github.com/alphagov/signonotron2/commit/f96b78749c55d60e9974f5b186981cf0af8782b7
- removing '/admin' and '/superadmin' routing namespaces in order to apply policies in a simple manner: https://github.com/alphagov/signonotron2/commit/4dc331336a359e759530c21d5a90bb420193853f
- authorising some actions that were earlier not authorised: https://github.com/alphagov/signonotron2/commit/161aca112b26a2ff3322b31e7a19d30e8e922f5e
- introduce RSpec to leverage Pundit's [mini-DSL for testing policies](https://github.com/elabs/pundit#rspec) that works with RSpec
- writing unit tests for policies


Gemfile
```diff
@@ -53,6 +53,7 @@ end
 gem 'logstasher', '0.4.8'
 
 group :test do
+  gem 'rspec-rails', '~> 3.0.0'
```
alext: Is there a reason to use 3.0 as opposed to 3.1?

alext:
:+1:  This looks good to me.  Given the magnitude of the changes, we could do with at least one more pair of eyes over this.

This will also require thorough testing on preview, so we should hold off merging until we're confident there are no other pressing changes that need to go out etc.

Gemfile
vinayvinay: that was just an oversight, updated it.

Gemfile
```diff
@@ -22,10 +22,10 @@ gem 'devise-encryptable', '0.1.1'
 gem 'devise_security_extension', '0.7.2', git: "https://github.com/alphagov/devise_security_extension.git", branch: "graceful_return_to_behaviour"
 gem 'devise_zxcvbn', '1.1.1'
 gem 'devise-async', '0.8.0'
+gem 'pundit'
```
jamiecobbett: I'm strongly in favour of pinning gem versions to document intent and to make it much easier and clearer to track changes over time.

app/policies/user_policy.rb
```diff
@@ -0,0 +1,62 @@
+class UserPolicy < BasePolicy
+
+  def new?
+    # invitations#new
+    current_user.superadmin? || current_user.admin? || current_user.organisation_admin?
+  end
+  alias_method :index?, :new?
+
+  def edit?
+    return current_user.superadmin? if record.api_user?
```
jamiecobbett: Shouldn't this line/behaviour be duplicated in `def new?`?

app/policies/organisation_policy.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def can_assign?
+    return true if current_user.superadmin? || current_user.admin?
+    return current_user.organisation.subtree.pluck(:id).include?(record.id) if current_user.organisation_admin?
+
+    return false
+  end
+
+  class Scope < Scope
```
jamiecobbett: It isn't clear to me which `Scope` this is inheriting from. Is it `BasePolicy::Scope` or `ApplicationPolicy::Scope`? If it had an explicit namespace that would make it clearer.

jamiecobbett:
I've made a few comments but other than that I can't spot anything. This is a very large change to a complex app though.

app/policies/user_policy.rb
vinayvinay: this line was left behind after i wrote a separate `ApiUserPolicy`. i've taken it out.

app/policies/organisation_policy.rb
vinayvinay: It is `BasePolicy::Scope`. I've made it explicit.

vinayvinay:
@alext, @jamiecobbett: thanks for the feedback, i've made the changes. though we have test coverage for the authorisation logic, it is not enough. this being a cross-cutting change, we should test extensively on Preview before deploying. once we merge this, we should try and avoid Signon deploys for sometime.

PR #301, opened 2015-01-06
Admin email change flow

vinayvinay:
www.agileplannerapp.com/boards/173808/cards/6192

for users who have moved from one organisation to another, and consequently have an updated email in Signon, resetting their password is confusing.

they won't receive password reset instructions on their new email if they've not confirmed it yet. sending reset instructions to their old address works, but on most occasions they would have lost access to the old address by then.

hence, we're skipping re-confirmation when an admin updates a user's email address and instead we send a notification to a user's old and new email address informing them about the update.

also adding events to the users access log when email is changed and the change is accepted.


vinayvinay:
closing for now, need to finalise the email change flow for invited-but-not-yet-accepted users.

PR #302, opened 2015-01-06
Admin email change flow

vinayvinay:
www.agileplannerapp.com/boards/173808/cards/6192

for users who have moved from one organisation to another, and consequently have an updated email in Signon, resetting their password is confusing.

they won't receive password reset instructions on their new email if they've not confirmed it yet. sending reset instructions to their old address works, but on most occasions they would have lost access to the old address by then.

hence, we're skipping re-confirmation when an admin updates a user's email address and instead we send a notification to a user's old and new email address informing them about the update.

also adding events to the users access log when email is changed and the change is accepted.

notification email text:

> Hello Louisa Lockwood,

> A Signon administrator has changed your GOV.UK Signon account email from louisa.lockwood@example.gov.uk to louisa.lockwood@new-example.gov.uk.

> You will now need to sign in using your new email louisa.lockwood@new-example.gov.uk.

> If your email address shouldn't have changed you should contact a managing GOV.UK editor in your organisation or your parent organisation.

> All the best,

> GOV.UK team
> Government Digital Service

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -17,11 +17,17 @@ class Admin::UsersController < ApplicationController
   def update
     authorize! :read, Organisation.find(params[:user][:organisation_id]) if params[:user][:organisation_id].present?
 
-    email_before = @user.email
+    @user.skip_reconfirmation!
     if @user.update_attributes(translate_faux_signin_permission(params[:user]), as: current_user.role.to_sym)
       @user.permissions.reload
       PermissionUpdater.perform_on(@user)
-      @user.invite! if @user.invited_but_not_yet_accepted? && (email_before != @user.email)
+
+      if email_change = @user.previous_changes[:email]
+        @user.invite! if @user.invited_but_not_yet_accepted?
+        email_change.each do |to_address|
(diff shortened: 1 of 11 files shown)


alext:
One thought.  It would be useful to log the old and new email addresses in the event log.  Not sure how easy/possible that is though.

This is not a blocker for this PR.

app/views/devise/mailer/email_changed_by_admin_notification.html.erb
```diff
@@ -0,0 +1,12 @@
+<p>Hello <%= @user.name %>,</p>
+
+<p>Someone has changed your <%= link_to 'GOV.UK', 'https://www.gov.uk' %> Signon <%= account_name %> email from <%= @email_was %> to <%= @user.email %>.</p>
```
alext: "Someone" sounds somewhat suspicious here.  Would something like "A signon administrator" be better to make it clear that it's an authorised person who's made the change?

alext:
Code changes look good.  Just the one comment about the copy.

vinayvinay:
@alext: i've updated the copy to say: "A Signon administrator ...". the event log suggestion is good, i'll do that as a separate PR because it needs some changes to `EventLog`.

PR #303, opened 2015-01-06
Display and filter users by organisations

danielroseman:
Add a column for organisation in the users index, and allow filtering by org.

### DIFF
--- a/app/views/admin/users/index.html.erb
+++ b/app/views/admin/users/index.html.erb
@@ -29,11 +29,13 @@
   <tbody>
     <% @users.each do |user| %>
       <tr>
-        <td class="email" nowrap>
+        <td class="email">
           <%= user.suspended? ? "<del>".html_safe : "" %>
-            <%= link_to "#{user.name}", edit_admin_user_path(user) %>
+            <strong>
+              <%= link_to "#{user.name}", edit_admin_user_path(user) %>
+            </strong>
           <%= user.suspended? ? "</del>".html_safe : "" %>
-          <br><%= user.email %>
+          <br><span class="text-muted"><%= user.email %></span>
         </td>
         <td class="role"><%= user.role.humanize %></td>
         <td class="organisation"><%= user.organisation.try(:name) %></td>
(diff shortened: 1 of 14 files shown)


app/views/admin/users/index.html.erb
```diff
@@ -28,12 +29,14 @@
   <tbody>
     <% @users.each do |user| %>
       <tr>
-        <td class="email">
+        <td class="email" nowrap>
```
fofr: The `nowrap` attribute on a td element is obsolete. This should be achieved using CSS.

fofr:
I like this feature.

I think the organisation dropdown should show an autocomplete select2 box when opened, to avoid a long list. The items in the list should also contain the abbreviations to make searching easier, eg DEFRA, MOJ.

screen shot 2015-01-06 at 18 33 34

fofr:
The selected filter value might benefit from showing the shortened organisation name if available:

screen shot 2015-01-06 at 18 40 13

For example:

screen shot 2015-01-06 at 18 42 22

app/views/admin/users/index.html.erb
```diff
@@ -28,12 +29,14 @@
...
           <%= user.suspended? ? "<del>".html_safe : "" %>
-            <%= link_to "#{user.name} <#{user.email}>", edit_admin_user_path(user) %>
+            <%= link_to "#{user.name}", edit_admin_user_path(user) %>
           <%= user.suspended? ? "</del>".html_safe : "" %>
+          <br><%= user.email %>
```
fofr: I like these tweaks. User email could probably be set back a bit, and the name emboldened:

screen shot 2015-01-06 at 18 45 19
screen shot 2015-01-06 at 18 45 39

fofr:
Organisation admins can only see users within their organisation. 

In this case:
* it could be helpful to default to that organisation’s filter and show some form of readonly label
* or the filter should be hidden

Perhaps:
screen shot 2015-01-06 at 18 55 04

danielroseman:
All comments addressed except changing to autocomplete.

fofr:
Your changes look good to me. I am working on adding the select2 organisation drop down now.

Closing while I work on it.

fofr:
Have pushed work on a filterable list of organisation in the user filter UI. Re-opening PR.

filtering-organisations

danielroseman:
Changes look good.
Rebased and force pushed.

PR #304, opened 2015-01-07
Indicate which environment an email is from

jamiecobbett:
Include the environment name in the From field, unless it's production.
instance_name isn't set in production.

Currently the only way to distinguish which environment an email is from is by
links in the email.

I believe this is a problem for many users as they have accounts on production and
preview which they use on a daily basis.

In the same way that the app displays which environment it is running in,
emails should include the environment they came from.


vinayvinay:
@jamiecobbett: this is a much needed change. we use the `instance_name` rails config to add environment information in emails generated using [`UserMailer`](https://github.com/alphagov/signonotron2/blob/master/app/mailers/user_mailer.rb#L60-L78). it'll be good to use the same logic, to set `email_from`.

according to that logic, we don't call-out the Production environment explicitly. so if the environment is not mentioned, it is Production. which would less confusing to many of our users, who do not need to know there's an environment other than the one they're on. so what you've done in `config/initializers/email_from.rb`, will be taken care of.

fofr:
I heartily support this PR :heart:

jamiecobbett:
@vinayvinay thanks - I hadn't spotted that. I've rejigged it to share that logic between the two mailers

vinayvinay:
@jamiecobbett: looks good... you've missed updating 'from' for emails sent by devise:
https://github.com/alphagov/signonotron2/blob/master/config/initializers/devise.rb#L9

jamiecobbett:
@vinayvinay I was struggling to get a clean implementation in application config/initializers that would also set that variable, so decided to merge it with the approach in `UserMailer`.

I think that all emails we send will get an appropriate `from` because I've set the from address in [`BaseMailer`](https://github.com/alphagov/signonotron2/pull/304/files#diff-0081d9bc3de39aedc4bff23237ffa764R2)

vinayvinay:
@jamiecobbett: devise email will send an appropriate `from` if `Devise::Mailer` inherits from `BaseMailer`. thankfully, there's a config to do that: [`Devise.parent_mailer`](https://github.com/plataformatec/devise/blob/master/lib/devise.rb#L237-L241).

jamiecobbett:
@vinayvinay rebased, but it isn't beautiful.

config/initializers/devise.rb
```diff
@@ -6,7 +7,8 @@
   # ==> Mailer Configuration
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class with default "from" parameter.
-  config.mailer_sender = "GOV.UK Signon <noreply-signon@digital.cabinet-office.gov.uk>"
+  class Dummy; include MailerHelper; end # Disposable class so we can call `email_from`
+  config.mailer_sender = Dummy.new.email_from
```
vinayvinay: @jamiecobbett: we could make this: `Class.new.extend(MailerHelper).email_from` and avoid the dummy class, not a huge improvement, so feel free to decide whether you want to change it.

alext: Alternatively we could have `MailerHelper` `extend self`, which would make all its instance methods available as class methods on the module.

app/mailers/noisy_batch_invitation.rb
```diff
@@ -7,7 +9,7 @@ def make_noise(batch_invitation)
     @batch_invitation = batch_invitation
 
     user_count = batch_invitation.batch_invitation_users.count
-    subject = "[SIGNON] #{@user.name} created a batch of #{user_count} users"
+    subject = "[SIGNON] #{@user.name} created a batch of #{user_count} users in #{instance_name}"
```
alext: This subject won't be great in production where `instance_name` isn't set.

jamiecobbett: Good point - rebased.

config/initializers/devise.rb
jamiecobbett: I've gone for @vinayvinay's suggestion as it seemed slightly clearer.

PR #305, opened 2015-01-07
Allow passing options to EventLog

vinayvinay:
This makes `EventLog` flexible to addition of parameters and removes the need to pass `nil` as a parameter on occasions when trailing parameters need to be passed.

Adds a new attribute on `EventLog`: `trailing_message`, which allows us to record additional attributes about the event as a free-form string. Includes the first use of `trailing_message` in order to record old and new email address when the email is changed.

### DIFF
--- a/app/models/event_log.rb
+++ b/app/models/event_log.rb
@@ -11,7 +11,8 @@ class EventLog < ActiveRecord::Base
   UNSUCCESSFUL_LOGIN = "Unsuccessful login"
   SUSPENDED_ACCOUNT_AUTHENTICATED_LOGIN = "Unsuccessful login attempt to a suspended account, with the correct username and password"
   UNSUCCESSFUL_PASSPHRASE_CHANGE = "Unsuccessful passphrase change"
-  EMAIL_CHANGE_INITIATED = "Email changed"
+  EMAIL_CHANGED = "Email changed"
+  EMAIL_CHANGE_INITIATED = "Email change initiated"
   EMAIL_CHANGE_CONFIRMED = "Email change confirmed"
 
   # API users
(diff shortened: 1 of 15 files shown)


app/models/event_log.rb
```diff
@@ -11,7 +11,7 @@ class EventLog < ActiveRecord::Base
   UNSUCCESSFUL_LOGIN = "Unsuccessful login"
   SUSPENDED_ACCOUNT_AUTHENTICATED_LOGIN = "Unsuccessful login attempt to a suspended account, with the correct username and password"
   UNSUCCESSFUL_PASSPHRASE_CHANGE = "Unsuccessful passphrase change"
-  EMAIL_CHANGE_INITIATIED = "Email change initiated"
+  EMAIL_CHANGE_INITIATIED = "Email changed"
```
alext: I'm not convinced by this change of wording.  As I understand it, the email change does not necessarily take effect until it's confirmed, so this could be misleading.

I can see that this gets tricky because of the 2 different flows though (admin and used flows).  Maybe we need separate events for each flow - 2 for the user flow, and one for the admin flow.

app/models/event_log.rb
```diff
@@ -38,14 +39,15 @@ class EventLog < ActiveRecord::Base
   belongs_to :initiator, class_name: "User"
   belongs_to :application, class_name: "Doorkeeper::Application"
 
-  def self.record_event(user, event, initiator = nil, application = nil)
-    attributes = { uid: user.uid, event: event }
-    attributes.merge!(initiator_id: initiator.id) if initiator
-    attributes.merge!(application_id: application.id) if application
-
+  def self.record_event(user, event, options={})
```
alext: :+1: Much cleaner this way.

alext:
One comment about the messaging, otherwise looks good to me.

app/models/event_log.rb
vinayvinay: Yes, it is tricky. You're right and I've added another event for the user flow for sake of clarity:
```
Email change confirmed
Email change initiated by Vinay Patel from abc@example.com to xyz@example.com
```

PR #307, opened 2015-01-14
User email change should be notified at old email

vinayvinay:
following-up on these changes: https://github.com/alphagov/signonotron2/pull/302

When a user updates their own email address:

1. they should be sent a confirmation email at their new email address (exisiting behaviour), and
2. they should be notified about that change being initiated at their old emails address (new).

There's no corresponding story for this change, is an outcome of discussions in Slack.

/cc: @fofr, @marksheldonGDS

### DIFF
--- a/app/views/user_mailer/email_changed_notification.html.erb
+++ b/app/views/user_mailer/email_changed_notification.html.erb
@@ -1,8 +1,8 @@
 <p>Hello <%= @user.name %>,</p>
 
-<p>Your <%= link_to 'GOV.UK', 'https://www.gov.uk' %> Signon <%= account_name %> email is being changed from <%= @user.email %> to <%= @user.unconfirmed_email %>.</p>
+<p>Your <%= link_to 'GOV.UK', 'https://www.gov.uk' %> Signon <%= account_name %> email address is being changed from <%= @user.email %> to <%= @user.unconfirmed_email %>.</p>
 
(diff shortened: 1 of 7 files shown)


app/views/user_mailer/email_changed_notification.html.erb
```diff
@@ -0,0 +1,12 @@
+<p>Hello <%= @user.name %>,</p>
+
+<p>Your <%= link_to 'GOV.UK', 'https://www.gov.uk' %> Signon <%= account_name %> email is being changed from <%= @user.email %> to <%= @user.unconfirmed_email %>.</p>
+
+<p>You are required to follow the instructions sent on your new email <%= @user.email %> to confirm this change.</p>
```
alext: Should this be "...instructions sent **to** your new email..."?

alext:
Do these new emails need to include the environment info (ie preview/prod etc) like the other emails?

alext:
Other than the comments about the copy, this looks good to me.

app/views/user_mailer/email_changed_notification.html.erb
fofr: Can we get a content designer to approve the copy?

vinayvinay:
@alext, @fofr: i've updated the copy based on inputs from content editors. the environment information was present in the first line in the email body, i've included it in the subject too.

PR #311, opened 2015-01-21
Store permissions in a join table instead of a serialised attribute 

vinayvinay:
[Pundit changes](https://github.com/alphagov/signonotron2/pull/298) are tested and deployed to production, so ready for review.

#### Problems with current permissions table
- The existing permissions table stores permissions as a text field which contains permissions as a serialised array. This has made it difficult for us to write reporting scripts in the past and complicates code.
- It stores one row per application per user, which is not necessary as a user may not have access to all applications.

#### Solution: join table
We're moving this data to a join table between user and supported permissions which is the more appropriate way to model such data. This will also allow rails to clean up records for supported permissions that no longer exist and ensure that these associations don't break when supported permissions are renamed.

Making this change simplifies the way we pass permissions attributes around in Signon, which is apparent from the code changes.

#### Supplementary changes
- I've renamed the `permissions` relation on `User` to `application_permissions` to disambiguate between user's permissions derived from their role, and permissions that a user has in the context of oauth applications: https://github.com/alphagov/signonotron2/commit/e2126f392b312a05263b5ca26c13e6c68ad002fc
- Add helper methods on the user to consistently do things that were done in different ways across the app: `user#permissions_for`, `user#has_access_to?`


KushalP:
I really like this. Especially the `user.permissions_for? app_name` stuff. Good work.

db/migrate/20150109083425_create_join_table_user_application_permissions.rb
```diff
@@ -0,0 +1,14 @@
...
+    create_table :user_application_permissions do |t|
+      t.integer :user_id, null: false
+      t.integer :application_id, null: false
+      t.integer :supported_permission_id, null: false
+      t.datetime :last_synced_at
+      t.timestamps
+    end
+
+    add_index :user_application_permissions, :user_id
+    add_index :user_application_permissions, [:user_id, :application_id]
```
alext: We could do with a unique index on `[:user_id, :application_id, : supported_permission_id]` to backup the uniqueness validation.

db/migrate/20150121073933_create_join_table_batch_invitation_application_permissions.rb
```diff
@@ -0,0 +1,9 @@
+class CreateJoinTableBatchInvitationApplicationPermissions < ActiveRecord::Migration
+  def change
+    create_table :batch_invitation_application_permissions do |t|
+      t.integer :batch_invitation_id, null: false
+      t.integer :supported_permission_id, null: false
+      t.timestamps
+    end
```
alext: We could do with a unique index here as well.

db/migrate/20150109083425_create_join_table_user_application_permissions.rb
vinayvinay: @alext: good point, added.

vinayvinay:
[Pundit changes](https://github.com/alphagov/signonotron2/pull/298) are tested and deployed to production, so this PR is ready for review and merge.

db/schema.rb
```diff
@@ -136,6 +143,18 @@
...
+    t.integer  "user_id",                 :null => false
+    t.integer  "application_id",          :null => false
+    t.integer  "supported_permission_id", :null => false
+    t.datetime "last_synced_at"
+    t.datetime "created_at",              :null => false
+    t.datetime "updated_at",              :null => false
+  end
+
+  add_index "user_application_permissions", ["user_id", "application_id"], :name => "index_user_application_permissions_on_user_id_and_application_id"
+  add_index "user_application_permissions", ["user_id"], :name => "index_user_application_permissions_on_user_id"
```
alext: These don't seem to match the indexes created in the migrations.  It's probably worth running the migrations against a freshly loaded database to ensure these are consistent.

Also, see #319 which may be related.

app/models/enhancements/application.rb
```diff
@@ -9,9 +9,9 @@ class ::Doorkeeper::Application < ActiveRecord::Base
 
   default_scope order('oauth_applications.name')
   scope :support_push_updates, where(supports_push_updates: true)
-  scope :can_signin, lambda { |user| joins(:permissions)
-                                      .where(permissions: { user_id: user.id })
-                                      .where('permissions.permissions LIKE ?', '%signin%') }
```
alext: Good to see this query gone.

db/migrate/20150109083425_create_join_table_user_application_permissions.rb
```diff
@@ -0,0 +1,15 @@
...
+    add_index :user_application_permissions, [:user_id, :application_id, :supported_permission_id], unique: true
```
alext: There's no need to have both this and the previous indexes.  Queries using the `user_id` and `application_id`, or just the `user_id` will be covered by this index.  Suggest we remove the others.

app/views/api_users/index.html.erb
```diff
@@ -30,7 +30,7 @@
       </tr>
     </thead>
     <tbody>
-      <% @api_users.each do |user| %>
+      <% ApiUser.all.each do |user| %>
```
alext: Why this change?  The controller is still populating `@api_users`, and using an `include` to make it more efficient...

alext:
This looks really good.  A few minor comments, otherwise good work.

elliotcm:
Hi @vinayvinay 

Looks like you've already fixed some of the issues in this one: https://github.com/alphagov/signonotron2/pull/319

Probably easier for you to sort the last bits manually since for some unknown reason the tests are failing on my branch.

vinayvinay:
@elliotcm: i've cherry-picked your commits, and worked-around the test failures by ignoring order of elements in the permissions array, because they don't matter: https://github.com/alphagov/signonotron2/commit/b5cdbbee60af4152fbaf02b01894a302fe95092e.

vinayvinay:
@alext: i've addressed your comments. thanks!

PR #312, opened 2015-01-21
Allow adding a note to a user when editing.

elliotcm:
DO NOT MERGE BEFORE https://github.com/alphagov/signonotron2/pull/298

Allows tracking of training, reasons for granting
or revoking specific permissions etc.

Requested by GOV.UK content team.

- [x] Data input
- [ ] Data output

screenshot 2015-01-21 13 05 55

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -60,6 +60,10 @@ class UsersController < ApplicationController
           end
         end
 
+        if params[:note] && (params[:note][:details].present? || params[:note][:occurred_on].present?)
+          @user.notes << Note.new(params[:note])
+        end
+
         redirect_to users_path, notice: "Updated user #{@user.email} successfully"
       else
         render :edit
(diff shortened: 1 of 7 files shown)


rboulton:
Seems good, but tests failed.

alext:
Also, can we hold off merging while #298 is being tested on preview.

fofr:
I'd love to see some screenshots too.

vinayvinay:
0. worth considering whether this information should be an entry in the user's event log, instead of storing it as a `note`.
0. how are we thinking of sharing these notes with the GOV.UK content team when they want it? i'm guessing it wouldn't be very costly to surface this on a page that's linked from the user's edit page.

elliotcm:
@rboulton Thanks, I'll sort them out.

@alext No problem.  Added a warning.

@fofr I've added a screenshot of the entry form, I'll add another of the display once it's done.

@vinayvinay 
1. Worth investigating.  I'll take a look now.
2. I was planning on doing that as another PR but it makes as much sense as part of this one.  I'll set it up.

elliotcm:
The scope of this has been changed last minute.  I'll do a new PR once it's all changed.

PR #314, opened 2015-01-22
Make the user's name a link to edit themselves

fatbusinessman:
signon-quick-link

It's quite a common use case, at least for GOV.UK developers, to want to change their own permissions to test some permission-based feature in an app, which involves going to the user section and searching for their user record. If they have to do this multiple times, it's tedious and introduces the risk that they'll change permissions for someone else by mistake.

The original destination for this link was to the Signon root page: this is consistent with other apps, where it makes sense to link through to Signon; when you're in Signon already, this isn't particularly useful, especially when the "GOV.UK Signon" link in the header takes you there already.

The logic to determine the link target depends on whether the user has permission to edit themselves. As the logic currently stands, every user in the system has permission to edit themselves to at least some extent, but we shouldn't assume this will always be the case.

### DIFF
--- a/test/integration/user_link_test.rb
+++ b/test/integration/user_link_test.rb
@@ -1,0 +1,16 @@
+require_relative "../test_helper"
+
+class UserLinkTest < ActionDispatch::IntegrationTest
+  context "logged in as an admin" do
+    setup do
+      @admin = create(:admin_user, :name => "Adam Adminson", :email => "admin@example.com")
+      visit new_user_session_path
+      signin(@admin)
+    end
+
+    should "link to the current user's edit page" do
+      click_on "Adam Adminson"
+      assert page.has_content?('Edit “Adam Adminson”')
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


fatbusinessman:
@fofr: I suspect you may have Opinions on this.

test/integration/user_link_test.rb
```diff
@@ -0,0 +1,16 @@
...
+  context "logged in as an admin" do
+    setup do
+      @admin = create(:admin_user, :name => "Adam Adminson", :email => "admin@example.com")
+      visit new_user_session_path
+      signin(@admin)
+    end
+
+    should "link to the current user's edit page" do
+      click_on "Adam Adminson"
+      assert page.has_content?('Edit “Adam Adminson”')
```
jamiecobbett: :+1:

fatbusinessman: Any reference to @adammaddison is purely coincidental.

fofr:
I like this change very much :+1:

alext:
Fun fact:  On preview, for a non-admin user, `/users/<n>/edit` will work for any value of n, and let you edit yourself.  This may be a bug from the pundit changes. /cc @vinayvinay

jennyd:
This makes me happy; I have often wanted this :+1:

PR #320, opened 2015-01-30
Add change notes to users

elliotcm:
A replacement for https://github.com/alphagov/signonotron2/pull/312, this provides two pieces of functionality:

1. The ability to (optionally) add custom text when editing a user's permissions, sitting in the existing `trailing_message` field.
2. Automatic tracking of permissions changes and the person who changed them, with a diff stored in a new `data` field.

Both are rendered in the event log.  This does not change the behaviour of any existing events.

Screenshots:

screenshot 2015-01-30 14 55 08

screenshot 2015-01-30 14 54 52


elliotcm:
Tests are failing due to a problem in master, will rebase when possible.

@fofr I imagine you'll want input on this one.

fofr:
This looks great.

The only drawback I can see is that form edits presume that the changes are related to permissions. They could be changing a name or an email. Which shows like this:

screen shot 2015-02-02 at 13 20 01

Perhaps "Permission changed" should be "User edited"? And if there are no permissions added or removed the More info column should be empty.

fofr:
When changing email this leads to two entries in the log:

screen shot 2015-02-02 at 13 23 51

elliotcm:
@fofr I'll update it so it only creates the changed permissions log if permissions have actually changed.

elliotcm:
@fofr Issue resolved:
screenshot 2015-02-04 11 22 57

fofr:
UI looks good to me. :+1:

test/integration/granting_permissions_test.rb
```diff
@@ -28,4 +28,48 @@ class GrantingPermissionsTest < ActionDispatch::IntegrationTest
...
+
+    within ".new-note" do
+      fill_in "Add a change note (optional)", with: "Completed author training"
+    end
+
+    click_button "Update User"
+
+    last_log = @user.event_logs.last
+    assert_equal "Completed author training", last_log.trailing_message
+  end
```
vinayvinay: should we make this test more comprehensive by asserting for presence of event log data on the event logs page instead of asserting on `event_logs.last`?

lib/permission_differ.rb
```diff
@@ -0,0 +1,41 @@
...
+      app_diff = perms_1[app_name] - perms_2[app_name]
+
+      if app_diff.any?
+        diff.merge(app_name => app_diff)
+      else
+        diff
+      end
+    end
+  end
+end
```
vinayvinay: some unit tests for this differ would be good, else this logic remains completely untested.

vinayvinay:
a couple of improvements suggested for tests, rest looks good.

jamiecobbett:
This has been inactive for 5 months - I'm inclined to close it as it can't be merged. What do you think @elliotcm?

elliotcm:
I agree with Vinay's comments and they're not too hard to implement, it might still be worth keeping.

elliotcm:
That said, I'll likely never get to it so it's up to whichever team would have to pick it up as to whether there's value to them.

marksheldonGDS:
Logging and saving change notes when a user's details, application access or application permissions are edited should be considered as part of the planned improvement work to Signon.

PR #322, opened 2015-02-02
Upgrade doorkeeper

vinayvinay:
there have been many enhancements and performance improvements to this gem:
https://github.com/doorkeeper-gem/doorkeeper/blob/master/CHANGELOG.md

many changes are to make it compatible with newer versions of Rails, so these are first steps to prepare for upgrading Signon to Rails 4.

some noteworthy backward incompatible changes:
- removing `doorkeeper_for` in favour of a `before_filter`
- forcing usage of SSL in redirect uri

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,7 @@ gem 'devise_zxcvbn', '1.1.1'
 gem 'devise-async', '0.8.0'
 gem 'pundit', '0.3.0'
 
-gem 'doorkeeper', '0.6.7'
+gem 'doorkeeper', '2.1.0'
 gem 'ancestry', '2.0.0'
 
 gem 'gds-api-adapters', '7.11.0'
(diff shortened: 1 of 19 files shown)


db/migrate/20150122082200_add_scopes_to_oauth_applications.rb
```diff
@@ -0,0 +1,5 @@
+class AddScopesToOauthApplications < ActiveRecord::Migration
```
danielroseman: I'm not sure I understand why we need this. According to your link, this is only relevant if we add `enable_application_scopes`, which we're not doing - and I don't see where we are actually using application-level scopes anyway. What am I missing?

vinayvinay: @danielroseman - you're right. i've removed that migration. in the CHANGELOG.md, they've called it a backwards incompatible change, so i was lead to believe that it is needed unconditionally.

PR #324, opened 2015-02-04
Create user update supported permissions

vinayvinay:
users sending push updates to applications are [given 'user_update_permission' permission](https://github.com/alphagov/signonotron2/blob/master/lib/sso_push_credential.rb#L14) while generating
their access token.

gds-sso [checks for presence of this permission](https://github.com/alphagov/gds-sso/blob/6c37cfe4e8c74b9d530bac5ea9c7cb73b768c193/app/controllers/api/user_controller.rb#L44-L46) while authorising user API access.

earlier this worked because permissions were stored as strings, and there wasn't a 1-1 correspondence between granted permissions and an application's supported permission. now with permissions being stored as association records in a join table, every application that supports push updates must have a supported permission by the name 'user_update_permission'.

fixes: [errbit](https://errbit.preview.alphagov.co.uk/apps/52e678360da1158195000002/problems/54cbb4bb0da115e62c000d27)

### DIFF
--- a/db/migrate/20150204065936_add_user_update_supported_permission_to_applications.rb
+++ b/db/migrate/20150204065936_add_user_update_supported_permission_to_applications.rb
@@ -1,0 +1,9 @@
+require 'enhancements/application'
+
+class AddUserUpdateSupportedPermissionToApplications < ActiveRecord::Migration
+  def change
+    Doorkeeper::Application.where(supports_push_updates: true).each do |application|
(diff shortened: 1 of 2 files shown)


alext:
I worry that surfacing this permission in the Admin UI will lead to confusion.  Having said that, I'm not sure what the correct solution for this is.  

I'm thinking that we may need to model a global set of supported permissions to cover this sort of case.

vinayvinay:
superseded by https://github.com/alphagov/signonotron2/pull/326

PR #326, opened 2015-02-04
Supported permissions can be hidden from admin UI

vinayvinay:
we need a way to hide supported permissions from the admin ui, which should only be granted through code. as of now the only such supported permission is 'user_update_permission', which is granted to users accessing the user API in gds-sso: https://github.com/alphagov/signonotron2/blob/master/lib/sso_push_credential.rb

gds-sso checks for presence of this permission while authorising user API access: https://github.com/alphagov/gds-sso/blob/6c37cfe4e8c74b9d530bac5ea9c7cb73b768c193/app/controllers/api/user_controller.rb#L44-L46

we need to add a 'user_update_permission' supported permission to apps that support push updates. earlier this worked because permissions were stored as strings, and there wasn't a 1-1 correspondence between granted permissions and an application's supported permission. now with permissions being stored as association records in a join table, every application that supports push updates must have a supported permission by the name 'user_update_permission'.

fixes: [errbit](https://errbit.preview.alphagov.co.uk/apps/52e678360da1158195000002/problems/54cbb4bb0da115e62c000d27)

### DIFF
--- a/app/models/enhancements/application.rb
+++ b/app/models/enhancements/application.rb
@@ -17,6 +17,7 @@ class ::Doorkeeper::Application < ActiveRecord::Base
                                   .where(supported_permissions: { name: 'signin', delegatable: true })
 
   after_create :create_signin_supported_permission
+  after_save :create_user_update_supported_permission
 
   def self.policy_class; ApplicationPolicy; end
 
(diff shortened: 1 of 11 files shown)


alext:
The code changes look good to me.  The only question I have is around the naming of the new property.  `grantable_by_admin` overlaps with the role names a little, and would raise the question on whether a superadmin can grant that permission...

I'm not finding it easy to come up with a better name though.  Maybe `grantable_from_ui`, or we could reverse it and name it something like `system_internal_permission`.

alext:
Another detail.  We need an `after_create` on applications so that this new permission gets created for any future applications.

vinayvinay:
@alext: i've renamed to `grantable_from_ui`. i find this name to be more explicit about the intended behaviour than `system_internal_permission`. also added an `after_save` to create `user_update_permission` if it doesn't exist, because we need this permission for applications that support push updates.

PR #329, opened 2015-02-17
Bugfix - viewing user with unsynced permissions

heathd:
In some circumstances a crash could occur when viewing a user with unsynced
permissions. The precise scenario in which this could occur is if the first
permission in the users permission list was unsynced, and later ones were
synced.

In theory this should never occur because all permissions are synced at
the same time. However the association from user to application_permissions is
unordered, so the order in which the records are returned from mysql when
reading them could change arbitrarily.

The fix here is to make the display code more robust to this possible scenario.

I've also refactored the view slightly so that it's clear that only the
`<span>` label class changes according to the `sync_needed?` status.

https://www.pivotaltracker.com/story/show/88500338

### DIFF
--- a/app/views/shared/_user_permissions.html.erb
+++ b/app/views/shared/_user_permissions.html.erb
@@ -52,12 +52,11 @@
         </td>
         <% if user_object.persisted? %>
           <td>
-            <% if permissions.map(&:last_synced_at).compact.present? %>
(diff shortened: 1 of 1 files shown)


heathd:
@vinayvinay, when you have a moment, could you take a look please?

PR #334, opened 2015-04-07
Remove sqlite3 development dependency

### LINKED ISSUE #333: Remove sqlite3 development dependency
I see that the sqlite gem is required in the development gems, given that mysql is used what are the thoughts on removing this extra dependency?

matthewford:
Fixes #333

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,7 +42,6 @@ group :assets do
 end
 
 group :development do
-  gem 'sqlite3'
   gem 'quiet_assets'
 end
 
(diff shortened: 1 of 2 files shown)


jennyd:
Thanks for cleaning this up.

alext:
The gds-sso test suite needs this to include sqlite3.  It spins up a copy of signonotron2 using sqlite3 to do a full-end-to-end test.  The tests have been failing since this was merged: https://ci-new.alphagov.co.uk/job/govuk_gds_sso/620/console...

jennyd:
Thanks for spotting that, @alext - fixed in https://github.com/alphagov/signonotron2/pull/339 (although the build is failing for other reasons now :cry: ).

jamiecobbett:
We could use the GitHub multi-build status to show/run both the gds-sso build and the signonotron build for signon PRs - that would have caught this problem.

PR #336, opened 2015-04-08
Introduce strong parameters

jennyd:
We need to [upgrade Devise to 3.2](https://www.pivotaltracker.com/story/show/91090666) and Devise Invitable to 1.4, which assumes that we're using Rails 4, or at least strong params. This PR is a preliminary step to doing the full Devise upgrade. It also helps to get us nearer to Rails 4, which we should upgrade to soon in any case.

Lots of these changes relate to specific behaviour of Devise and its extensions - look at individual commit messages for more explanation of those.

We've relied on existing test coverage extensively for these changes - the coverage [looks pretty good](https://ci-new.alphagov.co.uk/job/govuk_signonotron2_branches/ws/coverage/rcov/index.html) so we're fairly confident that we haven't broken anything, but should keep this on Preview for a few days before deploying to Production to be sure.

We'd appreciate a thorough review of these changes: both @mikejustdoit and I are new to Devise and (most of) Signon so it's taken us lots of archaeology to get this far.

### DIFF
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -29,7 +29,6 @@ class InvitationsController < Devise::InvitationsController
       respond_with resource, :location => after_invite_path_for(resource)
     else
       # workaround for invitatable not providing a build_invitation which could be authorised before saving
-      resource_params = params[resource_name]
       user = User.new(resource_params)
       user.organisation_id = resource_params[:organisation_id]
       authorize user
@@ -59,4 +58,37 @@ class InvitationsController < Devise::InvitationsController
     users_path
   end
 
+  def resource_params
+    UserParameterSanitiser.new(
+      user_params: unsanitised_user_params,
+      current_user_role: current_user_role,
+    ).sanitise
+  end
+
+  # TODO: once we've upgraded Devise and DeviseInvitable, `resource_params`
+  # hopefully won't be being called for actions like `#new` anymore and we
+  # can change the following `params.fetch(:user)` to
+  # `params.require(:user)`. See
+  # https://github.com/scambra/devise_invitable/blob/v1.1.5/app/controllers/devise/invitations_controller.rb#L10
+  # and
+  # https://github.com/plataformatec/devise/blob/v2.2/app/controllers/devise_controller.rb#L99
+  # for details :)
+  def unsanitised_user_params
+    params.fetch(:user, {})
+  end
+
+  # NOTE: `current_user` doesn't exist for `#edit` and `#update` actions as
+  # implemented in our current (out-of-date) versions of Devise
+  # (https://github.com/plataformatec/devise/blob/v2.2/app/controllers/devise_controller.rb#L117)
+  # and DeviseInvitable
+  # (https://github.com/scambra/devise_invitable/blob/v1.1.5/app/controllers/devise/invitations_controller.rb#L5)
+  #
+  # With the old attr_accessible approach, this would fall back to the
+  # default whitelist (i.e. equivalent to the `:normal` role) and this
+  # this preserves that behaviour. In fact, a user accepting an invitation
+  # only needs to modify `password` and `password_confirmation` so we could
+  # only permit those two params for the `edit` and `update` actions.
+  def current_user_role
+    current_user.try(:role).try(:to_sym) || :normal
+  end
 end
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -59,10 +59,16 @@ class InvitationsController < Devise::InvitationsController
   end
 
   def resource_params
-    UserParameterSanitiser.new(
+    sanitised_params = UserParameterSanitiser.new(
       user_params: unsanitised_user_params,
       current_user_role: current_user_role,
     ).sanitise
+
+    if params[:action] == "update"
+      sanitised_params.merge(invitation_token: invitation_token)
+    else
+      sanitised_params
+    end
   end
 
   # TODO: once we've upgraded Devise and DeviseInvitable, `resource_params`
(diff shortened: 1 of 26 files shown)


test/unit/batch_invitation_user_test.rb
```diff
@@ -10,12 +10,15 @@ class BatchInvitationUserTest < ActiveSupport::TestCase
 
     should "store invitation attributes against a user" do
       user = create(:batch_invitation_user, batch_invitation: @batch_invitation)
-      invitation_attributes = {
+
+      # the attributes that're passed to User#invite! should be a permitted
+      # params object
+      invitation_attributes = ActionController::Parameters.new({
```
boffbowsh: Tiny pedantic comment: you don't need the braces here any more.

test/unit/lib/user_parameter_sanitiser_test.rb
```diff
@@ -0,0 +1,84 @@
...
+      user_params = { name: "Mary", birthday: "today!" }
+
+      @sanitised_params = UserParameterSanitiser.new(
+        user_params: user_params,
+        current_user_role: :normal,
+        permitted_params_by_role: permitted_params_by_role,
+      ).sanitise
+    end
+
+    should "remove unpermitted params" do
```
boffbowsh: Should you also check `permitted?` is false here?

test/unit/lib/user_parameter_sanitiser_test.rb
```diff
@@ -0,0 +1,84 @@
...
+      @sanitised_params = UserParameterSanitiser.new(
+        user_params: user_params,
+        current_user_role: :normal,
+        permitted_params_by_role: permitted_params_by_role,
+      ).sanitise
+    end
+
+    should "return permitted params object" do
+      assert_instance_of ActionController::Parameters, @sanitised_params
+      assert @sanitised_params.permitted?
```
boffbowsh: Worth checking the permitted params haven't been lost? I know it's a thin wrapper around `ActionController::Parameters` but might be worth checking

test/unit/lib/user_parameter_sanitiser_test.rb
mikejustdoit: I think `permitted?` would still be true, since `@sanitised_params` has been `#permit`ed. If there's any value in testing that we could?

app/models/enhancements/application.rb
```diff
@@ -1,12 +1,26 @@
...
 class ::Doorkeeper::Application < ActiveRecord::Base
+  include ActiveModel::ForbiddenAttributesProtection
+
   has_many :permissions, :dependent => :destroy
   has_many :supported_permissions, :dependent => :destroy
 
-
-  attr_accessible :name, :description, :uid, :secret, :redirect_uri, :home_uri
-  attr_accessible :supports_push_updates, role: :superadmin
+  # TODO: Remove this accessible attributes list once Signonotron has upgraded
```
evilstreak: We worked around a similar problem in the strong parameters upgrade on Whitehall by making `attr_accessible` a no-op: https://github.com/alphagov/whitehall/commit/a13a719118048c3fb296706a857846847b47ad07

I think using a similar approach here would let us drop the `attr_accessible` call.

app/models/user.rb
```diff
@@ -3,6 +3,10 @@
 class User < ActiveRecord::Base
   include Roles # adds validations and accessible attributes
```
evilstreak: Is this comment still accurate? Does it add accessible attributes any more?

app/controllers/invitations_controller.rb
```diff
@@ -59,4 +58,37 @@ def after_invite_path_for(resource)
...
+    sanitised_params = UserParameterSanitiser.new(
+      # TODO: once we've upgraded Devise and DeviseInvitable, `resource_params`
+      # hopefully won't be being called for actions like `#new` anymore and we
+      # can change the following `params.fetch(:user)` to
+      # `params.require(:user)`. See
+      # https://github.com/scambra/devise_invitable/blob/v1.1.5/app/controllers/devise/invitations_controller.rb#L10
+      # and
+      # https://github.com/plataformatec/devise/blob/v2.2/app/controllers/devise_controller.rb#L99
+      # for details :)
+      user_params: params.fetch(:user, {}),
```
evilstreak: Given that the two params passed to `UserParameterSanitiser` each need substantial comments explaining the behaviour, I think it would be worth extracting them into private methods and moving the comments to those methods.

lib/user_parameter_sanitiser.rb
```diff
@@ -0,0 +1,44 @@
+class UserParameterSanitiser
+  def initialize(user_params:, current_user_role:, permitted_params_by_role: nil)
+    @user_params = user_params
+    @current_user_role = current_user_role
+    @permitted_params_by_role = permitted_params_by_role ||
+      default_permitted_params_by_role
```
evilstreak: Rather than default the keyword parameter to `nil` and use `||` to set a fallback here, what about adding it to the params?

```ruby
def initialize(user_params:, current_user_role:, permitted_params_by_role: default_permitted_params_by_role)
```

evilstreak:
:+1: Good work

app/models/enhancements/application.rb
mikejustdoit: Thanks for the tip @evilstreak. I've discussed this with Dom, but for anybody else reading: Unfortunately after trying the approach taken for Whitehall, it turns out this particular case is different in that the offending `attr_accessible` call originates from a module included in Doorkeeper's Application model, which our own model seems to be a *reopening* of and therefore makes it "pretty gross" to solve

test/unit/lib/user_parameter_sanitiser_test.rb
mikejustdoit: Thanks @boffbowsh, good point. I've just added a couple of `should "permit the permitted params"`s

vinayvinay:
this is good stuff! :cake: 

only suggestion i have is to move the `UserParameterSanitiser` usages out from the controller, into `Policy` classes, and use the [`permitted_attributes` Pundit helper](https://github.com/elabs/pundit#strong-parameters) instead. this will be a cosmetic change, and i would prefer to do that as a separate PR sometime later when most of the `FIXME`s and `TODO`s are taken care of.

PR #341, opened 2015-04-21
Signon correctly handles organisation slug changes

jamiecobbett:
This reduces the load on 2nd line when an organisation's slug changes. It also enables us to send content_ids rather than slugs when telling apps about users, thus reducing their knowledge of/dependency on organisation slugs.


alext:
This change looks good to me, however, I don't think we can deploy it all in one go.  I think we first need to deploy a change to give organisations a content_id field, and populate that field from the organisations API.  Then once they're populated, we can deploy a second change to to key the update on the content_id.

jamiecobbett:
You're right, I'll close and split it into two Pull Requests.

PR #343, opened 2015-04-21
Automatically handle organisation slug changes

jamiecobbett:
~~DO NOT MERGE until https://github.com/alphagov/signonotron2/pull/342 has been deployed and the field has been populated.~~


test/factories/organisation.rb
```diff
@@ -2,6 +2,7 @@
   factory :organisation do
     sequence(:slug) { |n| "ministry-of-funk-#{n}000" }
     sequence(:name) { |n| "Ministry of Funk #{n}000" }
+    sequence(:content_id) { SecureRandom.uuid }
```
alext: This doesn't need to be a sequence.  It can just be:
```ruby
content_id { SecureRandom.uuid }
```

test/unit/organisations_fetcher_test.rb
```diff
@@ -20,10 +20,14 @@ class OrganisationsFetcherTest < ActiveSupport::TestCase
       :organisation,
       name: 'Ministry Of Misery',
       slug: slug,
+      content_id: "a1b2c3",
```
alext: I think it would be good to use a well-formed content-id here just to avoid any confusion.

test/factories/organisation.rb
jamiecobbett: Rebased

test/unit/organisations_fetcher_test.rb
jamiecobbett: Fair enough - rebased.

jamiecobbett:
One wrinkle here is that there are 51 orphaned organisations in production - orphaned as a result of slug changes or of being deleted from Whitehall. I'm not entirely sure what to do about these.

alext:
Is that on preview or production? (signon isn't synced across...)

Can we write a migration to clean them up?  It might be a case of dumping the details into a spreadsheet, and doing some analysis to figure out what to do with each one.  Not a fun task...

jamiecobbett:
51 in production and 84 in preview - I think the extra ones in preview are from nonsense organisations being created in Whitehall.

jamiecobbett:
@alext I've addressed the problem in 5d0455f

alext:
:+1: looks good to me.  Merge when green...

PR #346, opened 2015-05-01
Display when organisations have closed

jamiecobbett:
This makes it clearer when choosing organisations to assign to a user which
ones are plausible. An example of this is Highways Agency which has become a
different organisation called Highways England.

This could happen more when the new government is formed and organisations are
closed and created with potentially similar/confusing names.

screen shot 2015-05-01 at 11 26 14

### DIFF
--- a/test/unit/organisations_fetcher_test.rb
+++ b/test/unit/organisations_fetcher_test.rb
@@ -20,6 +20,7 @@ class OrganisationsFetcherTest < ActiveSupport::TestCase
       :organisation,
       name: 'Ministry Of Misery',
       slug: slug,
+      closed: true,
     )
     assert_equal(1, Organisation.count)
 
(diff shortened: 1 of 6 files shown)


test/unit/organisations_fetcher_test.rb
```diff
@@ -70,6 +73,8 @@ class OrganisationsFetcherTest < ActiveSupport::TestCase
     assert_equal [movies], fun.children
   end
 
+
+
```
jennyd: Did you mean to add this extra whitespace here?

jamiecobbett: :space_invader: :space_invader: rebased.

jennyd:
There are some [tests for `name_with_abbreviation`](https://github.com/alphagov/signonotron2/blob/5e93bec9186fa0265076aaa84dcd4f905882f7a4/test/unit/organisation_test.rb#L23-L38) which you could extend to cover this change.

jamiecobbett:
@jennyd rebased.

jennyd:
:sparkles: :door:

PR #348, opened 2015-05-05
Upgrade Devise to 3.0.4

jennyd:
This is a small step on the way to recent versions of all the things. Devise 3.0 supports Rails 4, so this will enable us to do that upgrade soon.


benilovj:
The changes made in this PR look reasonable and from [Devise's CHANGELOG](https://github.com/plataformatec/devise/blob/master/CHANGELOG.md), it doesn't look like anything else needs changing but I feel confident enough on what's different between devise 2.2.8 and 3.0.4 to merge this.

app/controllers/invitations_controller.rb
```diff
@@ -58,6 +58,12 @@ def after_invite_path_for(resource)
     users_path
   end
 
+  # TODO: remove this method when we're on a version of devise_invitable which
```
jamiecobbett: I see that we can't safely upgrade to it yet, but can we say which version drops the requirement?

jamiecobbett:
One minor comment, otherwise :+1:

app/controllers/invitations_controller.rb
jennyd: Done

PR #349, opened 2015-05-12
Upgrade Devise to v3.1.0

bishboria:
Another installment of the Upgrade Devise Saga.

Devise v3.1.0 is compatible with both rails 3 and 4. This seemed like the simplest devise version to change before requiring rails 4 upgrade.


evilstreak:
:+1:

config/initializers/devise.rb
```diff
@@ -307,4 +307,7 @@
   # Time period for account expiry from last_activity_at
   # config.expire_after = 90.days
 
+  config.secret_key = '9c46442808d803d907849ad47dc929851e1603686b8be5913347884cfe4d6f108ca3edeb8749c22430f215c1c90a0d1b456c103c5e30317027459087f24c7240'
```
boffbowsh: What's the source of this? Will it need overriding in production?

db/migrate/20150507160746_remove_limit_from_invitation_token.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveLimitFromInvitationToken < ActiveRecord::Migration
+  def up
+    change_column :users, :invitation_token, :string, :limit => nil
```
boffbowsh: This will change the column to a `VARCHAR(255)`, iirc. Is this correct?

config/initializers/devise.rb
bishboria: Copy/Paste/Forget error. Fixed here: https://github.com/alphagov/signonotron2/pull/350

PR #350, opened 2015-05-14
Store devise secret_key in an environment variable

bishboria:
Part of upgrading devise.

Previous `secret_key` was auto-generated. Load the key from an environment variable instead.


bradleywright:
This will need a PR in Puppet to add the environment variable. Let's do that first before we merge this.

bishboria:
We're working on that right now

boffbowsh:
:+1:. I thought about suggesting a default for if that env var isn't set, but I guess it's better for it to fail loudly.

bradleywright:
I think we can merge this now?

PR #338, opened 2015-04-09
Upgrade to Ruby 2.2 

eritiro:
Upgrade to ruby 2.2.1

Gems updated:
mysql2 v0.3.11 to v0.3.18
json v1.7.7 to v1.8.2
unicorn  v4.3.1 to v4.8.3

PR #352, opened 2015-06-01
Update signon to Rails 4

bishboria:
Part of the Upgrade Devise Saga

Devise 3.1 is the last version with both Rails 3 and 4 compatibility, so moving signon to rails 4 first before trying Devise 3.2


Gemfile.lock
```diff
@@ -70,10 +72,13 @@ GEM
...
+    coderay (1.1.0)
+    columnize (0.9.0)
     connection_pool (1.2.0)
     crack (0.4.2)
       safe_yaml (~> 1.0.0)
-    database_cleaner (0.7.2)
+    database_cleaner (1.4.1)
+    debugger-linecache (1.2.0)
     devise (3.1.0)
       bcrypt-ruby (~> 3.0)
```
jackscotti: When bundling the repo I get the following warning:
```
The bcrypt-ruby gem has changed its name to just bcrypt.  Instead of
installing `bcrypt-ruby`, you should install `bcrypt`.  Please update your
dependencies accordingly.
```

app/controllers/sessions_controller.rb
```diff
@@ -18,7 +18,8 @@ class SessionsController < Devise::SessionsController
       timeout_callback.present?,
       timeout_callback.kind == :before,
       timeout_callback.raw_filter.is_a?(Proc),
-      timeout_callback.per_key == {:if=>[], :unless=>[]},
+      # FIXME: per_key doesn't work any more, find an alternative
```
evilstreak: Does this still need fixing?

app/mailers/user_mailer.rb
```diff
@@ -1,7 +1,7 @@
 class UserMailer < ActionMailer::Base
   include MailerHelper
 
-  default from: lambda { email_from }
+  default from: email_from()
```
evilstreak: Should this be a `Proc`, as in [noisy_batch_invitation.rb](https://github.com/alphagov/signonotron2/pull/352/files#diff-e0c9d29652611247039064bbd37e2c6dR4)?

app/models/batch_invitation_user.rb
```diff
@@ -5,9 +5,9 @@ class BatchInvitationUser < ActiveRecord::Base
 
   validates :outcome, inclusion: { :in => [nil, "success", "failed", "skipped"] }
 
-  scope :processed, where("outcome IS NOT NULL")
-  scope :unprocessed, where("outcome IS NULL")
-  scope :failed, where(outcome: "failed")
+  scope :processed, -> { where("outcome IS NOT NULL") }
+  scope :unprocessed, -> { where("outcome IS NULL") }
```
evilstreak: Is it worth updating these to the new syntax now, `where.not(outcome: nil)` and `where(outcome: nil)`?

app/policies/user_policy.rb
```diff
@@ -7,6 +7,7 @@ def new?
   alias_method :index?, :new?
 
   def edit?
+    return unless current_user
```
evilstreak: In what situation was `current_user` not set here? Is it worth addressing the root problem there?

config/application.rb
```diff
@@ -19,19 +12,9 @@ class Application < Rails::Application
...
-    # Only load the plugins named here, in the order given (default is alphabetical).
-    # :all can be used as a placeholder for all plugins not explicitly named.
-    # config.plugins = [ :exception_notification, :ssl_requirement, :all ]
-
-    # Activate observers that should always be running.
-    # config.active_record.observers = :cacher, :garbage_collector, :forum_observer
-
     # Set Time.zone default to the specified zone and make Active Record auto-convert to this zone.
     # Run "rake -D time" for a list of tasks for finding time zone names. Default is UTC.
-    config.time_zone = 'London'
```
evilstreak: Do we really want to drop this?

app/controllers/sessions_controller.rb
bishboria: the entire patch for < 3.2 is completely gone in the devise 3.2.1 branch

Gemfile.lock
bishboria: bcrypt is a devise dependency, we don't list it in the Gemfile or use it directly in the app…

test/models/batch_invitation_test.rb
```diff
@@ -5,7 +5,10 @@ class BatchInvitationTest < ActiveSupport::TestCase
     ActionMailer::Base.deliveries = []
 
     @app = create(:application)
-    @bi = create(:batch_invitation, supported_permission_ids: [@app.signin_permission.id])
+    @bi = create(:batch_invitation)
```
evilstreak: Does this work as `@bi = create(:batch_invitation, supported_permissions: [@app.signin_permission])`?

app/controllers/sessions_controller.rb
evilstreak: I thought we were still only on devise 3.1? If this isn't something we need to do any more, we should remove the `FIXME` comment.

app/policies/user_policy.rb
bishboria: looks like it a false alarm while changing other stuff.

test/models/batch_invitation_test.rb
bishboria: it does!

boffbowsh:
All looks ok to me

boffbowsh:
Should probably upgrade Ruby as part of this, some work has been done in #338

boffbowsh:
Stylistically, there's a few scopes that use various combinations of `proc`, `Proc.new`, `lambda` and a stabby lambda. I think there's enough going on in this PR already to not worry too much about these, but running over the whole thing with Rubocop afterwards is probably a good idea.

bishboria:
Would it be best to get this and #353 done, and then the ruby upgrade + rubocop changes?

boffbowsh:
Yeah I guess so.

PR #354, opened 2015-06-08
Missing `uglifier` dependency in Gemfile

bishboria:
Deploy on preview (#1453) failed due to uglifier not being found.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ gem 'govuk_admin_template', '1.4.2'
 
 gem 'airbrake', '3.1.15'
 gem 'plek', '1.7.0'
-gem 'json', '1.7.7'
+gem 'json', '1.8.0'
 gem 'whenever', '~> 0.9.4', require: false
 
 gem 'uuid'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -10,7 +10,7 @@ gem 'govuk_admin_template', '1.4.2'
 
 gem 'airbrake', '3.1.15'
 gem 'plek', '1.7.0'
-gem 'json', '1.7.7'
+gem 'json', '1.8.0'
```
bradleywright: Seems unrelated to change?

Gemfile
```diff
@@ -41,6 +41,10 @@ gem 'rake', '10.4.1'
 
 gem 'sass-rails', '4.0.3'
 
+group :production do
```
bradleywright: This should be in `assets`: the command failing is:

    rake#RAILS_ENV=production RAILS_GROUPS=assets assets:precompile

bishboria: I spoke to @jennyd and she recommended removing the group I added…

Which is the preference?

bradleywright: Having just read the Rails 4 upgrade guide @jennyd is correct.

Gemfile
bishboria: Uglifier needs that version of json

PR #355, opened 2015-06-08
Make invalid and missing error messages the same

bradleywright:
In order to avoid username enumeration attacks on Signon, make all error
messages on the login screen the same.

The keys were gotten from:

https://github.com/plataformatec/devise/blob/4acb5043249d6fb5636d9f066aafcb0a11ba2765/config/locales/en.yml#L12-L15

but we had previously only overridden one of them.


boffbowsh:
:+1:

jabley:
:+1: EFG already does something similar.

bradleywright:
Thanks @boffbowsh!

PR #356, opened 2015-06-09
Fix queries using application.id on User model

jennyd:
These were accidentally broken in e96e674 - see commit messages for full details.


boffbowsh:
Looks good to me

PR #357, opened 2015-06-09
Simplify setting devise secret key

bishboria:
We only really care about the secret key being an environment variable in production. Set to a default dummy value in development and test.


boffbowsh:
:+1:

PR #358, opened 2015-06-10
Prevent unverified email changes

jamiecobbett:
Normal users shouldn't be able to directly change their own email address - they have to confirm the change via a link sent to the new email address (a notification is also sent to their old address).

Admins have been able to make email changes without requiring the user to confirm for some time. 

We accidentally gave normal users access to/a link to the admin-style page for themselves, which allows them to change their own email without confirmation.

By preventing their access to the full fledged edit page, we also mitigate some usability problems for normal users - it included various fields and links which would lead to the user seeing errors.

cc @jennyd @bishboria

### DIFF
--- a/spec/policies/user_policy_spec.rb
+++ b/spec/policies/user_policy_spec.rb
@@ -23,7 +23,7 @@ describe UserPolicy do
     end
   end
 
-  user_management_actions = [:edit?, :create?, :update?, :unlock?, :suspension?, :update_passphrase?,
+  user_management_actions = [:edit?, :create?, :update?, :unlock?, :suspension?,
     :cancel_email_change?, :resend_email_change?, :event_logs?]
 
   user_management_actions.each do |permission_name|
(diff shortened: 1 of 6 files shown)


bishboria:
:+1:

jamiecobbett:
Rebased on top of changes on master.

spec/policies/user_policy_spec.rb
```diff
@@ -64,7 +64,8 @@
     end
   end
 
-  (user_management_actions - [:event_logs?]).each do |permission_name|
+  self_management_actions = [:edit_email_or_passphrase?, :update_passphrase?, :cancel_email_change?, :resend_email_change?]
```
jennyd: Shouldn't `:update_email?` also be in this list?

jamiecobbett: Yes, rebased.

test/controllers/users_controller_test.rb
```diff
@@ -56,7 +56,7 @@ def change_user_password(user_factory, new_password)
       end
 
       should "show the unconfirmed_email" do
-        get :edit, id: @user.id
+        get :edit_email_or_passphrase, id: @user.id
```
jennyd: The context description needs updating for this change.

jamiecobbett: Well spotted - rebased.

PR #359, opened 2015-06-11
Remove sqlite dependency

jamiecobbett:
Originally Signon used this in development, but this was removed some years ago: 7b55f8dcd9c0368370738a77104d85fc1dd2fbb3. gds-sso has been using sqlite to run Signon in its integration tests, but this has now been switched to MySQL: https://github.com/alphagov/gds-sso/pull/73

The build is currently failing because the gds-sso build is failing because of a bug relating to Doorkeeper. This change is housekeeping, so can happily wait until everything else has settled before being merged.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,8 +45,6 @@ gem 'uglifier', '2.7.1'
 
 group :development do
   gem 'quiet_assets', '1.0.2'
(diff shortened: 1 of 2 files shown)


alext:
:+1: looks good to me.  We can probably remove sqlite from the puppet manifests as well now that it's no longer needed.

PR #360, opened 2015-06-12
Fix devise redirection bug

bishboria:
Devise 3.2.1 introduced a bug where it would discard querystring parameters when saving
the return URL for users who weren't signed in.


test/integration/authorise_application_test.rb
```diff
@@ -6,12 +6,17 @@ class AuthoriseApplicationTest < ActionDispatch::IntegrationTest
...
-  should "not confirm the authorisation for a signed-out user" do
-    visit "/oauth/authorize?response_type=code&client_id=#{@app.uid}&redirect_uri=#{@app.redirect_uri}"
-    assert_response_contains("You need to sign in")
-    signin(@user)
+  should "not confirm the authorisation until the user signs in" do
+    begin
+      visit "/oauth/authorize?response_type=code&client_id=#{@app.uid}&redirect_uri=#{@app.redirect_uri}"
+      assert_response_contains("You need to sign in")
+      signin(@user)
+    rescue ActionController::RoutingError
```
boffbowsh: Why might you get this exception here?

bishboria: During testing, requests for all domains get routed to Signon; including the capybara browser being redirected to other apps.

The browser gets a redirect to url of the destination app. This then gets routed to Signon but Signon doesn't know how to handle the route.

Should I add a small comment in the test similar to the above explanation?

bishboria: We've rebased to fix and added an explanatory comment

jamiecobbett:
I worked on this with @bishboria, but I think it is good to merge, it fixes the gds-sso build and it makes Signon deployable again.

boffbowsh:
:+1:

test/integration/authorise_application_test.rb
```diff
@@ -6,22 +6,44 @@ class AuthoriseApplicationTest < ActionDispatch::IntegrationTest
     @user = create(:user)
   end
 
-  should "not confirm the authorisation for a signed-out user" do
+  should "not confirm the authorisation until the user signs in" do
     visit "/oauth/authorize?response_type=code&client_id=#{@app.uid}&redirect_uri=#{@app.redirect_uri}"
     assert_response_contains("You need to sign in")
-    signin(@user)
 
-    refute Doorkeeper::AccessGrant.find_by(resource_owner_id: @user.id)
```
jennyd: Did you mean to remove this assertion? Without it, the test isn't checking that the access grant is only created after signin so the test description isn't quite accurate.

jamiecobbett: This was mistakenly switched to a `refute` in b19763283d94a545b140958bb4a5d99db00a8f19 - I think the test was therefore wrong - so this is switching it back. You are right though, we should have a `refute` before the `signin(@user)` line.

jamiecobbett: PR: https://github.com/alphagov/signonotron2/pull/361

PR #364, opened 2015-06-16
Explicitly mark session cookies as "Secure"

jamiecobbett:
This means that cookies will only be sent over encrypted connections, thus
making it harder for an attacker to read/replay someone else's session.

The secure option was accidentally removed in 0ad105f.


bishboria:
:+1:

PR #367, opened 2015-06-17
Version bump of the admin gem

bishboria:
Due to a vulnerability with jquery-rails. See See
https://groups.google.com/forum/#!topic/rubyonrails-security/XIZPbobuwaY
for details

* admin gem also handles google analytics so analytics code in
  aplication.html.erb can be removed


bishboria:
Handily, this change also uses the fixed version of `rack`

fofr:
Looks good to merge, once tests pass

PR #368, opened 2015-06-17
Rails rack security hotfix

jamiecobbett:
*Please review, but don't merge*

Signon has a large number of changes that are pending, and in any case we would need to upgrade Signon master from 4.0.x to 4.1.x to get the security fixes as 4.0.x is unsupported.

Instead, we'll do a hotfix branched from the currently deployed version - release_534.


boffbowsh:
Bon :+1:

evilstreak:
Can this be closed now?

jamiecobbett:
@evilstreak yes - this has been deployed and master has had the equivalent fixes applied already.

PR #369, opened 2015-06-17
Upgrade rails to 4.1.x (4.1.11)

jamiecobbett:
This gets us back on to a currently-supported version of Rails (3.2.x, 4.1.x
and 4.2.x are currently supported, but 4.0.x is not).

ActiveRecord has an implicit dependency on up-to-date mysql2 and Rails requires
updated minitest.

Rails 4.1.1 includes the security fixes released on the 16th June and we're already on jquery-rails and  rack versions with those fixes.


boffbowsh:
This seems terribly easy? Can we deploy it to preview and have a click around?

PR #371, opened 2015-07-01
Add rake task for granting content preview permission.

elliotcm:
This rake task adds permissions to the "Content
Preview" aka authentication-proxy application for
all users who a) are not suspended and b) do not
already have it.

It also triggers a push to the application in
question to sync the permissions.

https://trello.com/c/LKNZT6qZ/185-content-preview-permissions-are-a-default-permission-in-signon

### DIFF
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -82,4 +82,21 @@ namespace :users do
 
     UserPermissionsExporter.new(ENV['EXPORT_DIR'], Logger.new(STDOUT)).export_signon
   end
+
+  desc "Grant access to Content Preview for all active users who don't have it"
+  task :grant_content_preview_access => :environment do
+    if content_preview = Doorkeeper::Application.find_by(name: "Content Preview")
+      User.not_suspended.find_each do |user|
+        next if user.authorised_applications.include?(content_preview)
+
+        user.grant_application_permission(content_preview, "signin")
+
+        if content_preview.supports_push_updates?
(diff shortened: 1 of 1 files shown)


lib/tasks/users.rake
```diff
@@ -82,4 +82,21 @@ namespace :users do
 
     UserPermissionsExporter.new(ENV['EXPORT_DIR'], Logger.new(STDOUT)).export_signon
   end
+
+  desc "Grant access to Content Preview for all active users who don't have it"
+  task :grant_content_preview_access => :environment do
+    if content_preview = Doorkeeper::Application.find_by(name: "Content Preview")
+      User.not_suspended.each do |user|
```
alext: Is it worth using `find_each` here to make it more memory-efficient?

alext:
Is there already a mechanism in place to automatically add this to any users created after this point (similar to permissions for the support app)?

lib/tasks/users.rake
elliotcm: We've only 5k active users but it can't hurt, I'll update to reflect.

elliotcm:
@alext There'll be discussions happening shortly between @bradleywright and 1st line support about that, but it's unsure at the moment whether that will happen, which is why I made it a rake task rather than a migration and made it only update users which don't have the permission already.

My plan was to delete the rake task if/when we add that functionality.

lib/tasks/users.rake
elliotcm: @alext updated.

PR #374, opened 2015-07-07
Clearer password reset error message

tommyp:
The previous message was also displayed if you reused the previous
password. This commit updates it to also include this possible solution
to the error.

### DIFF
--- a/app/views/devise/passwords/reset_error.html.erb
+++ b/app/views/devise/passwords/reset_error.html.erb
@@ -5,6 +5,6 @@
 </div>
 
 <div class="alert alert-warning">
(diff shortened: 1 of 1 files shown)


app/views/devise/passwords/reset_error.html.erb
```diff
@@ -5,6 +5,6 @@
 </div>
 
 <div class="alert alert-warning">
-  <p>That passphrase reset didn’t work. It may be that the link you were using has expired or was invalid.</p>
+  <p>That passphrase reset didn’t work. It may be that you've used this passphrase already, or the link you were using has expired or was invalid.</p>
```
fofr: Typography pedantry: Can we use a smart quote in "you’ve"? (Sorry)

PR #378, opened 2015-07-22
Remove devise_security_extension gem

boffbowsh:
As preparation for the upgrade to Ruby 2.2, we want to remove the dependency on [devise_security_extension](https://github.com/phatworx/devise_security_extension). The version we were using was incompatible with Ruby 2.2 when we [tried to upgrade previously](https://github.com/alphagov/signonotron2/pull/375), and bumping it introduced new errors.

The gem is too poorly-tested to be relied upon, and regularly seems to introduce breaking changes. It includes functionality we don't need for signon, we use it for passphrase expiry and ensuring users don't reuse passphrases. It was decided that a better approach would be to ensure good tests for the functionality within the app, then re-implement it.

This PR adds a few tests to help validate behaviour, then essentially copies the functionality from the gem (don't worry, it's MIT licensed) and tidies it up a bit. This includes adding the `govuk-lint` tool to the CI pipeline to lint the changes in this PR.

The next step after this PR is to bump the Ruby version to 2.2 again and fix any failures.

TODO:

- [x] Fix GDS SSO tests


boffbowsh:
Fix for gds-sso: https://github.com/alphagov/gds-sso/pull/75

PR #379, opened 2015-07-24
Upgrade to Ruby 2.2.2

boffbowsh:
Neither kgio 2.7.4 or json 1.8.0 would compile, so bumped both
versions. kgio is only bumped via unicorn, so upgraded that too.
Syck is no longer provided, so bump safe_yaml to accommodate this.

Trello: https://trello.com/c/sgAorLlb/58-upgrade-signon-to-ruby-2-2-medium


jamiecobbett:
:fireworks:

PR #380, opened 2015-07-27
Ensure oauth authorisation respects password expiry

boffbowsh:
Trello: https://trello.com/c/wQufRA6Z/65-signon-client-apps-don-t-respect-passphrase-expiry

When Doorkeeper performs authentication and authorization of a user in
order to allow them access to an application, it doesn’t check whether
the user’s password is expired.

This change adds a check to the configuration of Doorkeeper, where it
will redirect the user to the passphrase expiry screen when they
attempt to access an application if required.

/cc @jamiecobbett @bradleywright


jamiecobbett:
Tested successfully in development. Good work :+1:

PR #381, opened 2015-07-28
Upgrade to Rails 4.2.3

boffbowsh:
As per title. Best viewed commit-by-commit in `?w=1` mode.

A few highlights:

- Upgrades Devise to 3.5.1
- Fixes a lot of deprecation warnings:
  - CSS selectors in tests (`a[href=foo]` should be `a[href="foo"]`)
  - Performance tests are extracted into a gem, and weren't used anyway so removed the infrastructure around them
  - Devise's `confirm!` is now `confirm`.
  - ActionMailer's `deliver` should now be `deliver_now` or `deliver_later`
  - `serve_static_assets` is now `serve_static_files`
  - Rails now prefers random test ordering
- Deliver all email asynchronously unless already part of a rake or background task
- Switch to using ActiveJob for background tasks instead of Sidekiq directly
  - Provides abstraction in case we need to switch to another queuing system
  - Tidies up test code
  - Removes need for `devise-async` gem as ActionMailer can now background delivery itself
- Remove tests and code around email blacklist detection for invites 
  - As the invite emails were being backgrounded anyway even before this PR, there was no way of detecting delivery failures anyway. We now no longer pretend in the tests that we can.


Gemfile
```diff
@@ -16,11 +16,10 @@ gem 'whenever', '~> 0.9.4', require: false
 gem 'uuid'
 
 # Gems used to provide Authentication and Authorization services.
-gem 'devise', '3.2.2'
-gem 'devise_invitable', '1.4.0'
-gem 'devise-encryptable', '0.1.1'
+gem 'devise', '~> 3.4'
```
jamiecobbett: I think this should specify the exact version because Devise is so complicated - changes to the version we're using should explicit. Additionally, our [styleguide](https://github.com/alphagov/styleguides/blob/master/ruby.md#gemfile) says pessimistic locking should be used *when the project follows semver*, which Devise doesn't claim to, and implicitly that we should use it with the PATCH version, not the MINOR version.

config/sidekiq.yml
```diff
@@ -8,3 +8,4 @@
 :queues:
   - default
   - mailer
```
jamiecobbett: I think its worth adding a comment to say this should be removed once the changes are deployed, as per the commit message.

app/models/user.rb
```diff
@@ -24,7 +24,6 @@ class User < ActiveRecord::Base
          :validatable, :timeoutable, :lockable,                # devise core model extensions
          :invitable,    # in devise_invitable gem
          :suspendable,  # in signonotron2/lib/devise/models/suspendable.rb
-         :async,        # in devise_async gem, send mailers async'ly
```
jamiecobbett: Nice spot :+1:

jamiecobbett:
This is looking good :+1: I'll leave it open for a little while to give others a chance, otherwise I'll merge tomorrow.

benilovj:
Looks really good! Merging.

PR #382, opened 2015-07-29
Rubocop all the things

boffbowsh:
As we (core formats team) are going to be giving Signon a bit of love over the next few ${unit_of_work}s, and now that it's upgraded to latest Ruby and Rails, we figure it's a good time to get it fully up to spec on the styleguide front.

Rubocop managed to autocorrect the vast majority of contraventions, with just a few manual changes required. They're split into separate commits for app, lib, test, config and spec directories.

This change also makes rubocop lint the whole repo in CI, rather than just the diff. This shouldn't make a difference to anything, but it's more correct to lint the whole repo every time if we're aiming for an entirely clean repo.

/cc @jamiecobbett @benilovj 

allthethings


mattbostock:
\o/

jamiecobbett:
:+1: for intent

alext:
I'm not convinced these massive cleanup commits are a good idea.  It makes it harder to use `git blame` and the like to find the history and reasoning for a specific line of code. Signon is an application where that's more important than most.

Would it not be better to adopt an approach where things are cleaned up as they're touched?

jamiecobbett:
> It makes it harder to use git blame and the like to find the history and reasoning for a specific line of code.

I almost never use `blame`, I use `git log -S"string i'm interested in"` as it doesn't just stop with the most recent change to that line and includes occurrences across files (by default).

mattbostock:
@alext: I'm not sure I agree; if the code cleanup is contained in logical commits (as it is here), I don't think that's good enough reason not to do it.

What you describe is essentially a tooling problem IMHO; I'd also like a tool that would make git blame easier, when the git pickaxe does not suffice. Something like [git-browse](https://github.com/georgebrock/git-browse).

boffbowsh:
Also, it's really easy to start at a particular commit in `git blame`, so if you happen to need to know info about a particular line that was changed as part of this PR, just start at `<commit>~1`. It affects just ~5% of LOC in the repo anyway, which isn't a huge amount.

boffbowsh:
I'd also argue that because this is signon, consistent style across the repo is very important in order to reduce mistakes, both when writing new code and reasoning about old.

app/controllers/users_controller.rb
```diff
@@ -40,7 +40,7 @@ def index
 
   def update
     raise Pundit::NotAuthorizedError if current_user.organisation_admin? &&
-      ! current_user.organisation.subtree.map(&:id).include?(params[:user][:organisation_id].to_i)
+                                        ! current_user.organisation.subtree.map(&:id).include?(params[:user][:organisation_id].to_i)
```
mattbostock: This indentation looks excessive; can this Rubocop behaviour be disabled or tweaked?

app/controllers/users_controller.rb
```diff
@@ -143,7 +146,7 @@ def should_include_permissions?
 
   def paginate_users
     if any_filter?
-      unless @users.kind_of?(Array)
+      if not @users.is_a?(Array)
```
mattbostock: It would seem more idiomatic to use `unless` here.

spec/policies/user_policy_spec.rb
```diff
@@ -24,7 +24,7 @@
   end
 
   user_management_actions = [:edit?, :create?, :update?, :unlock?, :suspension?,
-    :cancel_email_change?, :resend_email_change?, :event_logs?]
+                             :cancel_email_change?, :resend_email_change?, :event_logs?]
```
mattbostock: Ditto my comment above about indentation.

app/controllers/users_controller.rb
alext: Actually it would be better to reverse the conditional here.  The rubocop rule is to avoid `unless ... else` constructs.  Using `if not` will fix the warning, but not the intent behind the warning.

mattbostock: Yep, agree.

spec/policies/user_policy_spec.rb
boffbowsh: This is correct according to the main [Ruby style guide](https://github.com/bbatsov/ruby-style-guide#align-multiline-arrays), but really the elements should be on multiple lines (and wrapped in `%i()`, so I've added that.

boffbowsh:
Made a change in `govuk-lint` for this: https://github.com/alphagov/govuk-lint/compare/indented-multiline-operations.

If this PR gets approval, I'll merge that change and do the gem version shuffle here.

boffbowsh:
Bamp

mattbostock:
LGTM :+1:

I'll leave it for someone else with better knowledge of this application and more experience with Ruby syntax to merge.

@boffbowsh Do you want to remove the 'Do not merge' phrase from the PR title?

boffbowsh:
Not until the govuk-lint gem shuffle has happened

boffbowsh:
https://github.com/alphagov/govuk-lint/pull/13 needs merging

boffbowsh:
This can now be merged

boffbowsh:
Bump.

PR #383, opened 2015-07-31
Minor post-Rails 4.2.3 fixes for background jobs

boffbowsh:
A couple of errbit errors have cropped up in production, https://errbit.production.alphagov.co.uk/apps/52ef92340da11517d5000003/problems/55bb0f21657863081b5b0000 and https://errbit.production.alphagov.co.uk/apps/52ef92340da11517d5000003/problems/55ba34a2657863726d6d0000. This addresses those errors. 

Review commit-by-commit, each commit has full details of the change.

/cc @jamiecobbett

### DIFF
--- a/lib/inactive_users_suspension_reminder.rb
+++ b/lib/inactive_users_suspension_reminder.rb
@@ -5,7 +5,8 @@ class InactiveUsersSuspensionReminder
                         Net::SMTPServerBusy,
                         Errno::ETIMEDOUT,
                         Errno::EHOSTUNREACH,
-                        Errno::ECONNREFUSED]
+                        Errno::ECONNREFUSED,
+                        EOFError]
 
   def initialize(users, days_to_suspension)
     @users, @days_to_suspension = users, days_to_suspension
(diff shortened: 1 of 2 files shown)


lib/inactive_users_suspension_reminder.rb
```diff
@@ -19,14 +20,18 @@ def send_reminders
         UserMailer.suspension_reminder(user, @days_to_suspension).deliver_now
         Rails.logger.info "#{self.class}: Successfully sent email to #{user.email}."
       rescue *ERRORS_TO_RETRY_ON => e
+        notify_airbrake(e, user)
         Rails.logger.debug "#{self.class}: #{e.class} - #{e.message} while sending email to #{user.email} during attempt (#{(tries..3).count}/3)."
         sleep(3) and retry if (tries -= 1) > 0
 
         Rails.logger.warn "#{self.class}: Failed to send suspension reminder email to #{user.email}."
-        notify_airbrake(e, user)
```
alext: This changes the notification behaviour here.  Perviously it would only notify Errbit when it had given up retrying. This is desireable, otherwise the Errbit notifications can get quite noisy.

boffbowsh: Good point, thanks. Fixed, rebased, force-pushed.

lib/inactive_users_suspension_reminder.rb
```diff
@@ -25,8 +26,12 @@ def send_reminders
         Rails.logger.warn "#{self.class}: Failed to send suspension reminder email to #{user.email}."
         notify_airbrake(e, user)
       rescue => e
-        Rails.logger.warn "#{self.class}: #{e.response.error.message} while sending email to #{user.email}."
         notify_airbrake(e, user)
+        begin
+          Rails.logger.warn "#{self.class}: #{e.response.error.message} while sending email to #{user.email}."
+        rescue NoMethodError
```
jamiecobbett: Instead of waiting for an error to be raised, maybe we could do something like:
```ruby
if e.responds_to?(:response)
  Rails.logger.warn "#{self.class}: #{e.response.error.message} while sending email to #{user.email}."
else
  Rails.logger.warn "#{self.class}: #{e.message} while sending email to #{user.email}."
end
```

PR #385, opened 2015-08-05
Add optional argument for apps not to update domain.

joshmyers:
We need a way to skip certain apps i.e. licensify 
as that is staying at the old domain name. The 
argument to skip apps is optional. If want
to skip an application the arguments are passed in
as so: 

rake applications:migrate_domain['licensify, foo']

### DIFF
--- a/lib/tasks/applications.rake
+++ b/lib/tasks/applications.rake
@@ -20,10 +20,11 @@ namespace :applications do
     puts "config.oauth_secret = '#{a.secret}'"
   end
 
-  desc 'Updates domain name for applications'
-  task migrate_domain: :environment do
+  desc 'Updates domain name for applications, optionally pass CSV of apps to ignore'
+  task :migrate_domain, [:apps_to_ignore] => :environment do |_t, args|
+    apps_to_ignore = args.any? ? args[:apps_to_ignore].split(',') : []
     raise "Requires OLD_DOMAIN + NEW_DOMAIN specified in environment" unless ENV['OLD_DOMAIN'] && ENV['NEW_DOMAIN']
-    Doorkeeper::Application.find_each do |application|
+    Doorkeeper::Application.where.not(name: apps_to_ignore).find_each do |application|
       [:redirect_uri, :home_uri].each do |field|
         new_domain = application[field].gsub(ENV['OLD_DOMAIN'], ENV['NEW_DOMAIN'])
         if application[field] != new_domain


lib/tasks/applications.rake
```diff
@@ -21,10 +21,12 @@ namespace :applications do
...
   end
 
-  desc 'Updates domain name for applications'
-  task :migrate_domain => :environment do
+  desc 'Updates domain name for applications, optionally pass in apps to ignore'
+  task :migrate_domain, [:apps_to_ignore] => :environment do |t, args|
+    apps_to_ignore = args.any? ? args[:apps_to_ignore].split(' ') : []
     raise "Requires OLD_DOMAIN + NEW_DOMAIN specified in environment" unless ENV['OLD_DOMAIN'] && ENV['NEW_DOMAIN']
     Doorkeeper::Application.find_each do |application|
+      next if apps_to_ignore.include?(application)
```
bradleywright: It would be more efficient to use `where.not(home_uri: args[:apps_to_ignore])` - it can take an array.

bradleywright: This line also won't work - `application` is an ActiveRecord object, and `apps_to_ignore` is a list of strings.

joshmyers: good point, it would have to be application[:name]. I will change to use `where.not`. Thanks @bradleywright

lib/tasks/applications.rake
```diff
@@ -20,10 +20,11 @@ namespace :applications do
     puts "config.oauth_secret = '#{a.secret}'"
   end
 
-  desc 'Updates domain name for applications'
-  task migrate_domain: :environment do
+  desc 'Updates domain name for applications, optionally pass in apps to ignore'
+  task :migrate_domain, [:apps_to_ignore] => :environment do |_t, args|
+    apps_to_ignore = args.any? ? args[:apps_to_ignore].split(' ') : []
```
alexmuller: Would it be better to split on something like a comma to prevent command line weirdness with spaces potentially running different rake tasks?

bradleywright: That's a good point - the names of apps do have spaces in them.

joshmyers: Aye, good point @alexmuller - I have amended the PR and addressed your comment.

@alexmuller / @bradleywright OK to merge? I'd like to get [this](https://www.pivotaltracker.com/story/show/98809362) story finished.

PR #387, opened 2015-08-11
Add some more logging on password reset flows

edds:
Log when a user displays a password reset page and when they fail to
reset their password. This will help us diagnose if a user is indeed
attempting to reset their password and if they are failing to reset
their password. Currently we were only getting the happy case giving us
no visibility if anything had gone wrong.


benilovj:
Code LGTM

PR #389, opened 2015-08-17
Don't allow autocompletion of passwords on the New Password fields

issyl0:
- This will help mitigate against browsers autocompleting the new
  password fields with the user's old password, leading to them not
  being able to log in the next time, ie:
  https://govuk.zendesk.com/agent/tickets/1104158.

### DIFF
--- a/app/views/devise/passwords/_change_password_panel.html.erb
+++ b/app/views/devise/passwords/_change_password_panel.html.erb
@@ -20,7 +20,8 @@
                                :"data-weak-words" => user_email_tokens(user).join(","),
                                :"data-min-password-length" => minimum_password_length,
                                :required => true, :"aria-required" => true,
-                               :class => 'form-control input-md-5' %>
+                               :class => 'form-control input-md-5',
+                               :autocomplete => 'off' %>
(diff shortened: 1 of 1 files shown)


edds:
Has this been tested in that situation? I ask because the advice is:

> For this reason, many modern browsers do not support autocomplete="off" for login fields.
>
> - if a site sets autocomplete="off" for a form, and the form includes username and password input fields, then the browser will still offer to remember this login, and if the user agrees, the browser will autofill those fields the next time the user visits this page.
> - if a site sets autocomplete="off" for username and password input fields, then the browser will still offer to remember this login, and if the user agrees, the browser will autofill those fields the next time the user visits this page.
>
> This is the behavior in Firefox (since version 38), Google Chrome (since 34), and Internet Explorer (since version 11).

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

issyl0:
Honestly: no. I don't know how to test it now that that ticket is solved though. I suppose we ought to be careful what we do here. This was just the approach suggested by @marksheldonGDS, linking to those same docs. Closing. Sorry.

PR #392, opened 2015-09-04
Minimal 2-step Verification process

boffbowsh:
This is the minimally-styled MVP for 2-step Verification: https://trello.com/c/Y0H24xy4/86-implement-minimal-2sv-for-selected-beta-users-in-production-and-preview-large

It requires styling as per https://github.com/alphagov/signonotron2/tree/prototype-2fa which will be done as a separate story.

This is intended to be deployed to production where members of Core Formats will be able to explicitly go and enable 2SV on their account for testing and user research purposes.

The QR code is rendered in the page as a PNG using a data-uri

Near-future stories will style the pages correctly and allow certain classes of users to be marked as requiring 2SV.

This PR relies on a fork of the `two_factor_authentication` gem. Our change has been contributed upstream and is awaiting merge: https://github.com/Houdini/two_factor_authentication/pull/54

/cc @jamiecobbett @fofr @benilovj


jamiecobbett:
I think it would be worth adding an integration test to show that 2SV is required when resetting a passphrase.

boffbowsh:
@jamiecobbett do you mean that if a passphrase has been reset by admin, the same 2SV secret still applies? Or that additional 2SV entry is required in order to change your own passphrase?

boffbowsh:
I've updated this to enforce sign in after password reset.

app/views/devise/two_step_verification/new.html.erb
```diff
@@ -0,0 +1,88 @@
+<h1>Set up 2-step verification</h1>
```
jamiecobbett: This page has a title of `| GOV.UK Signon`. Would you mind calling `content_for(:title)`, probably with the same copy?

The page for entering the code has the same problem, but we need to override that template to solve that problem, so I'm happy to wait for the frontend ticket for that /cc @fofr .

jamiecobbett:
Apart from the minor comment, this looks good to merge :+1:

boffbowsh:
Fixed styling, fixed page title, added QR code

fofr:
:key: :iphone: :+1:

PR #393, opened 2015-09-08
Show the deployment instance name in the 2SV issuer

boffbowsh:
For non-production environments, prefix "GOV.UK Signon" with the instance name in the issuer field of the QR code. This will show up in the description label in mobile Authenticator apps:

photo 08-09-2015 10 17 08

### DIFF
--- a/config/initializers/instance_name.rb
+++ b/config/initializers/instance_name.rb
@@ -3,4 +3,8 @@
 
 # Human-friendly name for this signon instance. This is used when generating
 # reminder emails that need to disambiguate between instances.
-Rails.application.config.instance_name = nil
+if Rails.env.development? || Rails.env.test?
+  Rails.application.config.instance_name = "development"
+else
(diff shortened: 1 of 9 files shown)


config/initializers/instance_name.rb
```diff
@@ -3,4 +3,4 @@
 
 # Human-friendly name for this signon instance. This is used when generating
 # reminder emails that need to disambiguate between instances.
-Rails.application.config.instance_name = nil
+Rails.application.config.instance_name = "development"
```
jamiecobbett: This file isn't overwritten in production, so this would be used in production. Maybe add something like `Rails.env.development? ? "development" : nil`

boffbowsh: Updated and rebased

jamiecobbett:
Looks good to me - @fofr are you happy with this?

fofr:
Yep :+1:

PR #395, opened 2015-09-11
Speed up slow user export functionality

boffbowsh:
Trello: https://trello.com/c/AbFXHJ5u/104-504-timeout-when-trying-to-download-a-csv-of-signon-users-medium

The "Export as CSV" button in signon will generally result in a 504 in production due to it taking over 15 seconds to generate the CSV (actually around 10 minutes on a VM). This is mostly avoidable by fetching data in more optimised ways rather than relying on ActiveRecord's association magic.

CSVs with no filtering are now returned in around 4 seconds on a VM, which should be quick enough.

This pull request best reviewed commit by commit.

Possible future refactoring would be to pass a scope of users into the presenter and have it generate the CSV itself.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -58,6 +58,27 @@ class User < ActiveRecord::Base
   scope :last_signed_in_after, lambda { |date| web_users.not_suspended.where('date(current_sign_in_at) >= date(?)', date) }
   scope :not_recently_unsuspended, lambda { where(['unsuspended_at IS NULL OR unsuspended_at < ?', UNSUSPENSION_GRACE_PERIOD.ago]) }
 
+  scope :with_status, lambda { |status|
+    case status
+    when USER_STATUS_SUSPENDED
+      where.not(suspended_at: nil)
+    when USER_STATUS_INVITED
+      where.not(invitation_sent_at: nil).where(invitation_accepted_at: nil)
+    when USER_STATUS_PASSPHRASE_EXPIRED
+      with_need_change_password
+    when USER_STATUS_LOCKED
+      where.not(locked_at: nil)
+    when USER_STATUS_ACTIVE
+      where(suspended_at: nil, locked_at: nil).
+        where(arel_table[:invitation_sent_at].eq(nil).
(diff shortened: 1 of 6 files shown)


app/models/user.rb
```diff
@@ -58,6 +58,25 @@ class User < ActiveRecord::Base
   scope :last_signed_in_after, lambda { |date| web_users.not_suspended.where('date(current_sign_in_at) >= date(?)', date) }
   scope :not_recently_unsuspended, lambda { where(['unsuspended_at IS NULL OR unsuspended_at < ?', UNSUSPENSION_GRACE_PERIOD.ago]) }
 
+  scope :with_status, lambda { |status|
+    case status
```
benilovj: if a new status is added (eg for 2sv) and we forget to update this code (which is quite likely), this statement will not filter correctly. Is it worth adding a final condition that sends an exception to Errbit when the status hasn't been matched?

lib/devise/models/password_expirable.rb
```diff
@@ -11,6 +11,19 @@ module PasswordExpirable
 
       included do
         before_save :update_password_changed
+
+        scope :with_need_change_password, -> do
+          if self.expire_password_after.is_a?(Fixnum) || self.expire_password_after.is_a?(Float)
```
benilovj: The expression `self.expire_password_after.is_a?(Fixnum) || self.expire_password_after.is_a?(Float)` appears 3 times in this module, and it's not the most readable thing in the world. Is it worth extracting it into a well-named method?

app/models/user.rb
boffbowsh: I think it's probably worth just raising an exception completely. Good spot.

benilovj:
Apart from the nitpicks, seems reasonable to me.

app/models/user.rb
benilovj: @fofr just made an excellent suggestion to write a unit test that loops through `User::USER_STATUSES` and makes sure that all statuses are covered

boffbowsh:
Comments addressed and rebased

PR #396, opened 2015-09-11
Revert "Don't automatically log users in after password reset"

boffbowsh:
This reverts commit 41c1e1edf410f46b8239bf89e9c3156a974c33b2 as per https://trello.com/c/0WvaVtE6/109-undo-adding-sign-in-screen-after-successful-passphrase-reset.

This is being reverted because as part of the wider 2SV story, we will be requiring 2SV for users before they can reset their password. Thus, forcing them to be logged out on a successful password reset would mean they'd have to enter a 2SV code twice.

/cc @fofr

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -193,9 +193,6 @@ Devise.setup do |config|
   # change their passwords.
   config.reset_password_within = 6.hours
 
-  # Don't automatically sign a user in once they've reset their password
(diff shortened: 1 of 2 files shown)


fofr:
It's worth adding that this also caused a regression – users without 2SV would be logged out after completing the arduous passphrase reset process, prolonging the time until they can log in and do what they signed in to do.

fofr:
:+1:

PR #400, opened 2015-09-14
Reimplement two_factor_authentication gem within signon

boffbowsh:
After hitting a few too many hurdles with this gem and considering the upcoming stories around 2SV, we decided to re-implement it internally and remove the features we don't need. This should reduce friction going forwards and allow us to maintain the project more easily.

This is mostly a copy and paste from two_factor_authentication, but with occurrences of that term replaced with two_step_verification or similar.

This PR best reviewed commit-by-commit. Due to the nature of the work, each commit is probably not fully functional, they're just separated by approximate areas of work.


boffbowsh:
Context for purposes of reviewing this:

Currently this is only intended for use in testing and user research by the Core Formats team. There is a staged rollout planned including a small number of members of GDS, and will be very gradually rolled out to the rest of GDS then wider government and local government. In theory, merging this should be zero impact for non-Core Formats members.

Integration test coverage added whilst adding the gem in the first place has been used to verify this port.

Trello stories:
- [This story](https://trello.com/c/fsnQl7V1/111-adopt-2sv-implementation-from-the-gem-into-our-own-code)
- [Minimal 2SV MVP](https://trello.com/c/Y0H24xy4/86-implement-minimal-2sv-for-selected-beta-users-in-production-and-preview-large)
- [Hooks needed for logging 2SV actions](https://trello.com/c/LbIGKlNh/108-log-user-actions-for-mvp-2sv)

Things that would be harder to implement without this change:

- Email recovery
  - If a user doesn't have access to their 2SV device, they can request an email containing a one-time password which will allow them to log in. Implementing this with the gem would require monkey-patching the Warden hook.
- Multi-device support
  - When we enable this, the OTP secret will move from the user model to a device table. Doing this would require contributing this change upstream to the gem, which would require a difficult migration for the gem's other users.
- Prompting users to set up 2SV
  - Certain users will need to be flagged as either mandatory or encouraged to set up 2SV. This would require new hooks that would have to be compatible with the ones provided by the gem.

/cc @jennyd

jamiecobbett:
:+1:

app/models/user.rb
```diff
@@ -202,29 +204,27 @@ def send_devise_notification(notification, *args)
...
     if result
       EventLog.record_event(self, EventLog::TWO_STEP_VERIFIED)
     else
       EventLog.record_event(self, EventLog::TWO_STEP_VERIFICATION_FAILED)
     end
+
     result
   end
 
+  def max_login_attempts?
```
alext: Is the naming of this method correct? It looks to be related specifically to 2SV, so should probably be named to indicate that.

alext:
One small point about method naming, otherwise looks good to me.

app/models/user.rb
boffbowsh: Good point. This is carried over from the gem, so will rename to something more important. Thanks.

PR #401, opened 2015-09-22
Add 2SV session memory

boffbowsh:
This was missed in the removal of the two_factor_authentication gem
from this repo.

This also improves upon the security of the original implementation by
verifying user id and expiration, making it impossible to copy the
value of the signed gem and re-use it later, extend the expiration
period, or use your cookie for another user.

/cc @jamiecobbett

### DIFF
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -15,7 +15,7 @@ class Devise::TwoStepVerificationController < DeviseController
       expires_seconds = User::REMEMBER_2SV_SESSION_FOR
       if expires_seconds && expires_seconds > 0
         cookies.signed['remember_2sv_session'] = {
-          value: {user_id: current_user.id, expires: expires_seconds.from_now},
+          value: {user_id: current_user.id, valid_until: expires_seconds.from_now},
           expires: expires_seconds.from_now
         }
       end
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -12,6 +12,13 @@ class Devise::TwoStepVerificationController < DeviseController
     render(:show) && return if params[:code].nil?
 
     if current_user.authenticate_otp(params[:code])
+      expires_seconds = User::REMEMBER_2SV_SESSION_FOR
+      if expires_seconds && expires_seconds > 0
+        cookies.signed['remember_2sv_session'] = {
+          value: {user_id: current_user.id, expires: expires_seconds.from_now},
+          expires: expires_seconds.from_now
+        }
+      end
       warden.session(:user)['need_two_step_verification'] = false
       sign_in :user, current_user, bypass: true
       set_flash_message :notice, :success
--- a/lib/devise/hooks/two_step_verification.rb
+++ b/lib/devise/hooks/two_step_verification.rb
@@ -1,5 +1,8 @@
 Warden::Manager.after_authentication do |user, auth, _options|
   if user.respond_to?(:need_two_step_verification?)
-    auth.session(:user)['need_two_step_verification'] = user.need_two_step_verification?
+    cookie = auth.env["action_dispatch.cookies"].signed["remember_2sv_session"]
+    unless cookie && cookie[:user_id] == user.id && cookie[:expires] > Time.now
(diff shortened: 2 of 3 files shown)


benilovj:
LGTM but best to get more eyes on this

test/integration/sign_in_test.rb
```diff
@@ -76,6 +76,27 @@ class SignInTest < ActionDispatch::IntegrationTest
       assert_selector "input[name=code]"
     end
 
+    should "not prompt for a verification code twice in 30 days" do
```
jamiecobbett: Is it worth adding "per device" or "per browser" for clarity?

boffbowsh: Doen

app/controllers/devise/two_step_verification_controller.rb
```diff
@@ -12,6 +12,13 @@ def update
     render(:show) && return if params[:code].nil?
 
     if current_user.authenticate_otp(params[:code])
+      expires_seconds = User::REMEMBER_2SV_SESSION_FOR
+      if expires_seconds && expires_seconds > 0
+        cookies.signed['remember_2sv_session'] = {
+          value: {user_id: current_user.id, expires: expires_seconds.from_now},
```
jamiecobbett: Looking at the [docs](http://api.rubyonrails.org/v4.2.4/classes/ActionDispatch/Cookies.html), the `expires: expires_seconds.from_now` on this line looks unnecessary

boffbowsh: This is added as an additional value to be signed in order to prevent users from extending the period of the cookie or copying it to be reused in a new period.

boffbowsh: Perhaps `valid_until` would be a better name?

lib/devise/hooks/two_step_verification.rb
```diff
@@ -1,5 +1,8 @@
 Warden::Manager.after_authentication do |user, auth, _options|
   if user.respond_to?(:need_two_step_verification?)
-    auth.session(:user)['need_two_step_verification'] = user.need_two_step_verification?
+    cookie = auth.env["action_dispatch.cookies"].signed["remember_2sv_session"]
+    unless cookie && cookie[:user_id] == user.id && cookie[:expires] > Time.now
```
jamiecobbett: I'd be tempted to pull this line and the line above out into a method called something like `has_valid_2sv_cookie?`

app/controllers/devise/two_step_verification_controller.rb
jamiecobbett: I think the important thing to express is that isn't redundant and does need to be part of the signed value. Maybe we should have a test which tampers with the unsigned expiry?

boffbowsh:
Renamed the `expires` key in the signed cookie
Added tests for potential attack vectors

@jamiecobbett

test/integration/sign_in_test.rb
```diff
@@ -133,5 +154,38 @@ class SignInTest < ActionDispatch::IntegrationTest
...
+    should "not permit an expired cookie to be used to bypass 2SV" do
+      visit root_path
+      signin_with_2sv(email: "email@example.com", password: "some passphrase with various $ymb0l$")
+      remember_2sv_session = Capybara.current_session.driver.request.cookies["remember_2sv_session"]
+
+      Timecop.travel(30.days.from_now + 1) do
+        visit root_path # Clears the expired cookie from the Capybara session
+        Capybara.current_session.driver.request.cookies["remember_2sv_session"] = remember_2sv_session
+
+        visit root_path
```
jamiecobbett: If the previous `visit root_path` cleared the expired cookie, won't this do the same?

boffbowsh: It only cleared it as it is expired. The newly-set one won't be expired. `visit root_path` could be any path. Server isn't clearing the cookie, it's just forcing the Capybara driver to check which cookies are still valid.

jamiecobbett: OK, that makes sense. Sorry to be a pain, but could you add a comment to line 165 to say that setting the cookie here won't have an expiry (or has an expiry in the future)?

boffbowsh: Done and rebased

PR #402, opened 2015-09-23
Don't load simplecov unnecessarily

benlovell:
SimpleCov was running and producing coverage reports when I ran `rake`
in the development environment, which isn't good. Also, it wasn't
really configured properly for CI either.

The question is, does anyone care about coverage? If yes, merge this.

If no, I'll just blow it all away and forget it ever happened :boom:

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -1,4 +1,0 @@
-require 'simplecov-rcov'
-
-SimpleCov.start 'rails'
-SimpleCov.formatter = SimpleCov::Formatter::RcovFormatter
--- a/Gemfile
+++ b/Gemfile
@@ -67,8 +67,8 @@ group :test do
   gem 'mocha', '1.1.0', require: false
   gem 'webmock', '1.17.3'
   gem 'minitest', '5.1.0'
-  gem 'simplecov', '0.6.4'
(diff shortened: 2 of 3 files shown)


jamiecobbett:
I'm in favour of this or getting rid of it. Do others use the simplecov data?

benilovj:
:+1: to getting rid of it.

benlovell:
I'd prefer to see it gone too.

alext:
It is currently graphed in Jenkins. I don't know if anyone looks at the data, but I think Jenkins is configured to fail a build if it falls below a threshold (not a very high one admittedly - I've never seen an app failed for this reason).

Anyway, if we do remove it, we need to be sure to remove the corresponding config from Jenkins.

benlovell:
@alext @benilovj @jamiecobbett A quick update on this - I have to admit I had not spotted the `.simplecov` file that was being picked up in CI. I've now dropped that (it felt a little magic) and instead elected to load simplecov when the requisite environment variable is exported.

Merging this as-is will ensure coverage is not run locally, but reported in CI as ever.

benlovell:
Closing in favour of #403 after discussing with @jamiecobbett.

PR #403, opened 2015-09-28
Removes coverage

benlovell:
Nobody cares about coverage and it doesn't come for free.

This may require some changes in Jenkins after merging.


Gemfile.lock
```diff
@@ -392,3 +384,6 @@ DEPENDENCIES
   webmock (= 1.17.3)
   whenever (~> 0.9.4)
   zeroclipboard-rails
+
+BUNDLED WITH
+   1.10.6
```
alext: This shouldn't be present for a ruby 2.2.2 app (we're running bundler 1.9.4 with ruby 2.2.2).

benlovell: To stop the `BUNDLED_WITH` nonsense?

alext: because 1.9.4 was the latest version when we rolled out 2.2.2 (2.2.3 is deployed with 1.10.6).

Assuming you're running on the dev VM, this should be all set up for you by puppet.

benlovell:
@alext Gotcha, addressed.

benlovell:
Thanks :heart_eyes:

PR #404, opened 2015-09-28
Bumps ruby and gems

benlovell:
Let's keep on top of the dependencies so it's easier when security fixes come
along, plus it means we get to use the latest and greatest features of ruby and
rails as they become available.

Details are in the individual commits. I kept the sidekiq bump separate as it was quite a jump and while it should be OK, I wanted to draw attention to it.


Gemfile
```diff
@@ -30,17 +30,17 @@ gem 'ancestry', '2.0.0'
...
 gem 'statsd-ruby', '1.1.0'
 gem 'unicorn', '4.9.0'
-gem 'sidekiq', '2.17.2'
-gem 'sidekiq-statsd', '0.1.2'
+gem 'sidekiq', '~> 3.5.0'
+gem 'sidekiq-statsd', '0.1.4'
 gem 'activejob-retry'
 
-gem 'redis', '3.0.6'
+gem 'redis', '~> 3.2.1'
```
jamiecobbett: Looking at the changelog for redis, they don't appear to be following semver: https://github.com/redis/redis-rb/blob/master/CHANGELOG.md

Gemfile
```diff
@@ -1,11 +1,11 @@
...
 
-gem 'rails', '4.2.3'
+gem 'rails', '4.2.4'
 
-gem 'kaminari', '0.16.1'
+gem 'kaminari', '~> 0.16.3'
 gem 'bootstrap-kaminari-views', '0.0.5'
 gem 'alphabetical_paginate', '2.2.3'
-gem 'mysql2', '0.3.18'
+gem 'mysql2', '~> 0.3.20'
```
jamiecobbett: Looking at their [changelog](https://github.com/brianmario/mysql2/releases), they don't seem to follow semver, and I'm wary of making it easy to upgrade a database driver unintentionally

Gemfile
benlovell: They don't officially declare support for semver but the changelog seems pretty solid to me unless I'm missing something obvious (as I have been known to do on occasion). They're stating they'll break compatibility with their next major bump to the 4.x series - but that's the point isn't it?

benlovell:
Closed in favour of #406 that isn't from my fork.

PR #405, opened 2015-09-28
Flag for 2SV

benlovell:
**Note:** This is branched off #404 which is still pending review.

Provides the ability for super admins to flag users as requiring 2SV. Not too much point at this moment in time but neatly paves the way for prompting flagged users to enrol in 2SV.


app/views/users/_form_fields.html.erb
```diff
@@ -41,6 +41,12 @@
   <%= f.select :organisation_id, organisation_options(f), organisation_select_options, { class: "chosen-select form-control", 'data-module' => 'chosen' } %>
 </p>
 
+<% if policy(User).flag_2sv? %>
+  <p class="form-group">
+    <%= f.check_box :requires_2sv %> <%= f.label :requires_2sv, "Requires 2SV" %>
```
jamiecobbett: I think the wording of this should be tweaked - for one thing we should say "2 step verification", not "2SV". @fofr do you have any thoughts?

fofr: (It should be "2-step verification", with the hyphen.)

Is there a screenshot of how this looks?
Who is using this form and when are they using it?

benlovell: I'll correct this and post a screenshot. This feature is currently available to super-admins when they create an invitation or edit an existing user.

Are there any feels regarding the attribute naming? Is `requires_2sv` and `flag_2sv?` OK or would you prefer `requires_two_step_verification` and `flag_two_step_verification?`?

benlovell:
Closing in favour of #407 

@jamiecobbett @fofr let's continue the discussion over there.

PR #406, opened 2015-09-29
Bumps ruby and various gems

benlovell:
Opening this in place of #404 that was against my fork.


benlovell:
I'm dealing with the broken build. :see_no_evil:

benlovell:
@jamiecobbett @benilovj have at it!

PR #407, opened 2015-09-29
Flag for 2SV

benlovell:
Provides the ability for super admins to flag users as requiring 2SV. Once they've been flagged they will be prompted to complete 2SV enrolment and cannot do much until they enrol or defer.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,7 +17,7 @@ class ApplicationController < ActionController::Base
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
   def after_sign_in_path_for(resource)
-    if resource.require_2sv?
+    if resource.prompt_for_2sv?
       prompt_two_step_verification_path
     else
       super
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,6 +16,14 @@ class ApplicationController < ActionController::Base
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
+  def after_sign_in_path_for(resource)
+    if resource.require_2sv?
+      prompt_two_step_verification_path
+    else
+      super
+    end
+  end
+
   def after_sign_out_path_for(resource_or_scope)
     new_user_session_path
   end
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -54,7 +54,10 @@ class Devise::TwoStepVerificationController < DeviseController
     @otp_secret_key = params[:otp_secret_key]
     totp = ROTP::TOTP.new(@otp_secret_key)
     if totp.verify(params[:code])
-      current_user.update_attribute(:otp_secret_key, @otp_secret_key)
+      current_user.update_attributes(
+        otp_secret_key: @otp_secret_key,
+        require_2sv: false
+      )
       EventLog.record_event(current_user, EventLog::TWO_STEP_ENABLED)
       redirect_to "/", notice: "2-step verification set up"
     else
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -1,11 +1,16 @@
 class Devise::TwoStepVerificationController < DeviseController
-  before_filter :prepare_and_validate, except: :prompt
+  before_filter :prepare_and_validate, except: [:prompt, :defer]
   skip_before_filter :handle_two_step_verification
 
   attr_reader :otp_secret_key
   private :otp_secret_key
 
   def prompt
+  end
+
+  def defer
+    current_user.defer_two_step_verification
+    redirect_to root_path, notice: 'You will not be required to setup 2-step verification.'
   end
 
   def show
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -55,10 +55,7 @@ class Devise::TwoStepVerificationController < DeviseController
     @otp_secret_key = params[:otp_secret_key]
     totp = ROTP::TOTP.new(@otp_secret_key)
     if totp.verify(params[:code])
-      current_user.update_attributes(
-        otp_secret_key: @otp_secret_key,
-        require_2sv: false
-      )
+      current_user.update_attribute(:otp_secret_key, @otp_secret_key)
       EventLog.record_event(current_user, EventLog::TWO_STEP_ENABLED)
       redirect_to_prior_flow(notice: "2-step verification set up")
     else
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -10,7 +10,7 @@ class Devise::TwoStepVerificationController < DeviseController
 
   def defer
     current_user.defer_two_step_verification
-    redirect_to root_path, notice: 'You will not be required to setup 2-step verification.'
+    redirect_to stored_location_for(:user) || :root
   end
 
   def show
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -1,9 +1,12 @@
 class Devise::TwoStepVerificationController < DeviseController
-  before_filter :prepare_and_validate
+  before_filter :prepare_and_validate, except: :prompt
   skip_before_filter :handle_two_step_verification
 
   attr_reader :otp_secret_key
   private :otp_secret_key
+
+  def prompt
+  end
 
   def show
   end
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -30,7 +30,8 @@ class Devise::TwoStepVerificationController < DeviseController
       warden.session(:user)['need_two_step_verification'] = false
       sign_in :user, current_user, bypass: true
       set_flash_message :notice, :success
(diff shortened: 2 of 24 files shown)


app/views/users/_form_fields.html.erb
```diff
@@ -41,6 +41,12 @@
   <%= f.select :organisation_id, organisation_options(f), organisation_select_options, { class: "chosen-select form-control", 'data-module' => 'chosen' } %>
 </p>
 
+<% if policy(User).flag_2sv? %>
+  <p class="form-group">
+    <%= f.check_box :requires_2sv %> <%= f.label :requires_2sv, "Requires 2SV" %>
```
benlovell: @jamiecobbett said:
>I think the wording of this should be tweaked - for one thing we should say "2 step verification", not "2SV". @fofr do >you have any thoughts?

@fofr said:

>(It should be "2-step verification", with the hyphen.)

>Is there a screenshot of how this looks?
>Who is using this form and when are they using it?

benlovell: I'll correct this and post a screenshot. This feature is currently available to super-admins when they create an invitation or edit an existing user.

Are there any feels regarding the attribute naming? Is `requires_2v` and `flag_2v?` OK or would you prefer `requires_two_step_verification` and `flag_two_step_verification?`?

benlovell: @fofr 

screen shot 2015-09-29 at 14 10 42

fofr: Thanks for the screenshot :)

Can the checkbox be moved up, so it's between role and organisation. To begin with the need for 2SV is coupled with a user's role.

The `<p class="">` wrapper should also have the class "checkbox" as per the Bootstrap 3 pattern. This should stop the label from showing in bold.

benlovell: @fofr 

screen shot 2015-09-29 at 14 30 08

app/views/users/_form_fields.html.erb
```diff
@@ -36,6 +36,14 @@
   </p>
 <% end %>
 
+<% if policy(User).flag_2sv? %>
+  <p class="checkbox">
+    <%= f.label :requires_2sv do %>
+      <%= f.check_box :requires_2sv %> Require 2-step verification
```
jamiecobbett: To continue the discussion: I think I would have chosen "require_2sv" or "2sv_required", because we're requiring the user to have it, rather than the user requiring us to allow it. Let me know if that doesn't make sense.

app/views/users/_form_fields.html.erb
fofr: :+1: Thanks for making those changes. :smile:

app/views/users/_form_fields.html.erb
benlovell: @jamiecobbett addressed in b970620

benlovell:
This is now ready for some eyes. The history needs a rework, I'll do this after the approach has been OK'd. Preview:

screen shot 2015-10-01 at 14 09 25


/cc @jamiecobbett @boffbowsh @benilovj @binaryberry

fofr:
It looks like this is using an older version of the copy. The copy was updated in the prototype in this commit: https://github.com/alphagov/signonotron2/blob/752f32b827e6048b96564272afb689de2f907540/app/views/prototype/two_factor_prompt.html.erb

Sorry that wasn't more obvious.

benlovell:
@fofr Ah, this was linked off the trello card https://github.com/alphagov/signonotron2/compare/prototype-2fa#commitcomment-12642578

Not to worry, I'll get it corrected.

benlovell:
@fofr 

screen shot 2015-10-01 at 14 44 53

app/models/user.rb
```diff
@@ -35,6 +35,7 @@ class User < ActiveRecord::Base
          :password_archivable, # in signonotron2/lib/devise/models/password_archivable.rb
          :password_expirable   # in signonotron2/lib/devise/models/password_expirable.rb
 
+  validates :require_2sv, inclusion: [false], if: ->(u) { u.otp_secret_key.present? }
```
boffbowsh: This seems to ensure that `require_2sv` is set to false once 2SV has been set up for a user, but I don't see any code that sets this once 2SV is set up. This will also result in a confusing error message for admins if they try to turn this flag on for a user that already has 2SV set up.

benlovell: :see_no_evil: I missed the resetting of the flag. Good spot.

Addressed in 09b6bdd and 605023d

app/controllers/application_controller.rb
```diff
@@ -16,6 +16,14 @@ class ApplicationController < ActionController::Base
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
+  def after_sign_in_path_for(resource)
+    if resource.require_2sv?
```
boffbowsh: Now I think about it, I think it'd be better if this only triggered if 2SV is required and they haven't got it already set up. That way, if we flag someone as needing 2SV and they get it turned off at a later date (in an emergency, through a bug, as part of some future feature), we don't lose the information that this user is flagged as requiring 2SV.

So rather than changing the flag, just ignore it if 2SV is configured. This also protects against weird data failures or bugs where the 2SV flag could be turned back on once they have 2SV set up, leaving them stuck in a loop.

app/controllers/devise/two_step_verification_controller.rb
```diff
@@ -1,10 +1,18 @@
...
 
   attr_reader :otp_secret_key
   private :otp_secret_key
 
+  def prompt
+  end
+
+  def defer
+    current_user.defer_two_step_verification
+    redirect_to root_path, notice: 'You will not be required to setup 2-step verification.'
```
jamiecobbett: Two things here:
* this should redirect them to the place they were going to before they were asked to sign in (eg another app, or a specific page in signon). The root_path would only be used as a fallback. I think Devise handles this for a normal signin?
* chatting with @fofr he thinks we can leave the `notice` out

app/controllers/application_controller.rb
benlovell: Cool, addressed in b28cd91

app/controllers/devise/two_step_verification_controller.rb
benlovell: @jamiecobbett OK, I'll address these issues. Is it OK to leave deferral as a permanent thing for now? So if they defer, it will require a super-admin to flag them again. There is an audit trail via the event log recording, so perhaps it won't go unnoticed?

benlovell:
@jamiecobbett @boffbowsh @benilovj this will hopefully address the remaining issues. :checkered_flag:

app/controllers/devise/two_step_verification_controller.rb
```diff
@@ -49,9 +57,12 @@ def create
     @otp_secret_key = params[:otp_secret_key]
     totp = ROTP::TOTP.new(@otp_secret_key)
     if totp.verify(params[:code])
-      current_user.update_attribute(:otp_secret_key, @otp_secret_key)
+      current_user.update_attributes(
+        otp_secret_key: @otp_secret_key,
+        require_2sv: false
```
boffbowsh: I don't think we need to set this here, it should only be set to false for deferrals?

app/models/user.rb
```diff
@@ -35,6 +35,9 @@ class User < ActiveRecord::Base
          :password_archivable, # in signonotron2/lib/devise/models/password_archivable.rb
          :password_expirable   # in signonotron2/lib/devise/models/password_expirable.rb
 
+  validates :require_2sv,
```
boffbowsh: I don't think this validation is correct any more. Admins should be able to flag an already-enrolled user as requiring 2SV

app/models/user.rb
```diff
@@ -81,6 +84,25 @@ class User < ActiveRecord::Base
...
+    if deferred_2sv_at?
+      deferred_2sv_at < 24.hours.ago
+    else
+      super
+    end
+  end
+
+  def defer_two_step_verification
+    transaction do
+      self.require_2sv = false
```
boffbowsh: Ditto here, setting the `deferred_2sv_at` attribute should override the `require_2sv` flag. I feel `require_2sv` should be an intent from admins/higher ups that says "This user should, at some point, set up 2SV"

jamiecobbett: Agreed. It also says "and the user can't turn it off", should we add that functionality later.

app/controllers/devise/two_step_verification_controller.rb
jamiecobbett: Agreed.

app/models/user.rb
jamiecobbett: Agreed - though it is Superadmins, not Admins.

app/controllers/devise/two_step_verification_controller.rb
benlovell: Addressed in 0fce8a6.

app/models/user.rb
benlovell: b4ec25c :boom:

app/models/user.rb
benlovell: Resolved in ead81b2

benlovell:
@boffbowsh @jamiecobbett comments addressed. If this is OK I'll rework the history into something more concise.

boffbowsh:
:+1: @jamiecobbett want to review and merge?

jamiecobbett:
Really glad to see this merged, even if it was a bit painful getting there. Thanks for sticking with it @benlovell - I think I realise now just how much context the rest of us had built up, so it was quite a big ask for you to do this just after joining. Thank you :fireworks:

PR #410, opened 2015-10-02
Unlocking after too many 2SV attempts

boffbowsh:
https://trello.com/c/LYxVsMEb/114-unlocking-after-too-many-2sv-attempts-medium-ish

This is an implementation of the locking/unlocking requirement which piggy-backs on the devise locakble module. Instead of maintaining a separate locking behaviour for 2SV, we re-use the `User#lock_access!` and `User#unlock_access!` methods. 

This also refactors the locked account email to use the Devise unlock instructions built in email, reducing app-specific complexity. This requires setting up an unlock token for a user, but this isn't currently exposed in the email, so isn't usable by the user.

The potential downside to this method is that the user is locked out completely from logging in, if they enter 2SV incorrectly too many times, they are sent back to the email / passphrase screen. When they enter their credentials correctly, they are given an error message saying "Invalid email/passphrase". This is valid behaviour for an attacker attempting to guess a user's passphrase, but could be misleading for a user who has entered 2SV incorrectly.

This copy could be changed to something like:

"You have entered your credentials incorrectly, or your account may be locked for too many incorrect attempts"

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -179,8 +179,6 @@ class User < ActiveRecord::Base
   def lock_access!
     EventLog.record_event(User.find_by_email(self.email), EventLog::ACCOUNT_LOCKED)
     super
-
-    UserMailer.locked_account_explanation(self).deliver_later
   end
 
   def status
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -234,6 +234,11 @@ class User < ActiveRecord::Base
     second_factor_attempts_count.to_i >= MAX_2SV_LOGIN_ATTEMPTS.to_i
   end
 
+  def unlock_access! *args
+    super
+    update_attribute(:second_factor_attempts_count, 0)
+  end
+
 private
 
   # Override devise_security_extension for updating expired passwords
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -80,7 +80,6 @@ class User < ActiveRecord::Base
     end
   }
 
-
   def event_logs
     EventLog.where(uid: uid).order(created_at: :desc)
   end
@@ -176,8 +175,10 @@ class User < ActiveRecord::Base
   end
 
   # Override Devise::Model::Lockable#lock_access! to add event logging
-  def lock_access!
-    EventLog.record_event(User.find_by_email(self.email), EventLog::ACCOUNT_LOCKED)
+  def lock_access! opts = {}
+    event = locked_reason == :two_step ? EventLog::TWO_STEP_LOCKED : EventLog::ACCOUNT_LOCKED
+    EventLog.record_event(self, event)
+
     super
   end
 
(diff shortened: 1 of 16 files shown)


app/models/user.rb
```diff
@@ -177,11 +175,16 @@ def update_stats
...
+  def lock_access! opts = {}
+    event = case locked_reason
+            when :passphrase
+              EventLog::ACCOUNT_LOCKED
+            when :two_step
+              EventLog::TWO_STEP_LOCKED
+            end
+
+    EventLog.record_event(self, event)
+    super opts
```
benlovell: Calling `super` as a bare word passes the enclosing arguments by default. If you want to golf this down a little.

app/models/user.rb
```diff
@@ -177,11 +175,16 @@ def update_stats
   end
 
   # Override Devise::Model::Lockable#lock_access! to add event logging
-  def lock_access!
-    EventLog.record_event(User.find_by_email(self.email), EventLog::ACCOUNT_LOCKED)
-    super
-
-    UserMailer.locked_account_explanation(self).deliver_later
+  def lock_access! opts = {}
+    event = case locked_reason
```
benlovell: Nitpicking here, but is there any rationale for a `case` statement here but a regular conditional in `#locked_reason`?

app/models/user.rb
```diff
@@ -213,18 +216,34 @@ def authenticate_otp(code)
...
+      :passphrase
+    end
+  end
+
   def max_2sv_login_attempts?
     second_factor_attempts_count.to_i >= MAX_2SV_LOGIN_ATTEMPTS.to_i
   end
 
+  def unlock_access! *args
+    super *args
```
benlovell: Same here. It's definitely a stylistic thing though.

db/migrate/20151001095709_add_unlock_token_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddUnlockTokenToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :unlock_token, :string, length: 64
+    add_index :users, :unlock_token
```
benlovell: Could/should this be a unique too?

app/models/user.rb
boffbowsh: Only in that I wanted to be explicit about the cases and have nil as the default.

benlovell: :+1:

boffbowsh: Good point.

boffbowsh:
Comments addressed, rebased.

Added unlock copy too, @fofr

jamiecobbett:
One minor comment, otherwise :+1:

bradwright:
Could you clarify what copy is shown when in the flow? We want to avoid enumeration attacks where possible.

boffbowsh:
@bradwright the same copy is currently used for "Invalid email/passphrase" and "Locked due to too many failed attempts", be it attempts for passphrase or 2SV

benlovell:
LGTM :shipit:

PR #413, opened 2015-10-07
Refactor 2SV sign in routes

boffbowsh:
Moves the 2SV sign in routes into a new controller and makes them more
like Devise, with the concept of a 2SV session. `two_step_verification_session#new` and
`#create` are now used for displaying the 2SV code entry screen
and validating the code respectively.

This frees up the `two_step_verification` resource itself to allow `#update`
to be used to change the device used for 2SV.

/cc @jamiecobbett @benilovj @fofr @benlovell


app/controllers/devise/two_step_verification_session_controller.rb
```diff
@@ -0,0 +1,32 @@
+class Devise::TwoStepVerificationSessionController < DeviseController
+  skip_before_action :handle_two_step_verification
+  def new
```
benlovell: Maybe a line of whitespace would help to separate these elements?

app/controllers/devise/two_step_verification_session_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def create
```
benlovell: Maybe out of scope for this particular PR, but this action really could do with a tidy up. The logic here is getting a little gnarly. :eyes:

boffbowsh: Good point. I don't feel like tackling it right now though.

benlovell:
LGTM! :shipit:

app/views/devise/two_step_verification_session/new.html.erb
```diff
@@ -0,0 +1,30 @@
+<% content_for :title, "2-step verification" %>
```
jamiecobbett: Should we get rid of `app/views/devise/two_step_verification_session/show.html.erb`?

boffbowsh: Good point, done.

jamiecobbett:
:top: refactoring :+1:

PR #414, opened 2015-10-08
[Do not merge] Preserve the referring host for successful sign in redirect

boffbowsh:
Using `store_location_for` strips the hostname from the provided
location before storing it in the session, meaning the app will
eventually redirect to a relative URL after a successful sign in. This
meant that users weren’t redirected to their desired application if
they had been redirect to signon from the app.

The new `store_full_location_for` method is based on the original
`store_location_for` implementation: http://bit.ly/1Nps7U6, and
additionally ensures the host is part of the GOV.UK parent domain as
provided by Plek

Added tests to verify this behaviour.

/cc @benilovj @jamiecobbett

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -42,4 +42,14 @@ class ApplicationController < ActionController::Base
   def redirect_to_prior_flow(args = {})
     redirect_to stored_location_for(:user) || :root, args
   end
+
+  def store_full_location_for(resource_or_scope, location)
+    session_key = stored_location_key_for(resource_or_scope)
+    uri = parse_uri(location)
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -42,4 +42,14 @@ def user_not_authorized(exception)
   def redirect_to_prior_flow(args = {})
     redirect_to stored_location_for(:user) || :root, args
   end
+
+  def store_full_location_for(resource_or_scope, location)
```
benlovell: I reckon something of this nature would be accepted upstream to devise. A thought for the future, perhaps :wink:

benlovell:
Seems legit, save for the build failures of course.

boffbowsh:
Closing in favour of #415 and a larger fix

PR #417, opened 2015-10-09
Add ability for users to change the device used for 2SV

boffbowsh:
This merges the existing create flow with the update flow, as it’s not
currently possible to add 2SV devices, only replace them. The result is
the update of the user’s OTP key, either from nil to a valid key thus
setting up 2SV, or from an old key to a new one, thus changing the
device they’ve got set up.

The same form is displayed for both cases, the differences are in the
text of the logged event, and the flash messages on success. The flows
are differentiated by the presence of the `otp_secret_key` on the user.

This commit also refactors the controller action responsible for
updating the secret by extracting the logic to verify the entered code
and update the secret.

screen shot 2015-10-09 at 15 50 23

/cc @fofr @jamiecobbett @benlovell

### DIFF
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -5,25 +5,19 @@ class Devise::TwoStepVerificationController < DeviseController
   attr_reader :otp_secret_key
   private :otp_secret_key
 
-  def new
-    if current_user.otp_secret_key.present?
-      redirect_to root_path, alert: "2-step verification is already set up"
-    else
-      @otp_secret_key = ROTP::Base32.random_base32
-    end
+  def show
+    generate_secret
   end
 
-  def create
-    @otp_secret_key = params[:otp_secret_key]
-    totp = ROTP::TOTP.new(@otp_secret_key)
-    if totp.verify(params[:code])
-      current_user.update_attribute(:otp_secret_key, @otp_secret_key)
-      EventLog.record_event(current_user, EventLog::TWO_STEP_ENABLED)
-      redirect_to "/", notice: "2-step verification set up"
+  def update
+    mode = current_user.has_2sv? ? :change : :setup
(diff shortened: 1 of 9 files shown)


app/controllers/devise/two_step_verification_controller.rb
```diff
@@ -5,25 +5,19 @@ class Devise::TwoStepVerificationController < DeviseController
...
-    totp = ROTP::TOTP.new(@otp_secret_key)
-    if totp.verify(params[:code])
-      current_user.update_attribute(:otp_secret_key, @otp_secret_key)
-      EventLog.record_event(current_user, EventLog::TWO_STEP_ENABLED)
-      redirect_to "/", notice: "2-step verification set up"
+  def update
+    mode = current_user.has_2sv? ? :change : :setup
+    if verify_code_and_update
+      EventLog.record_event(current_user, success_event_for(mode))
+      redirect_to "/", notice: I18n.t("devise.two_step_verification.messages.success.#{mode}")
```
benlovell: `:root` as a named path would be better than `"/"` now.

app/controllers/devise/two_step_verification_controller.rb
```diff
@@ -5,25 +5,19 @@ class Devise::TwoStepVerificationController < DeviseController
...
     else
-      EventLog.record_event(current_user, EventLog::TWO_STEP_ENABLE_FAILED)
+      EventLog.record_event(current_user, failure_event_for(mode))
       flash.now[:invalid_code] = "Sorry that code didn’t work. Please try again."
-      render :new, status: 422
+      render :show, status: 422
```
benlovell: Using the named status `:unprocessable_entity` might be cleaner now.

fofr:
Design, copy and markup look good :+1:

benlovell:
:ship:

jamiecobbett:
:+1:

PR #419, opened 2015-10-12
Key the 2SV session cookie on a hash of the secret

boffbowsh:
Adding a hash of the secret key into the signed 2SV session cookie
means that if a user changes their 2SV device, they will need to enter
their new code on subsequent logins on all devices.

/cc @benlovell @jamiecobbett @benilovj


benlovell:
Rubocop did a sad.

PR #420, opened 2015-10-13
Permit super admins to disable 2SV for users

benlovell:
This gives super admins the ability to disable 2SV for selected users. It's
pretty naive in that we don't actually check the user is enrolled when
attempting to disable 2SV.

Also snuck in a couple of small and semi-related refactoring commits in 4789de6
and 265b927

@boffbowsh @benilovj @jamiecobbett et al.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -121,10 +121,10 @@ class UsersController < ApplicationController
     end
   end
 
-  def disable_two_step_verification
-    @user.disable_2sv!(current_user)
+  def reset_two_step_verification
+    @user.reset_2sv!(current_user)
 
-    redirect_to :root, notice: '2-step verification is now disabled'
+    redirect_to :root, notice: '2-step verification is now reset'
   end
 
   private
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -122,7 +122,9 @@ class UsersController < ApplicationController
   end
 
   def disable_two_step_verification
-    render text: 'WIP'
+    @user.disable_2sv!
+
+    redirect_to :root, notice: '2-step verification is now disabled'
   end
 
   private
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -122,7 +122,7 @@ class UsersController < ApplicationController
   end
 
   def disable_two_step_verification
-    @user.disable_2sv!
+    @user.disable_2sv!(current_user)
 
     redirect_to :root, notice: '2-step verification is now disabled'
   end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -121,6 +121,10 @@ class UsersController < ApplicationController
     end
   end
 
+  def disable_two_step_verification
+    render text: 'WIP'
+  end
+
   private
 
   def load_and_authorize_user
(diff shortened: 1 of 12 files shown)


benlovell:
@fofr I put it in the links bar before the well and also added a standard JS based confirmation. Hopefully this is good.

screen shot 2015-10-13 at 17 30 46

fofr:
@benlovell I think the two mechanisms around 2SV: [ ] Requires 2SV and the "Disable 2SV" link will be confusing as they're currently presented.

benlovell:
@fofr I agree. Do you have any thoughts or suggestions on how we can improve?

app/models/event_log.rb
```diff
@@ -21,6 +21,7 @@ class EventLog < ActiveRecord::Base
   EMAIL_CHANGE_INITIATED = "Email change initiated"
   EMAIL_CHANGE_CONFIRMED = "Email change confirmed"
   TWO_STEP_ENABLED = "2-step verification enabled"
+  TWO_STEP_DISABLED = "2-step verification disabled"
```
fofr: Could this message be explicit that it was disabled by a superadmin?

benlovell: It could, but on reflection it might be better to set the `initiator` so that the actual super admin who initiated the action was logged as part of the event?

fofr: Yes, that would be better :+1:

benlovell: Addressed in da31b4e

benlovell:
@fofr @boffbowsh 

When not set up
----------
not set up

When set up
-----------
set up

app/models/event_log.rb
boffbowsh: I feel like `disabled` should be `reset`, as that's the action that admins are clicking on. Sorry for pedantry :/

app/models/user.rb
```diff
@@ -11,6 +11,7 @@ class User < ActiveRecord::Base
 
   SUSPENSION_THRESHOLD_PERIOD = 45.days
   UNSUSPENSION_GRACE_PERIOD = 3.days
+  TWO_FACTOR_PROMPT_THRESHOLD_PERIOD = 24.hours
```
fofr: Could this be two_step rather than factor, just for consistency?

app/models/event_log.rb
benlovell: No you're right. I missed this.

app/views/users/_form_fields.html.erb
```diff
@@ -37,11 +37,32 @@
...
-  </p>
+  <dl>
+    <dt>Account security</dt>
+    <dd>
+      <% if f.object.has_2sv? %>
+        <p>2-step verification enabled</p>
+        <ul>
+          <li>
+            <%= link_to 'Reset 2-step verification',
+              disable_two_step_verification_user_path(@user),
```
fofr: Can this be `reset` instead? `reset_two_step_verification_user_path`

app/controllers/users_controller.rb
```diff
@@ -121,6 +121,12 @@ def update_passphrase
     end
   end
 
+  def disable_two_step_verification
+    @user.disable_2sv!(current_user)
```
fofr: Should this be reset_2sv rather than disable? @boffbowsh do you have any thoughts on this?

fofr:
New UI looks good. A couple of minor comments regarding calling it `disable` or `reset` in the code.

fofr:
:fireworks:

PR #421, opened 2015-10-14
Don't try to validate 2SV session cookie if 2SV not enabled

boffbowsh:
Actually check the result of `user.need_two_step_verification?` instead
of checking if we have the method on the user. This was left over from
the `two_factor_authentication` gem’s implementation of the hook, and
wasn’t replaced.

This was causing errors when users had a 2SV session cookie but then
had their secret key reset. Hashing `nil` causes an error. If we check
`need_two_step_verification?` first, we can bypass this.

/cc @jamiecobbett


jamiecobbett:
> Actually check the result of user.need_two_step_verification? instead
of checking if we have the method on the user.

Ouch. Good spot!

PR #422, opened 2015-10-15
Bugfix: data-sync-complete job

surminus:
If this job is run multiple times then redirects risk ending up
incorrect:

```
Migrating Specialist publisher - redirect_uri to new domain:
https://specialist-publisher.staging.staging.publishing.service.gov.uk/auth/gds/callback
```

Note the inclusion of `staging` twice in the URL.

This change seeks to fix that even if the job is run multiple times.


boffbowsh:
Good catch :+1:

jamiecobbett:
This actually breaks the OAuth interaction because the `home_uri` gets updated, but the `redirect_uri` doesn't.

PR #423, opened 2015-10-15
Remove `o-tron` references from readme

fofr:
We call the app “Signon” (one word, no hyphens at the moment).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Sign-on-o-Tron II
+# Signon
 
 This is a centralised OAuth 2-based single sign-on provider for GDS services.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,8 +13,7 @@ Details of our interpretation of OAuth are provided in
 
 The application has Rake tasks to create new users and client applications.
 
-To create a new client application to which Sign-on-o-Tron will provide sign-on
-services:
+To create a new client application which will provide sign-on services:
```
alext: This new wording isn't quite correct. It's suggesting that the client application will provide the sign-on services, when it's actually this application that provides them, and the client application uses them.

alext:
One small comment on the wording, otherwise this looks good to me.  I'm looking forward to the point where we can rename this repo.

README.md
fofr: Whoops. Fixed.

PR #424, opened 2015-10-16
Protect the 2SV session controller from unauthed users

boffbowsh:
This controller wasn’t protected from unauthenticated users, meaning
that if a user landed on the page without a session their
`current_user.authenticate_otp` call would fail, causing an error. The
correct behaviour is to redirect the user to the login path.

We have to use `force: true` in this call because the controller is a
Devise controller, which are excluded from authentication by default.


benlovell:
:ship:

PR #426, opened 2015-10-26
Improve error messaging and track user errors on forms to Google Analytics

fofr:
Collect anonymous metrics on the different types of user error when they’re changing their email or passphrase, or when setting up 2-step verification.

* Add a helper for tracking messages in analytics
* Start tracking the error messages in analytics

This also includes some improvements to the passphrase change forms:

* Avoid using the default devise error message. The default includes a poor heading and old markup. The default can be found here: https://github.com/plataformatec/devise/blob/7df57d5081f9884849ca15e4fde179ef164a575f/app/helpers/devise_helper.rb
* Consolidate errors into a new partial
* DRY up form by moving submit button into partial

## Before
pasted_image_at_2015_10_08_04_03_pm_720

## After
screen shot 2015-10-23 at 13 20 58

cc @boffbowsh @jamiecobbett

### DIFF
--- a/app/views/devise/passwords/_change_password_panel.html.erb
+++ b/app/views/devise/passwords/_change_password_panel.html.erb
@@ -64,3 +64,7 @@
     </div>
   </div>
 </div>
+
+<div>
+  <%= f.submit t("users.edit.change"), class: 'btn btn-primary add-top-margin' %>
+</div>
--- a/app/helpers/analytics_helper.rb
+++ b/app/helpers/analytics_helper.rb
@@ -1,0 +1,13 @@
+module AnalyticsHelper
+  def analytics_attributes_alert(type, message)
+    safe_message = flash_text_without_email_addresses(message)
+
+    "data-module=\"auto-track-event\"\
+     data-track-action=\"alert-#{type}\"\
+     data-track-label=\"#{safe_message}\"".html_safe
+  end
+
+  def analytics_attributes_alert_danger(message)
+    analytics_attributes_alert('danger', message)
+  end
+end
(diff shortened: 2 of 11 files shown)


boffbowsh:
You've got red on you[r commit].

app/views/devise/passwords/_change_password_panel.html.erb
```diff
@@ -64,3 +64,7 @@
     </div>
   </div>
 </div>
+
+<div>
+  <%= f.submit t("users.edit.change"), :class => 'btn btn-primary add-top-margin' %>
```
benlovell: Could we use the 1.9 style hash keys here, per the style guide?

benlovell:
Looks good otherwise. I'm looking into the build failures as they're happening against SSO on other branches/master too. :disappointed:

app/views/devise/passwords/_change_password_panel.html.erb
fofr: Good point, updated and rebased.

tijmenb:
This is super nice. Minor suggestion, maybe using the `content_tag` helper could prevent you from having to do the string concatenation / `html_safe` things.

```ruby
def track_analytics_data(type, message)
  {
    'module' => 'auto-track-event',
    'track-action' => "alert-#{type}",
    'track-label' => flash_text_without_email_addresses(message)
  }
end
```

```ruby
<%= content_tag :li, data: track_analytics_data(:danger, message) %>
  <%= message %>
<% end %>
```

That could apply to https://github.com/alphagov/signonotron2/blob/track-more-errors/app/views/shared/_bootstrap_flash_messages.html.erb too.

fofr:
@tijmenb Ooh, that's much nicer.

benlovell:
Also, you can avoid yielding to the block in `content_tag` and simplify this a teeny bit by passing `message` as the second argument.

fofr:
@tijmenb @benlovell Implemented that suggestion in ae612d4, looking much nicer.

tijmenb:
:+1:

PR #428, opened 2015-10-28
Add password reset error to event log

tijmenb:
Currently on 2nd line, there's a Zendesk ticket from someone who is having trouble resetting their password:

> it just comes up with the message that I may of used the password before (which I definitely haven’t) or that the link is old even though I have just received the link

We'd like to investigate, but the account access log just displays:

`11:56am, 28 Oct 2015	 Passphrase reset attempt failure`

Which isn't very helpful, as there can be multiple different causes for the user to not be `valid?` ([see the controller](https://github.com/alphagov/signonotron2/blob/c581c8afc4cb09058a3c653c9aa49f604dec2fb6/app/controllers/passwords_controller.rb#L17)). 

By adding the activerecord error messages to `trailing_message` we can get a lot more information and can help users better.

Discuss:

- [x] Good idea?
- [x] Will changing the `trailing_message` type into `text` cause performance issues?

### DIFF
--- a/app/controllers/passwords_controller.rb
+++ b/app/controllers/passwords_controller.rb
@@ -43,6 +43,7 @@ private
   end
 
   def record_password_reset_failure(user)
-    EventLog.record_event(user, EventLog::PASSPHRASE_RESET_FAILURE)
+    message = "(errors: #{user.errors.full_messages.join(', ')})".truncate(255)
+    EventLog.record_event(user, EventLog::PASSPHRASE_RESET_FAILURE, trailing_message: message)
   end
 end
--- a/test/integration/event_log_test.rb
+++ b/test/integration/event_log_test.rb
@@ -63,8 +63,11 @@ class EventLogIntegrationTest < ActionDispatch::IntegrationTest
     visit edit_user_password_path(reset_password_token: token_received_in_email)
 
     click_on "Change passphrase"
+    event_log = @user.event_logs.first
 
-    assert_equal EventLog::PASSPHRASE_RESET_FAILURE, @user.event_logs.first.event
+    assert_equal EventLog::PASSPHRASE_RESET_FAILURE, event_log.event
+    assert_match "Passphrase can't be blank", event_log.trailing_message
+    assert_match "Passphrase not strong enough", event_log.trailing_message
   end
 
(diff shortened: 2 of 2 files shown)


app/controllers/passwords_controller.rb
```diff
@@ -43,6 +43,7 @@ def record_reset_page_loaded
   end
 
   def record_password_reset_failure(user)
-    EventLog.record_event(user, EventLog::PASSPHRASE_RESET_FAILURE)
+    message = "(errors: #{user.errors.full_messages.join(', ')})"
```
benlovell: `to_sentence` may be preferable to `join` here.

benlovell:
Looks good to me other than my small comment :star2: Just needs a rebase and we're on.

boffbowsh:
I'm not sure about the changing of the column. This table is already very big in production, and runs on MySQL. So changing a varchar to a blob will be pretty bad for performance and storage. There is currently a ticket in our backlog (https://trello.com/c/zjBz0vTu/205-revisit-event-logging-in-signon-for-all-user-actions-in-2sv-process) to address the event log. Can I add this case to the ticket instead.

boffbowsh:
Perhaps in the meantime we should log this to the production log in a way that is searchable in Kibana?

app/controllers/passwords_controller.rb
tijmenb: Tried that, but it looked weird:

`join`:

> Errors: Passphrase can't be blank, Passphrase not strong enough. Consider adding a number, symbols or more letters to make it stronger.

vs

`to_sentence`:

> Errors: Passphrase can't be blank and Passphrase not strong enough. Consider adding a number, symbols or more letters to make it stronger.

In the second case it looks like the errors belong together.

benlovell: Ah, that is a bit weird.

tijmenb:
@boffbowsh yeah, I came to the same conclusion on the column. I've removed the migration add will truncate the message coming in. There shouldn't be too many cases where the errors have more than 255 chars, and even then the data will be useful.

boffbowsh:
Thanks @tijmenb :)

PR #431, opened 2015-11-03
Fix password reset flow

tijmenb:
The password reset flow is currently broken. It should work like this:

1. When users click on a time-expired reset password link, they should see an error message
2. When users click on a previously used or invalid reset password link, they should see an error message
3. When users provide an previously used password, too-weak password or two different passwords, they should see validation errors

Currently there is no validation on the token validity or expiry when the password reset page is loaded, and when the user provides an invalid input, they will see a generic, possibly incorrect error page instead of the password form with errors.

This means users can click on an expired password link and think everything is okay, input a new password and see a message that their link is invalid. Also, users can input an incorrect password and get the message "It may be that you’ve used this passphrase already, or the link you were using has expired or was invalid". We've had several Zendesk requests from people who couldn't change their passwords.

For example: https://govuk.zendesk.com/agent/tickets/1176278

This commit fixes the flow, and updates the integration tests to cover the scenarios.

This commit does not introduce new behaviour, this is the original flow before a change made related to the big Devise upgrade of this summer (930c07c for example).

## Before

screen shot 2015-11-03 at 10 27 01

Visiting the page with an incorrect token.

screen shot 2015-11-03 at 10 26 53

Typing in a too weak password.

## After

screen shot 2015-11-03 at 10 22 45

Visiting the page with an incorrect token.

screen shot 2015-11-03 at 10 23 39

Typing in a too-weak password.

/cc @fofr

### DIFF
--- a/app/controllers/passwords_controller.rb
+++ b/app/controllers/passwords_controller.rb
@@ -21,8 +21,8 @@ class PasswordsController < Devise::PasswordsController
       end
     end
   end
+protected
 
-  protected
   def after_resetting_password_path_for(resource)
     signed_in_root_path(resource)
   end
@@ -37,13 +37,16 @@ private
   end
 
   def record_reset_page_loaded
-    token = Devise.token_generator.digest(self, :reset_password_token, params[:reset_password_token])
-    user_from_params = User.find_by(reset_password_token: token)
     EventLog.record_event(user_from_params, EventLog::PASSPHRASE_RESET_LOADED) if user_from_params
   end
 
   def record_password_reset_failure(user)
     message = "(errors: #{user.errors.full_messages.join(', ')})".truncate(255)
     EventLog.record_event(user, EventLog::PASSPHRASE_RESET_FAILURE, trailing_message: message)
+  end
+
+  def user_from_params
+    token = Devise.token_generator.digest(self, :reset_password_token, params[:reset_password_token])
+    User.find_by(reset_password_token: token)
   end
 end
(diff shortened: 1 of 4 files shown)


boffbowsh:
Looks good to me from a code POV. Has copy been checked?

fofr:
It's just the one line of copy that's changed as far as I can see, and that clarifies the situation in which the "Try again" error message appears.

app/controllers/passwords_controller.rb
```diff
@@ -2,6 +2,16 @@ class PasswordsController < Devise::PasswordsController
   before_filter :record_password_reset_request, only: :create
   before_filter :record_reset_page_loaded, only: :edit
 
+  def edit
+    super
+
+    user = user_from_params
+    unless user && user.reset_password_period_valid?
+      render 'devise/passwords/reset_error'
+      return
```
benlovell: Is this necessary?

app/controllers/passwords_controller.rb
```diff
@@ -16,13 +26,12 @@ def update
...
         record_password_reset_failure(resource) if resource.persisted?
-        render 'devise/passwords/reset_error'
-        return
       end
     end
   end
 
-  protected
+protected
+
```
benlovell: Style guide says we should remove this line of whitespace.

tijmenb:
I've updated this PR.

benlovell:
Code looks good. :ship:

jamiecobbett:
This looks awesome, thanks @tijmenb

PR #433, opened 2015-11-05
Prompt future admins and superadmins to set up 2SV

boffbowsh:
https://trello.com/c/Xh7kLY4z/150-prompting-all-current-and-future-admins-and-super-admins-to-set-up-2sv-in-production-only-small

When a user is promoted to an admin or superadmin, they should have the `require_2sv` flag set.

This new behaviour caused a lot of tests to break, so first the test authentication helpers were refactored to make this transition easier.

Best reviewed commit-by-commit.

### DIFF
--- a/test/helpers/user_helpers.rb
+++ b/test/helpers/user_helpers.rb
@@ -1,24 +1,23 @@
 module UserHelpers
-  def signin(user_or_options)
-    email = user_or_options.is_a?(Hash) ? user_or_options[:email] : user_or_options.email
-    password = user_or_options.is_a?(Hash) ? user_or_options[:password] : user_or_options.password
+  def signin_with(user = nil, email: nil, password: nil, second_step: true, set_up_2sv: true)
+    user ||= User.find_by(email: email)
+    email ||= user.email
+    password ||= user.password
+
+    if user && user.require_2sv? && user.otp_secret_key.blank? && set_up_2sv
+      user.update_attribute(:otp_secret_key, ROTP::Base32.random_base32)
+    end
 
     fill_in "Email", with: email
     fill_in "Passphrase", with: password
     click_button "Sign in"
-  end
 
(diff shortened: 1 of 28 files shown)


benilovj:
:+1: looks very good; the only thing that's not clear is why 8a4cb27 is needed...

boffbowsh:
Added some more details.

test/helpers/user_helpers.rb
```diff
@@ -1,24 +1,23 @@
 module UserHelpers
-  def signin(user_or_options)
-    email = user_or_options.is_a?(Hash) ? user_or_options[:email] : user_or_options.email
-    password = user_or_options.is_a?(Hash) ? user_or_options[:password] : user_or_options.password
+  def signin_with(user = nil, email: nil, password: nil, second_step: true, set_up_2sv: true)
+    user ||= User.find_by(email: email)
+    email ||= user.email
+    password ||= user.password
+
+    if user && user.require_2sv && user.otp_secret_key.blank? && set_up_2sv
```
benlovell: Since it forms part of a boolean expression, it would be more idiomatic to express `require_2sv` in the predicate form.

test/models/user_test.rb
```diff
@@ -8,8 +8,53 @@ def setup
     @user = create(:user)
   end
 
-  test "`require_2sv` defaults to false" do
-    refute build(:user).require_2sv
+  context "#require_2sv" do
+    should "default to false for normal users" do
+      refute create(:user).require_2sv
```
benlovell: The same applies here and elsewhere in this test where the predicate form is not used. Stylistic nit-pick, soz.

benlovell:
Nice. Looks good other than the minor nit-picks. :heart_eyes:

jamiecobbett:
:fireworks:

PR #434, opened 2015-11-05
Allow time drift when setting up 2SV

jamiecobbett:
When you sign in, we allow codes from devices which have a time which lags
behind the time on the server. This is because user devices are frequently
not automatically updated with the correct time.

This fixes a bug where we didn't allow codes from such devices at setup time.

### DIFF
--- a/test/integration/two_step_verification_test.rb
+++ b/test/integration/two_step_verification_test.rb
@@ -94,7 +94,7 @@ class TwoStepVerificationTest < ActionDispatch::IntegrationTest
       end
 
       should "accept a valid code from a device which has a small time lag" do
-        old_code = Timecop.freeze(35.seconds.ago) { ROTP::TOTP.new(@new_secret).now }
+        old_code = Timecop.freeze(29.seconds.ago) { ROTP::TOTP.new(@new_secret).now }
 
         Timecop.freeze do
           fill_in "code", with: old_code
--- a/test/integration/two_step_verification_test.rb
+++ b/test/integration/two_step_verification_test.rb
@@ -92,6 +92,17 @@ class TwoStepVerificationTest < ActionDispatch::IntegrationTest
           assert SUCCESS, last_email.subject
         end
       end
+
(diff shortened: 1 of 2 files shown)


test/integration/two_step_verification_test.rb
```diff
@@ -92,6 +92,17 @@ class TwoStepVerificationTest < ActionDispatch::IntegrationTest
           assert SUCCESS, last_email.subject
         end
       end
+
+      should "accept a valid code from a device which has a small time lag" do
+        old_code = Timecop.freeze(35.seconds.ago) { ROTP::TOTP.new(@new_secret).now }
```
jamiecobbett: I am confident about the code change, but I couldn't totally explain the behaviour I saw when writing the tests:
* this offset would fail reliably with the old implementation and pass with the changes
* When I increased it to 45 seconds, it would fail with the new implementation

fofr:
Good catch. :+1:

benilovj:
:+1: merge when green.

test/integration/two_step_verification_test.rb
fofr: Did it consistently fail, or was it intermittent?

boffbowsh:
Hold off on merging. Full reply shortly

test/integration/two_step_verification_test.rb
boffbowsh: There will be points where this will fail with 35 seconds as well. The drift only allows for 30 seconds, so if this test runs in the first 5 seconds of the 30 second 2SV window, it'll fail. This needs to be 30 seconds max, or we should change the `MAX_2SV_DRIFT_SECONDS` to something else.

boffbowsh:
Another quick check of this please, I've changed the test to check with a 29-second-old code.

jamiecobbett:
Ideally 49610e5 would have a bit of explanation, but I don't feel strongly about it. Merge when green.

PR #437, opened 2015-11-11
Run `EventLog` validations

benlovell:
These validations were being silently ignored. To workaround the broken
behaviour we generate the `uid` earlier in the lifecycle, ensure
validations fail loudly via `create!` and move the event logging for
'successful passphrase change' to the correct place.

Once the `uid` was being set (thus passing validation) this event was being
logged continually, whereas before it would fail silently.


benilovj:
LGTM :+1: merge when green

PR #438, opened 2015-11-12
Log and persist events by ID

benlovell:
Trello: https://trello.com/c/zjBz0vTu

Declare and persist events by identifiers, rather than by their string
representations.

These changes will allow the `event_id` and `event` attributes/columns
to coexist until such time a migration is ran to convert the existing
`event` column values to their respective `event_id`s.

Some of the changes to support this:

* Override `event` reader and attempt to map message via `EVENTS`
* Add the `event_id` attribute for persistence
* Default `event` to an empty string so it is no longer mandatory
* Don't validate `event` when an `event_id` is provided
* Fixup broken specs that referred to the `event` directly
* Removed a redundant spec that was well covered elsewhere
* Validate presence of `event_id`
* Ensure `event_id` corresponds with `EVENTS` entry

A follow-up PR is due shortly to populate the `event_id`s of existing rows via
mapping from the `event`. After this we can dispense with the `event` column.

### DIFF
--- a/db/migrate/20151112110911_add_event_id_to_event_logs.rb
+++ b/db/migrate/20151112110911_add_event_id_to_event_logs.rb
@@ -1,0 +1,5 @@
+class AddEventIdToEventLogs < ActiveRecord::Migration
+  def change
+    add_column :event_logs, :event_id, :integer
+  end
+end
--- a/test/models/event_log_test.rb
+++ b/test/models/event_log_test.rb
@@ -1,8 +1,22 @@
 require 'test_helper'
 
 class EventLogTest < ActiveSupport::TestCase
+  context "#event" do
+    context "when the event has an `event_id`" do
+      should "return the correctly mapped event description" do
+        assert_equal "Account suspended", EventLog.new(event_id: EventLog::ACCOUNT_SUSPENDED.id).event
+      end
+    end
+
+    context "when the event has no `event_id`" do
+      should "return the correctly mapped event description" do
+        assert_equal "Account suspended", EventLog.new(event: EventLog::ACCOUNT_SUSPENDED.description).event
+      end
+    end
+  end
+
   test "can create a valid eventlog" do
-    assert EventLog.new(uid: :uid, event: :event).valid?
+    assert EventLog.new(uid: :uid, event_id: EventLog::TWO_STEP_ENABLED.id).valid?
   end
 
   test "requires a user uid" do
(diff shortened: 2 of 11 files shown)


jamiecobbett:
My apologies if this has already been discussed, but would this work better as a set of objects with properties such as `id`, `message`, `requires_initiator?` and so on? At the moment there's the possibility of an event type being added without a message, for example.

benlovell:
@jamiecobbett That's a good point. I think we can cover this by beefing up the validation on `event_id`. Give me a few moments.

benlovell:
@jamiecobbett d3ba8dc should guard against that particular circumstance. I agree that perhaps relying on constants and a hash lookup isn't the _cleanest_ possible approach, but thought it good enough for something stable that isn't core to the application.

db/migrate/20151112110911_add_event_id_to_event_logs.rb
```diff
@@ -0,0 +1,6 @@
+class AddEventIdToEventLogs < ActiveRecord::Migration
+  def change
+    add_column :event_logs, :event_id, :integer
+    change_column :event_logs, :event, :string, default: ''
```
jamiecobbett: In production, this table has over 7 million rows. I'm a bit nervous about this migration (and related ones in other PRs) potentially locking the table and blocking all user actions that write to it. Perhaps I'm misremembering the effects of `ALTER TABLE` statements though?

If that is the case, we'll want to schedule these deploys carefully for times of lower usage.

benlovell: Indeed! Against my VM database, with ~500k rows it took a good few minutes to complete.

boffbowsh: Would be better to not add the default here, and instead make the code carry on filling the `event` column with the event's description in order to maintain backwards compatibility.

The ADD COLUMN will still block, but should only be for a few seconds(?). Needs testing on staging, but an alternative solution might be to use https://www.percona.com/doc/percona-toolkit/2.1/pt-online-schema-change.html

test/models/event_log_test.rb
```diff
@@ -1,8 +1,16 @@
 require 'test_helper'
 
 class EventLogTest < ActiveSupport::TestCase
+  context "#event" do
+    context "when the event has an `event_id`" do
```
boffbowsh: Can you add the opposite case as well please?

db/migrate/20151112110911_add_event_id_to_event_logs.rb
benlovell: Good call. Amended.

PR #440, opened 2015-11-13
Populate `event_id` attributes from existing events

benlovell:
This is based on #438.

Simple rake tasks to populate the newly created `event_id` column via mappings from `event`.

### DIFF
--- a/lib/event_id_populator.rb
+++ b/lib/event_id_populator.rb
@@ -1,0 +1,7 @@
+class EventIdPopulator
+  def self.populate
+    EventLog::EVENTS.each do |event|
+      EventLog.where(event: event.description, event_id: nil).update_all(event_id: event.id)
+    end
+  end
+end
--- a/lib/tasks/event_logs.rake
+++ b/lib/tasks/event_logs.rake
@@ -14,4 +14,13 @@ namespace :event_logs do
       puts "All events were mapped"
     end
   end
+
+  desc 'Fix remaining `event_id`s for renamed entries'
+  task fix_remaining_event_ids: :environment do
+    events = EventLog.where(event_id: nil, event: 'Account locked')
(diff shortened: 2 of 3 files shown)


lib/event_id_populator.rb
```diff
@@ -0,0 +1,7 @@
+class EventIdPopulator
+  def self.populate
+    EventLog::EVENTS.each do |event|
+      EventLog.where(event: event.description, event_id: nil).update_all(event_id: event.id)
```
boffbowsh: Will this cover all the events? Have some of the events changed their description?

benlovell: The only renamed one I found is covered here: https://github.com/alphagov/signonotron2/pull/440/files#diff-0fe12a07e85e0e2a9814daf5085e6653R18

benlovell:
@boffbowsh Title changed to reflect newfound mergeability! :wink:

db/schema.rb
```diff
@@ -45,9 +45,9 @@
   add_index "batch_invitations", ["outcome"], name: "index_batch_invitations_on_outcome", using: :btree
 
   create_table "event_logs", force: :cascade do |t|
-    t.string   "uid",              limit: 255, null: false
-    t.string   "event",            limit: 255, null: false
-    t.datetime "created_at",                   null: false
+    t.string   "uid",              limit: 255,              null: false
+    t.string   "event",            limit: 255, default: "", null: false
```
boffbowsh: I don't think this is correct any more. Shouldn't be any schema changes

benlovell: :scream: rebase fail.

benlovell:
Closing this in favour of #443 since this was targeting `revisit-event-logging`.

PR #441, opened 2015-11-16
[DO NOT MERGE] Remove `event` column

benlovell:
This is based on #438. 

Removes the now redundant `event` column, as this has been superseded by
the `event_id` and virtual `event` attribute and mapping.

This should be merged/deployed only after #438 and #440 but is otherwise good for review.


benlovell:
Closing in favour of #445

PR #442, opened 2015-11-18
Standardise on integration instead of preview for logic

boffbowsh:
https://trello.com/c/2xcA8uQT/173-change-preview-only-2sv-behaviour-to-include-new-integration-environment

Uses the new `Plek.environment` to detect preview/integration for 2SV's preview-specific behaviour.

Todo:

- [ ] Plek gem bump dance (https://github.com/alphagov/plek/pull/39)

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -4,7 +4,7 @@ class UserMailer < Devise::Mailer
 
   default from: Proc.new { email_from }
 
-  helper_method :suspension_time, :account_name, :instance_name, :locked_time, :unlock_time, :preview?
+  helper_method :suspension_time, :account_name, :instance_name, :locked_time, :unlock_time, :production?
 
   def two_step_reset(user)
     @user = user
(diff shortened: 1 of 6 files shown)


app/mailers/user_mailer.rb
```diff
@@ -92,7 +92,7 @@ def subject_for(key)
       default: [:subject, key.to_s.humanize], app_name: app_name)
   end
 
-  def preview?
-    GovukAdminTemplate.environment_label == "Preview"
+  def integration?
+    Plek.environment == "integration"
```
benlovell: These string comparisons will proliferate forever. Maybe we could clean these up a bit by having Plek answer to `Plek.environment.{integration?,production?,development?,other?}`

boffbowsh: I kinda want to keep Plek fairly dumb and not start wrapping things in `ActiveSupport::StringInquirer`. Open to opinions on this however.

benlovell: Makes sense. One option as we'd discussed would be the `defined?(ActiveSupport)` dance or maybe we could just place the string comparison checks in one canonical predicate in this app? So at least then we don't repeat the `Plek.environment == "..."` things in various spots.

boffbowsh:
Closed in favour of #444

PR #444, opened 2015-11-19
Auto-flag adminish users for 2SV in production only

boffbowsh:
https://trello.com/c/2xcA8uQT/173-change-preview-only-2sv-and-wider-signon-behaviour-to-include-new-integration-environment

A new integration environment is being introduced to eventually replace
preview, so logic is being switched to only auto-flag users in
production environment.

This behaviour should eventually be controlled by a specific
configuration variable in line with 12 factor, but we need to get this
updated quickly in order to work with the new environment.

Additionally, this now depends on `instance_name` directly in line with
other logic in this app. For production, `instance_name` is `nil`.

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -4,7 +4,7 @@ class UserMailer < Devise::Mailer
 
   default from: Proc.new { email_from }
 
-  helper_method :suspension_time, :account_name, :instance_name, :locked_time, :unlock_time, :preview?
+  helper_method :suspension_time, :account_name, :instance_name, :locked_time, :unlock_time, :production?
 
   def two_step_reset(user)
     @user = user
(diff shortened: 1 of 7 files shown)


boffbowsh:
@jamiecobbett @benlovell

app/mailers/user_mailer.rb
```diff
@@ -92,7 +92,7 @@ def subject_for(key)
       default: [:subject, key.to_s.humanize], app_name: app_name)
   end
 
-  def preview?
-    GovukAdminTemplate.environment_label == "Preview"
+  def production?
+    Rails.application.config.instance_name.blank?
```
jamiecobbett: We made the right pragmatic decision here, but relying on `instance_name` being blank feels like it might bite us in the future.

PR #445, opened 2015-11-20
Remove `event` column

benlovell:
Removes the now redundant `event` column, as this has been superseded by
the `event_id` and virtual `event` attribute and mapping.

Opening in favour of #441 that targeted the wrong branch.


boffbowsh:
:+1: but will need applying manually with `pt-online-schema-change`

benlovell:
The build did a sad. Just looking into it now.

jamiecobbett:
Assuming the tests are sorted, let's merge and deploy on Monday.

PR #446, opened 2015-11-20
Clean up after 2SV hard go live date

benlovell:
Trello: https://trello.com/c/1gH5NTC6

2SV went live today so we can clean up the cruft. Removes deferral, go
live date checks and simplifies 2SV prompt logic.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,8 +11,6 @@ class User < ActiveRecord::Base
 
   SUSPENSION_THRESHOLD_PERIOD = 45.days
   UNSUSPENSION_GRACE_PERIOD = 3.days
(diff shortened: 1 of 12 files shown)


app/models/user.rb
```diff
@@ -84,31 +82,10 @@ class User < ActiveRecord::Base
...
-    end
-  end
-
-  def defer_two_step_verification
-    transaction do
-      touch(:deferred_2sv_at)
-
-      EventLog.record_event(self, EventLog::TWO_STEP_PROMPT_DEFERRED)
-    end
+    require_2sv?
```
boffbowsh: Could just be `require_2sv? && !has_2sv?`

I don't know if that's more or less readable.

benlovell: I don't think so, but I'm open to opinion!

PR #455, opened 2015-12-09
Allow superadmins to see users for app

tijmenb:
Team Finding Things has created a new app (Content tagger). People who have access to Panopticon should also have access to this app.

The process of providing access will be greatly helped by having a list of users that have access to panopticon. This PR adds a page for each application that lists the users that (in the past) have had access to an app.

When there are more than 100 users, pagination will be shown.

Trello: https://trello.com/c/pWDaETqu

### DIFF
--- a/app/controllers/doorkeeper_applications_controller.rb
+++ b/app/controllers/doorkeeper_applications_controller.rb
@@ -17,6 +17,11 @@ class DoorkeeperApplicationsController < ApplicationController
     end
   end
 
+  def users_with_access
+    authorized_users_user_ids = @application.authorized_tokens.select(:resource_owner_id)
+    @users = policy_scope(User).where(id: authorized_users_user_ids).includes(:organisation).page(params[:page]).per(100)
+  end
+
   private
 
   def load_and_authorize_application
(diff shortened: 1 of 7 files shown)


app/controllers/doorkeeper_applications_controller.rb
```diff
@@ -17,6 +17,11 @@ def update
     end
   end
 
+  def users_with_access
+    authorized_users_user_ids = @application.authorized_tokens.select(:resource_owner_id)
+    @users = policy_scope(User).where(id: authorized_users_user_ids).includes(:organisation).page(params[:page]).per(100)
```
jamiecobbett: I think this would be _users who have ever accessed an app_. This might include users who attempted to access an app but were denied access because they don't have the "signin" permission. 

I suspect what you're really after is to see the users who currently have the "signin" permission for a given app.

rboulton:
:sparkles: :+1: for intent!

app/controllers/doorkeeper_applications_controller.rb
tijmenb: Good to know! 

That won't be a straightforward query probably? It looks like I have to join `users`, `user_application_permissions` and `supported_permissions`. Am I on the right track?

tijmenb: @jamiecobbett this wasn't actually that complicated. See https://github.com/alphagov/signonotron2/commit/0c18c1b8a31f72d107db1f2601cb346c09c67bc4.

boffbowsh:
Nice :)

PR #456, opened 2015-12-11
Show user permissions in application user list

tijmenb:
Per feedback on https://trello.com/c/G9oupgco, this PR:

- Displays the permissions of the user on the application users page
- Sorts the users by last signin date

### DIFF
--- a/app/queries/users_with_access.rb
+++ b/app/queries/users_with_access.rb
@@ -1,0 +1,22 @@
+class UsersWithAccess
+  attr_reader :scope, :application
+
+  def initialize(scope, application)
+    @scope = scope
+    @application = application
+  end
+
+  def users
+    scope
+      .where(id: authorized_users_user_ids)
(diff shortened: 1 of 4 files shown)


app/queries/users_with_access.rb
```diff
@@ -0,0 +1,23 @@
+class UsersWithAccess
```
boffbowsh: I feel this would be neater as a scope on the `User` class. `queries` is a non-standard-Rails directory, other devs might not understand to look there.

tijmenb: I don't like putting this in the model because you end up with massive god models that know about everything in the system.

Is `app/something` not a reasonably established convention now? `app/presenters` and `app/policies` are also non-standard, but used here. Pretty sure `app/queries` is used on other projects as well.

jamiecobbett:
This is really nice, thanks @tijmenb!

PR #458, opened 2015-12-31
Bump admin gem to 4.0.1

fofr:
* Allow environment_style to be set to integration
* Use admin template config for app title

Some of the earlier dependencies on `environment_style` were refactored out in #444

cc @boffbowsh


app/views/layouts/application.html.erb
```diff
@@ -1,4 +1,3 @@
-<% content_for :app_title, 'GOV.UK Signon' %>
 <% content_for :page_title, "#{yield(:title)} | GOV.UK Signon" %>
```
benilovj: Should the page title still have the app name?

fofr: At the moment there isn't a pattern for doing this automatically. But I think there should be.

benlovell: @fofr will you act on this or should we get this merged?

fofr: I wasn't going to alter it as part of this PR. I think it's something that's needed in the admin template itself.

PR #460, opened 2016-01-06
Display 2SV set up/change device links

benlovell:
Displays the correct links depending on the current user's 2SV enrollment.
This opens 2SV up to all users on an opt-in basis.

### DIFF
--- a/app/views/root/index.html.erb
+++ b/app/views/root/index.html.erb
@@ -10,8 +10,8 @@
     <li><%= link_to "Change your email or passphrase", edit_email_or_passphrase_user_path(current_user) %></li>
     <% if current_user.has_2sv? %>
       <li><%= link_to "Change your 2-step verification phone", two_step_verification_path, class: "js-track" %></li>
-    <% elsif current_user.require_2sv? %>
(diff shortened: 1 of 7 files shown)


app/views/root/index.html.erb
```diff
@@ -10,8 +10,8 @@
     <li><%= link_to "Change your email or passphrase", edit_email_or_passphrase_user_path(current_user) %></li>
     <% if current_user.has_2sv? %>
       <li><%= link_to "Change your 2-step verification phone", two_step_verification_path, class: "js-track" %></li>
-    <% elsif current_user.require_2sv? %>
-      <li><%= link_to "Set up 2-step verification", two_step_verification_path, class: "js-track" %></li>
+    <% else %>
+      <li><%= link_to "Make your Signon account more secure", two_step_verification_path, class: "js-track" %></li>
```
fofr: I think _Signon_ here is redundant.

benlovell:
@fofr Yep, agree. Addressed in 41dccee.

fofr:
:+1:

PR #463, opened 2016-01-27
Make addressable an explicit dependency of signon

boffbowsh:
Addressable was only being loaded as a test dependency,
but being used in runtime code. This meant that the gem
was not being required in production, and uses of
`Addressable::URI` were resulting in errors.

/cc @benlovell


boffbowsh:
:+1: for @benlovell's

PR #464, opened 2016-01-29
Log to STDOUT and STDERR in production mode

boffbowsh:
To bring this more in line with 12-facgtor principals, this changes the
logging to go to STDOUT(json), and STDERR(everything else).

Prior art:
https://github.com/alphagov/government-frontend/pull/80
https://github.com/alphagov/publishing-api/pull/109

Related puppet change: https://github.com/alphagov/govuk-puppet/pull/4023

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -50,6 +50,7 @@ Signonotron2::Application.configure do
 
   # Use a different logger for distributed setups.
   # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
+  config.logger = ActiveSupport::TaggedLogging.new(Logger.new($stderr))
 
   # Use a different cache store in production.
   # config.cache_store = :mem_cache_store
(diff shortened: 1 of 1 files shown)


surminus:
:+1: merge when green

config/environments/production.rb
```diff
@@ -74,9 +75,11 @@
 
   # Disable automatic flushing of the log to improve performance.
   # config.autoflush_log = false
+  $real_stdout = $stdout.clone
```
jamiecobbett: Does this need to be a global?

jamiecobbett:
I have a feeling that we did something special with Signon's logs for security reasons. This may be a red herring, but @bradwright do you remember?

bradwright:
We only collect them in Kibana in Integration right now. That's controlled
in Puppet, so we'd need to check this behaviour will carry over (but we
also agreed with accreditors to start collecting them everywhere, so maybe
we shouldn't bother checking).

On Fri, 29 Jan 2016 at 16:51, Jamie Cobbett <notifications@github.com>
wrote:

> I have a feeling that we did something special with Signon's logs for
> security reasons. This may be a red herring, but @bradwright
> <https://github.com/bradwright> do you remember?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/alphagov/signonotron2/pull/464#issuecomment-176857443>
> .
>
-- 
Bradley Wright

Head of Technology, Digital Group
Government Digital Service
@bradwright

config/environments/production.rb
boffbowsh: I can't think why it would. Changed.

PR #466, opened 2016-02-02
support PostgreSQL

macool:
via env
```
SIGNONOTRON2_DB_ADAPTER=postgresql
```

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,29 +1,31 @@
-development: &development
-  adapter: mysql2
-  encoding: utf8
-  reconnect: false
-  database: signonotron2_development
-  pool: 5
-  username: signonotron2
-  password: signonotron2
+<% db_adapter = ENV.fetch("SIGNONOTRON2_DB_ADAPTER", "mysql") %>
 
-# Warning: The database defined as "test" will be erased and
-# re-generated from your development database when you run "rake".
-# Do not set this db to the same as development or production.
-test: &test
-  adapter: mysql2
+default: &default
   encoding: utf8
   reconnect: false
-  database: signonotron2_test
-  pool: 5
   username: signonotron2
   password: signonotron2
+  pool: 5
+
+mysql_default: &mysql_default
+  <<: *default
+  adapter: mysql2
+
+postgresql_default: &postgresql_default
+  <<: *default
+  adapter: postgresql
+  host: localhost
+
+development:
+  <<: *<%= db_adapter %>_default
+  database: signonotron2_development
+
+test: &test
+  <<: *<%= db_adapter %>_default
+  database: signonotron2_test
 
 production:
-  adapter: mysql2
-  encoding: utf8
-  reconnect: false
+  <<: *<%= db_adapter %>_default
   database: halberd_production
-  pool: 5
   username: halberd
   password:
--- a/db/migrate/20140114144500_signin_permissions_must_be_lowercase.rb
+++ b/db/migrate/20140114144500_signin_permissions_must_be_lowercase.rb
@@ -1,5 +1,6 @@
 class SigninPermissionsMustBeLowercase < ActiveRecord::Migration
   def up
-    SupportedPermission.update_all({ name: 'signin' }, { name: 'Signin' })
+    SupportedPermission.where(name: 'Signin')
+                       .update_all(name: 'signin')
   end
 end
--- a/db/migrate/20140115153833_update_api_user_in_users.rb
+++ b/db/migrate/20140115153833_update_api_user_in_users.rb
@@ -1,8 +1,8 @@
 class UpdateApiUserInUsers < ActiveRecord::Migration
   def up
     User.joins(:authorisations)
-        .where('oauth_access_tokens.expires_in > ?', 5.years.to_s)
-        .group(:email)
+        .where('oauth_access_tokens.expires_in > ?', 5.years.to_i)
+        .group(:id, :email)
         .update_all(api_user: true)
   end
 
--- a/db/migrate/20150212133251_fix_user_update_permission.rb
+++ b/db/migrate/20150212133251_fix_user_update_permission.rb
@@ -1,5 +1,6 @@
 class FixUserUpdatePermission < ActiveRecord::Migration
   def up
-    SupportedPermission.update_all({ grantable_from_ui: false }, { name: 'user_update_permission' })
+    SupportedPermission.where(name: 'user_update_permission')
+                       .update_all(grantable_from_ui: false)
   end
 end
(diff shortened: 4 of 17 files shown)


benilovj:
Hi @macool, thank you very much for this.

What's the thinking behind using `structure.sql` instead of `schema.rb` for the Postgres adapter? Are there things needed for the Postgres adapter that cannot be expressed in `schema.rb` (that's the only reason why it was done this way in the [Support API](https://github.com/alphagov/support-api/pull/63))?

db/schema.rb
```diff
@@ -16,8 +16,8 @@
   create_table "batch_invitation_application_permissions", force: :cascade do |t|
     t.integer  "batch_invitation_id",     limit: 4, null: false
     t.integer  "supported_permission_id", limit: 4, null: false
-    t.datetime "created_at",                        null: false
-    t.datetime "updated_at",                        null: false
+    t.datetime "created_at"
```
benilovj: why are these (and other not-null constraints) being changed?

config/database.yml
```diff
@@ -1,29 +1,31 @@
-development: &development
-  adapter: mysql2
+<% db_adapter = ENV["SIGNONOTRON2_DB_ADAPTER"].nil? ? "mysql" : ENV["SIGNONOTRON2_DB_ADAPTER"] %>
```
benlovell: This could be tidied up a little by fetching `ENV.fetch("SIGNONOTRON2_DB_ADAPTER", "mysql")`

db/migrate/20140114144500_signin_permissions_must_be_lowercase.rb
```diff
@@ -1,5 +1,6 @@
 class SigninPermissionsMustBeLowercase < ActiveRecord::Migration
   def up
-    SupportedPermission.update_all({ name: 'signin' }, { name: 'Signin' })
+    SupportedPermission.where({ name: 'Signin' })
```
benlovell: May as well remove the unnecessary braces around the options as well, if we're tidying this up. :+1:

db/migrate/20140115153833_update_api_user_in_users.rb
```diff
@@ -1,8 +1,8 @@
 class UpdateApiUserInUsers < ActiveRecord::Migration
   def up
     User.joins(:authorisations)
-        .where('oauth_access_tokens.expires_in > ?', 5.years.to_s)
-        .group(:email)
+        .where('oauth_access_tokens.expires_in > ?', 5.years.to_i.to_s)
```
benlovell: The `to_s` happens implicitly so it can be removed.

db/migrate/20150212133251_fix_user_update_permission.rb
```diff
@@ -1,5 +1,6 @@
 class FixUserUpdatePermission < ActiveRecord::Migration
   def up
-    SupportedPermission.update_all({ grantable_from_ui: false }, { name: 'user_update_permission' })
+    SupportedPermission.where({ name: 'user_update_permission' })
```
benlovell: Same here with removing the braces around both sets of options.

lib/signonotron2/db_adapter.rb
```diff
@@ -0,0 +1,23 @@
...
+    SUPPORTED_DBS.each do |db_adapter|
+      define_method "#{db_adapter}?" do
+        adapter == db_adapter
+      end
+    end
+
+    private
+
+    def adapter
+      @adapter ||= if ENV["SIGNONOTRON2_DB_ADAPTER"].blank?
```
benlovell: You can repeat the `fetch` trick here to tidy this up a little.

lib/signonotron2/db_adapter.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    SUPPORTED_DBS = %w(mysql postgresql).freeze
+
+    SUPPORTED_DBS.each do |db_adapter|
+      define_method "#{db_adapter}?" do
+        adapter == db_adapter
+      end
+    end
+
+    private
```
benlovell: Our style dictates we outdent the access modifier to the same level as `class`.

lib/signonotron2/db_adapter.rb
```diff
@@ -0,0 +1,23 @@
+module Signonotron2
+  class DbAdapter
+    include Singleton
+
+    SUPPORTED_DBS = %w(mysql postgresql).freeze
+
+    SUPPORTED_DBS.each do |db_adapter|
```
benlovell: I wonder if meta-programming these predicates actually saves us any code here... Maybe worth just defining them trivially?

jamiecobbett:
If we're doing this properly, we should have the CI build set up to run signon in both configurations.

benlovell:
@jamiecobbett Seems there are [changes to `jenkins.sh`](https://github.com/alphagov/signonotron2/pull/466/files#diff-55dd59ac7badaee5e6fa9e2a41407df7R19) to support this. Is there more to do?

macool:
thanks @benlovell and @benilovj for your comments, I'm reviewing each one of these

jamiecobbett:
@benlovell oops - you're right, thanks.

Have chatted with @daibach and he's happy for us to do this. If supporting both databases becomes problematic, we may back this change out.

db/schema.rb
macool: @benilovj I'm not sure why but running migrations from version 0 gave me this schema :confused: . But I'll run `schema:load` instead

macool:
@benilovj I'm using `structure.sql` for postgresql because running migrations using postgresql changes my `schema.rb`; postgresql version doesn't include `limit` option for integers and enables `plpgsql` extension. Please take a look at the diff: https://gist.github.com/macool/d713f87c1acbf1b234bc

macool:
@benilovj @benlovell I've amended and force pushed for you to review. thanks in advance

boffbowsh:
This looks really good to me. Thanks for your efforts in this, will save us a huge bunch of work in the future. Not looking forward to the actual migration though.

I've pushed the branch to our repo so that CI will run.

benlovell:
```
bundle exec govuk-lint-ruby --format html --out rubocop-8052cd2fd96932ad76604ba8373a4feefec831a8.html --format clang app test lib spec config
config/initializers/alphabetical_paginate_config.rb:10:1: C: Favor modifier if usage when having a single-line body. Another good alternative is the usage of control flow &&/||.
if Signonotron2.db_adapter.mysql?
^^

209 files inspected, 1 offence detected
```

macool:
@benlovell I've addressed this offence and amended

benlovell:
@macool No worries! This is a great contribution. Thanks :heart_decoration:

benlovell:
@macool Just a heads up to say this hasn't stalled rather it requires a little attention in alphagov/govuk-puppet that I'm currently addressing.

benlovell:
Closing in favour of #467.

PR #468, opened 2016-02-11
Extract department specific text to locale file

dwhenry:
**NOTE:** This pull request is an example only and not fully implemented.

This is a proposed solution that would allow the PensionWise team to use the Signonotron code with different text at key points within the application without requiring multiple files to be changed within a branch.

This would make merging changes from the main application easy as only the locale file would be significantly different.

### DIFF
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -27,7 +27,7 @@ class Devise::TwoStepVerificationController < DeviseController
   end
 
   def otp_secret_key_uri
-    issuer = "GOV.UK%20Signon"
+    issuer = I18n.t('devise.issuer')
     if Rails.application.config.instance_name
       issuer = "#{Rails.application.config.instance_name.titleize}%20#{issuer}"
     end
--- a/app/mailers/mailer_helper.rb
+++ b/app/mailers/mailer_helper.rb
@@ -4,18 +4,14 @@ module MailerHelper
   end
 
   def app_name
-    if instance_name.present?
-      "GOV.UK Signon #{instance_name}"
-    else
-      "GOV.UK Signon"
-    end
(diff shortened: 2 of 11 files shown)


boffbowsh:
It might be better to follow [12 factor config](http://12factor.net/config) principles here and allow the Issuer (and other variables) to be overridden in the environment. Then the change is a deploy-time concern.

dwhenry:
Hi Paul,

Total agree, apart from the size, would estimate about 50-60 bit of text
that could be different, this size could be significantly reduces depending
on how words are grouped. I still feel that using environment variables
would become cumbersome, but happy to investigate further and give
actual numbers - as mine are just gut feel.

David

boffbowsh:
Fair point. So I guess you'd convert everything to I18n format and then replace the locale file at deploy time? If so, :+1: for the approach

dwhenry:
@boffbowsh @benlovell I'm not happy with the view changes (e534dd) I have made.

I'm going to create a separate pull request without any view changes and then look to refactor the mailer views to use a layout - reducing duplication - which will hopefully put me in a better place for the view changes I want to make. 

Thoughts?

boffbowsh:
Sounds ok to me :+1:

app/controllers/devise/two_step_verification_controller.rb
```diff
@@ -27,7 +27,7 @@ def update
   end
 
   def otp_secret_key_uri
-    issuer = "GOV.UK%20Signon"
+    issuer = I18n.t('devise.issuer')
```
jamiecobbett: Won't this need to be URI encoded?

dwhenry: @jamiecobbett  you are correct - had done that in the locale file. This has been fixed in a separate branch https://github.com/alphagov/signonotron2/pull/471

PR #471, opened 2016-02-16
Extract department text phase 1

dwhenry:
Superceeds PR: https://github.com/alphagov/signonotron2/pull/468 without view changes

### DIFF
--- a/config/initializers/i18n.rb
+++ b/config/initializers/i18n.rb
@@ -1,0 +1,17 @@
+# this is required to catch errors thrown from calls to I18n directly - primarily mailers
+
+if Rails.env.test?
+  module I18n
+    class AlwaysRaiseExceptionHandler < ExceptionHandler
+      def call(exception, locale, key, options)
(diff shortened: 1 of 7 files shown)


config/initializers/i18n.rb
```diff
@@ -0,0 +1,17 @@
+# this is required to catch errors thrown from calls to I18n directly - primarily mailers
+
+if Rails.env.test?
+  module I18n
+    class AlwaysRaiseExceptionHandler < ExceptionHandler
+      def call(exception, locale, key, options)
+        if exception.is_a?(MissingTranslation)
+          raise exception.to_exception
```
benlovell: Does this preserve the original backtrace?

dwhenry: Hi Ben,

Sorry I missed this message. Yes, that is my understanding - and I believe that was the case when I used it.  Planning to do some more work on admin this afternoon so will double check then.

David

PR #472, opened 2016-02-17
Extract common user email components into layout file

dwhenry:


app/views/user_mailer/unlock_instructions.html.erb
```diff
@@ -1,5 +1,3 @@
-<p>Hello <%= @resource.name %>,</p>
```
benlovell: The layout refers to `@user` now. Is this going to cause a problem? I'm not sure about test coverage here, did you check?

app/views/user_mailer/unlock_instructions.text.erb
```diff
@@ -1,5 +1,3 @@
-Hello <%= @resource.name %>,
```
benlovell: Same here. These are overridden devise templates though I see we have them copied into the local view path under `devise/mailers`. It might be worth checking these are even being used. Again, coverage here is not great I believe.

dwhenry:
@benlovell I'm not 100% sure on those emails. They do render content - not sure how.

The implementation is such that I would expect them to use @resource instead of @user, still need to investigate further how they are actually working. 

Not sure if they are actually being used, they are not called directly but could be called from within devise itself. I will try to lock myself out of the application tomorrow and see if I get the emails.

Any help/thoughts would be appreciated.

benlovell:
@dwhenry I added 683d341 to address the issue I raised and closed this via #473. Thanks!

PR #467, opened 2016-02-06
Agnostic support for MySQL / PostgreSQL

benlovell:
I've opened this PR to supplant #466, kindly contributed by @macool.

Unfortunately, our CI server is not configured to build forks of our repos,
hence my opening this PR against the canonical, GDS fork.

We've been talking internally about migrating to postgres for some time and
this is definitely a huge step in the right direction.

PR #474, opened 2016-02-18
Replacement department name with locale lookup.

dwhenry:
This is the last piece of work to allow pension wise to merge our branch back in.

We still have some differences, which I will leave within our branch :

- Using Postgres instead of MySQL - existing branch attempting to solve this issue.
- Using Puma instead of Unicorn.
- Using Bugsnag instead of Airbrake.
- Adding Newrelic logging
- Using Rails12Factor instead of LogStasher
- Upgrade of SideKiq to v 4.0

Let me know if you think any should be merged in (or written to allow either to be used).

David

### DIFF
--- a/app/views/devise/mailer/invitation_instructions.html.erb
+++ b/app/views/devise/mailer/invitation_instructions.html.erb
@@ -1,6 +1,6 @@
 <p>Hello <%= @resource.name %>,</p>
 
-<p>Someone has invited you to GOV.UK Signon, you can accept it through the link below.</p>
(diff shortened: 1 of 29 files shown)


benlovell:
Can you reword b5a4ccd to include _why_ you're making this change?

benlovell:
Of those other differences you're maintaining:

* #467 is pending and will make us database-agnostic. 
* We'll eventually upgrade sidekiq

But the rest are not likely to be pulled upstream anytime soon.

dwhenry:
@benlovell updated commit message. 

Your comments on difference is inline with expectations.

benlovell:
Thanks :+1:

PR #475, opened 2016-02-18
Fix app_title display translation missing

jamiecobbett:
Since c79a5143d7dfcd8dd02e3913980cef879e269b2a, we were using Rails' I18n to
set the GovukAdminTemplate.app_title in an initializer. Unfortunately, whilst
the I18n code is available, it hasn't been configured at the time intializers
are loaded. This meant that users saw this in the page header:
```
  translation missing: en.department.name Signon
```
The tests didn't fail because the setting to make missing translations raise
errors (in config/environments/test.rb) is also loaded after initializers.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,3 +1,4 @@
+<% content_for :app_title, "#{t('department.name')} Signon" %>
 <% content_for :page_title, "#{yield(:title)} | #{t('department.name')} Signon" %>
 <% content_for :head do %>
   <%= csrf_meta_tags %>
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -1,3 +1,4 @@
+<% content_for :app_title, "#{I18n.t('department.name')} Signon" %>
 <% content_for :page_title, "#{yield(:title)} | #{t('department.name')} Signon" %>
```
fofr: One line has `I18n.t('department.name')`, the next `t('department.name')`, should these be consistent?

jamiecobbett: Good point - rebased.

benlovell:
@dwhenry This should be relevant to your interests.

PR #476, opened 2016-02-18
Fix emails not being sent

jamiecobbett:
Since ec36b330c44eecc9b345dccdd6daaf25f0a585bc, we have used I18n to get the
email_from value. Because initializers are loaded before I18n has been
configured with a load path, when the Devise.config.mailer_sender was set, it
received an I18n translation missing message, instead of the email address
string. This was then apparently used by Devise to set the Reply To field and
probably the From field.

When UserMailer and NoisyBatchInvitation mailer call I18n, it has had a load
path set and so override the From field with correct/valid values, but the
Reply To field isn't.

The result was that emails were probably being rejected either by our email
sending service (Amazon SES) or elsewhere, so emails weren't being sent to
users.

Reading the comment for Devise.config.mailer_sender, it isn't neccessary to set
it if you set a custom Devise.config.mailer, so I tried removing this call and
the reply_to was no longer set at all.

@dwhenry you might be affected by this bug.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -7,7 +7,7 @@ Devise.setup do |config|
   # ==> Mailer Configuration
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class with default "from" parameter.
-  config.mailer_sender = Class.new.extend(MailerHelper).email_from
+  # config.mailer_sender = Class.new.extend(MailerHelper).email_from
 
   # Configure the class responsible to send e-mails.
   config.mailer = "UserMailer"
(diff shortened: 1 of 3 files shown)


boffbowsh:
Tested this on Integration, emails are sent and received as expected :+1:

benlovell:
Well played @jamiecobbett 🌟

bradwright:
Good find!
On Fri, 19 Feb 2016 at 08:32, Ben Lovell <notifications@github.com> wrote:

> Well played @jamiecobbett <https://github.com/jamiecobbett> 🌟
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/alphagov/signonotron2/pull/476#issuecomment-186116873>
> .
>
-- 
Bradley Wright

Head of Technology, Digital Group
Government Digital Service
@bradwright

PR #477, opened 2016-02-19
Remove duplicate greetings in emails

boffbowsh:
The greetings and sign offs for emails are all set in the layout now,
so we should remove them from the individual email views.


jamiecobbett:
:+1: but will leave for a while in case anything else pops up.

benlovell:
:see_no_evil:

PR #481, opened 2016-03-03
Make tests agnostic of department name, email, etc.

dwhenry:
Remove department specific names from test suit to allow test to pass when an alternative department name is used.


dwhenry:
@benlovell - Any luck on getting this merged??

PR #482, opened 2016-03-08
Ignore dictionary file when searching via `ag`

benlovell:
Anytime I search this project the dictionary file comes up with hits
and this is annoying and wasteful.


boffbowsh:
I'm sorry your change to this file broke our tests because the wind was blowing in the wrong direction

benlovell:
:dizzy_face: :dash: :hankey:

PR #483, opened 2016-03-08
Update vendored 'chosen' css

gpeng:
Change `url` to `image_url` to get fingerprinted image paths. This fixes
404 for chosen-sprite.png (and @2x)

This is fixed in [the current version of harvesthq/chosen](https://github.com/harvesthq/chosen). @fofr Could we just upgrade to that? I chose to just fix in place for now to avoid any undesirables.

[Trello](https://trello.com/c/3LZxwXDw/161-delete-image-no-longer-appear-in-sign-on-after-carrenza-migration-small)

before

after


boffbowsh:
LGTM merge when :green_heart:

fofr:
Good to see this fixed :)

PR #484, opened 2016-03-17
Make test run agnostic of locale

dwhenry:
We have removed the GOV.UK specific text from the application so that we can use pension wise specific text on our server. This was done using a pension wise specific locale file.

In order to test that everything works as expected and all locale text has been populated we want to be able to run the tests suite with our alternative locale file.

As a result, I removed and GOV.UK specific text from the tests and used regular express matches instead.
- Remove reference to "digital.cabinet-office"
- Deal with from address having optional quotes
  around the name portion of the email if the name 
  contains any special characters.  In this case 
  the '.' in 'GOV.UK'

### DIFF
--- a/test/integration/batch_inviting_users_test.rb
+++ b/test/integration/batch_inviting_users_test.rb
@@ -22,7 +22,7 @@ class BatchInvitingUsersTest < ActionDispatch::IntegrationTest
       invited_user = User.find_by_email("fred@example.com")
       assert_not_nil invited_user
       assert invited_user.has_access_to?(application)
-      assert_equal "noreply-signon-development@digital.cabinet-office.gov.uk", last_email.from[0]
+      assert_match /noreply-signon-development@.+\.gov\.uk/, last_email.from[0]
       assert_equal nil, last_email.reply_to[0]
 
       assert_equal "fred@example.com", last_email.to[0]
--- a/test/integration/email_change_test.rb
+++ b/test/integration/email_change_test.rb
@@ -134,16 +134,20 @@ class EmailChangeTest < ActionDispatch::IntegrationTest
     end
 
     should "show an error and not send a confirmation if the email is blank" do
-      visit new_user_session_path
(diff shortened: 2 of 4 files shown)


benlovell:
What's the rationale for this and how do these changes make the tests agnostic of locale?

It seems those tests were passing before these changes, unless I'm missing something?

dwhenry:
HI Ben,

The rational is that we have removed the GOV.UK specific text from the application so that we can use pensionwise specific text ojn our server. we have achieve this using a locale file.

In order to test that everything works as expected and all locale text has been populated we want to be able to run the tests suite with our alternative locale file. 

As a result, I removed and GOV.UK specific text from the tests - mostly using regular express matches instead.

David

benlovell:
Hey @dwhenry. Thanks, you should rewrite your commit message with that as it paints a much clearer picture. :+1:

dwhenry:
@benlovell good point I will try and make the commit more descriptive of the thought process moving forward

dwhenry:
Any update on getting the PR merged?

test/integration/batch_inviting_users_test.rb
```diff
@@ -22,7 +22,7 @@ class BatchInvitingUsersTest < ActionDispatch::IntegrationTest
       invited_user = User.find_by_email("fred@example.com")
       assert_not_nil invited_user
       assert invited_user.has_access_to?(application)
-      assert_equal "noreply-signon-development@digital.cabinet-office.gov.uk", last_email.from[0]
+      assert_match /noreply-signon-development@.+\.gov\.uk/, last_email.from[0]
```
gpeng: I don't think we should be making this 'vague' in the integration tests. I appreciate that this approach is used in the unit tests but they are testing parts of the email address rather than the whole. Could we add another test that explicitly sets a different locale in the test and asserts that the locale config is being respected? Would that work for your use case @dwhenry ?

PR #486, opened 2016-04-06
Update text: managing editors can use support form

lutgendorff:
Email was stating that managing editors can un-suspend accounts. Managing editors don't have this permission.

Updated text to say that they can use the support form only.


fofr:
I think the HTML version needs updating too:
https://github.com/lutgendorff/signonotron2/blob/lutgendorff-patch-1-5/app/views/user_mailer/unlock_instructions.html.erb

PR #503, opened 2016-07-05
Change Redis namespace to match app name.

danielroseman:


elliotcm:
Good to merge when green.

PR #507, opened 2016-08-08
Improve cookies security

mgrassotti:
Trello: https://trello.com/c/DlBVF1KZ/426-improve-signon-cookie-security-small

Set the HttpOnly, Secure and SameSite=lax attributes for cookies, so 2-step-verification cookies could not be stolen by a man-in-the-middle attacker.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,6 +1,7 @@
 require File.expand_path('../boot', __FILE__)
 
 require 'rails/all'
+require_relative '../lib/same_site_security/middleware'
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
@@ -56,5 +57,7 @@ module Signonotron2
     config.autoload_paths << Rails.root.join('lib')
 
     config.active_job.queue_adapter = :sidekiq
+
+    config.middleware.insert_before 0, SameSiteSecurity::Middleware
   end
 end
--- a/lib/same_site_security/middleware.rb
+++ b/lib/same_site_security/middleware.rb
@@ -1,0 +1,20 @@
+module SameSiteSecurity
+  class Middleware
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      req = Rack::Request.new(env)
+      status, headers, response = @app.call(env)
+      if cookies = headers['Set-Cookie']
+        cookies = cookies.split("\n") unless cookies.is_a?(Array)
+
+        headers['Set-Cookie'] = cookies.map { |cookie|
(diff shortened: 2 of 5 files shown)


config/application.rb
```diff
@@ -56,5 +57,7 @@ class Application < Rails::Application
     config.autoload_paths << Rails.root.join('lib')
 
     config.active_job.queue_adapter = :sidekiq
+
+    config.middleware.insert_before 0, SameSiteSecurity::Middleware
```
boffbowsh: If you call this `SameSiteSecurityMiddleware` and put it in `lib/same_site_security_middleware.rb`, Rails will be able to autoload it without the `require_relative`

mgrassotti: Sorry, I am not sure I understand.
If I define `class SameSiteSecurityMiddleware ... end` in `lib/same_site_security_middleware.rb` 
and then I use 
```
config.middleware.insert_before 0, SameSiteSecurityMiddleware
```
I get `NameError: uninitialized constant Signonotron2::Application::SameSiteSecurityMiddleware`

boffbowsh: Perhaps `::SameSiteSecurityMiddleware` ? Either way, the class/module/file layout isn't correct. The convention is that nested classes are in directories, so `SameSiteSecurity::Middleware` should be `lib/same_site_security/middleware.rb`

mgrassotti: I think that could be because autoload paths (including `/lib`) are not yet loaded in application.rb.
Anyway I created a new the directory to comply with the module/class standard structure.

lib/same_site_security/middleware.rb
```diff
@@ -0,0 +1,20 @@
+module SameSiteSecurity
+  class Middleware
```
andrewgarner: We should add some test coverage for this class.

mgrassotti: 👍

spec/unit/same_site_security/middleware.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  context "when called with a GET request" do
+    let(:request) { Rack::MockRequest.new(subject) }
+
+    it "sets cookies attributes properly" do
+      env = Rack::MockRequest.env_for("/a-protected-url")
+      status, headers = subject.call(env)
+
+      cookies = headers['Set-Cookie']
+      expect(cookies).to include('SameSite=Lax')
```
boffbowsh: Can we also test that `_signonotron2_session=abcd` is left untouched please

mgrassotti: 👍

PR #509, opened 2016-08-19
Remove sprockets deprecation messages

benlovell:
After the recent Rails bumps these messages had appeared due to a
transitive bump to sprockets:

```
DEPRECATION WARNING: Sprockets method `register_engine` is deprecated.
Please register a mime type using `register_mime_type` then
use `register_compressor` or `register_transformer`.
```

Bumping sass-rails resolves this issue.


boffbowsh:
Pushed branch to run tests 👍

PR #510, opened 2016-08-30
Bump Ruby / Bundler dependencies

benlovell:
This is tied to alphagov/gds-sso#105. 

While it serves that particular purpose, it's probably a good idea to bring this project up to date with the latest Ruby. We've been running our production Pension Wise signon instance on 2.3.1 for around a month now without issue. Although perhaps there are alphagov-specific considerations to be had...

3389913 may be contentious (and _technically_ unnecessary) so feel free to drop it or ask me to, if necessary.

cc @boffbowsh


boffbowsh:
I'm going to rewrite the signon starter/stopper to work around this

PR #511, opened 2016-09-19
YAML secrets for Devise

alexmuller:
Depends on https://github.com/alphagov/govuk-puppet/pull/4877 and https://github.gds/gds/deployment/pull/1083 being in production.

I think this approach is safer than what we have at the moment.

### DIFF
--- a/config/devise.yml
+++ b/config/devise.yml
@@ -1,0 +1,11 @@
+development:
+  pepper: 'fake-pepper'
+  secret_key: 'fake-secret-key'
+
+test:
+  pepper: 'fake-pepper'
+  secret_key: 'fake-secret-key'
+
+production:
+  pepper: <%= ENV['DEVISE_PEPPER'] %>
+  secret_key: <%= ENV['DEVISE_SECRET_KEY'] %>
--- a/config/devise.yml
+++ b/config/devise.yml
@@ -5,6 +5,7 @@ development:
 test:
   pepper: 'fake-pepper'
   secret_key: 'fake-secret-key'
+  stretches: 1
 
 production:
   pepper: <%= ENV['DEVISE_PEPPER'] %>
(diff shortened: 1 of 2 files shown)


boffbowsh:
This will want a rebase against master

alexmuller:
Rebased, and this is ready to be reviewed now:

```
fab staging class:backend do:'cat /etc/govuk/signon/env.d/DEVISE_PEPPER'
fab staging class:backend do:'cat /etc/govuk/signon/env.d/DEVISE_SECRET_KEY'
fab production class:backend do:'cat /etc/govuk/signon/env.d/DEVISE_PEPPER'
fab production class:backend do:'cat /etc/govuk/signon/env.d/DEVISE_SECRET_KEY'
```

boffbowsh:
Looks like we're waiting on the new pepper being deployed.

alexmuller:
Interesting, that's working for me. Is it not for you?

boffbowsh:
The pepper is the same in staging and production, is that correct? But I have now just realised the point of the other PRs is to introduce the env var in the first place, not to change them. So, if they're expected to be the same, 👍

alexmuller:
Yep, that's right. It has to be the same because it's used to encrypt passwords which are synced with the database overnight.

alexmuller:
Tests aren't running here for some reason, will see what's going on.

alexmuller:
https://ci.dev.publishing.service.gov.uk/job/govuk_gds_sso_signon_branches/544/console is still failing for some reason.

alexmuller:
In order to get the gds-sso tests to pass here I had to temporarily change the gds-sso branch name on CI to use alphagov/gds-sso@d2885dfc9cff1c923c0c4e4acb936e031df41d73.

PR #513, opened 2016-09-19
[Do not merge] Grant user permissions for manuals publisher

Rosa-Fox:
- Find users with permissions for specialist_publisher and grant them with those same permissions for manuals publisher


boffbowsh:
If you rebase this on master and push it again, the tests will pass

PR #516, opened 2016-09-29
Record invalid password reset page in analytics

NeilvB:
[Trello card](https://trello.com/c/uviL5BFo/41-add-google-analytics-tracking-to-invalid-passphrase-reset-link-error-message)

Unlike many of the other sign-in and password editing pages, this particular journey was not being recorded in our analytics.


kalleth:
LGTM, and appears to track the same logic as lives in `app/views/shared/_bootstrap_flash_messages.html.erb` so we should be all gravy. :shipit:

PR #518, opened 2016-09-30
Record role changes in the event log

### LINKED ISSUE #299: Role changes should be logged in the event log
I'm logging this as an issue here because I don't know if it's been created in any team's product backlog yet.

The `EventLog` should contain details of role changes, and should include in the log the datetime the change was the made, the user who made the change and the user whose account was affected by the change.

alexmuller:
This adds event log entries like:

> Role changed by Alex Muller from admin to superadmin

Which is helpful when looking at compromised accounts.

Fixes #299.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -58,6 +58,10 @@ class UsersController < ApplicationController
         end
       end
 
+      if role_change = @user.previous_changes[:role]
+        EventLog.record_role_change(@user, role_change.first, role_change.last, current_user)
+      end
+
       redirect_to users_path, notice: "Updated user #{@user.email} successfully"
     else
       render :edit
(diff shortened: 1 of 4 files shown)


test/models/event_log_test.rb
```diff
@@ -66,6 +66,14 @@ class EventLogTest < ActiveSupport::TestCase
     end
   end
 
+  test 'records role changes with the details of the roles' do
```
gpeng: I know the linter isn't complaining but double quotes are used everywhere else in this file. Can we change them here to keep it consistent?

app/controllers/users_controller.rb
```diff
@@ -58,6 +58,10 @@ def update
         end
       end
 
+      if role_change = @user.previous_changes[:role]
```
tijmenb: Is there a good place to test this code as well? There's probably enough logic there to warrant an integration or controller test.

alexmuller:
Thanks for the feedback! Updated.

PR #523, opened 2016-11-21
Configure using env vars

steventux:
Part of https://trello.com/c/BrKRdn4f/495-rebuild-gov-uk-app-deployment-pipeline

Uses ENV vars for Redis config and adds a default unicorn config.

Depends on https://github.com/alphagov/govuk-puppet/pull/5106


config/unicorn.rb
```diff
@@ -0,0 +1,12 @@
+worker_processes 4
+stdout_path "#{ENV.fetch('GOVUK_APP_LOGROOT')}/app.out.log"
+stderr_path "#{ENV.fetch('GOVUK_APP_LOGROOT')}/app.err.log"
+before_exec do |server|
+    ENV["BUNDLE_GEMFILE"] = "#{ENV.fetch('GOVUK_APP_ROOT')}/Gemfile"
```
boffbowsh: Why is this required?

steventux: This is a straight copy of [the file being loaded here](https://github.gds/gds/alphagov-deployment/blob/master/signonotron2/to_upload/unicorn.rb#L4)
Admittedly I didn't employ too much thinking about the contents.

gpeng:
I think the Airbrake initializer also needs to be updated

steventux:
thanks for spotting this @gpeng, updated

PR #526, opened 2016-12-22
Remove Share Sale Publisher

steventux:
This application doesn't exist any more so remove it from Signon


alexmuller:
How much data will this get rid of? I'm thinking mostly of audit logs.

steventux:
ok I will have a look into this, might be better to do a soft delete.

steventux:
@alexmuller this should soft-delete now because https://github.com/alphagov/signonotron2/pull/527

alexmuller:
Looks like magic to me! Will let somebody else who knows more Rails than me have a read too.

PR #527, opened 2016-12-23
Add soft deletion

steventux:
We have a few applications we'd like to retire from signon but we need to keep the records of them.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160420145549) do
+ActiveRecord::Schema.define(version: 20161223162747) do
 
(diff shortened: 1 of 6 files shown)


steventux:
@boffbowsh updated with actual working tests

db/schema.rb
```diff
@@ -11,7 +11,17 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160420145549) do
+ActiveRecord::Schema.define(version: 20161223162747) do
+
+  create_table "_event_logs_old", force: :cascade do |t|
```
boffbowsh: A new table has crept in here

steventux:
@boffbowsh sorry not sure where the table + index crept in from, removed now. The perils of raising PRs so close to xmas :christmas_tree: :christmas_tree: :christmas_tree:

PR #528, opened 2017-01-05
Add Jenkinsfile

steventux:
https://ci.integration.publishing.service.gov.uk/job/signonotron2/job/add-jenkinsfile/1/console

I think this works as it should.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,75 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'signonotron2'
+DB_ADAPTERS = ['mysql', 'postgresql']
+
+node {
+  def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
+
+  properties([
+    buildDiscarder(
+      logRotator(
+        numToKeepStr: '50')
+      ),
+    [$class: 'RebuildSettings', autoRebuild: false, rebuildDisabled: false],
+    [$class: 'ThrottleJobProperty',
+      categories: [],
+      limitOneJobWithMatchingParams: true,
+      maxConcurrentPerNode: 1,
+      maxConcurrentTotal: 0,
+      paramsToUseForLimit: 'Apps using "signonotron2_test" database',
+      throttleEnabled: true,
+      throttleOption: 'category'
+    ],
+  ])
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,78 @@
...
+      maxConcurrentPerNode: 1,
+      maxConcurrentTotal: 0,
+      paramsToUseForLimit: 'Apps using "signonotron2_test" database',
+      throttleEnabled: true,
+      throttleOption: 'category'
+    ],
+  ])
+
+  try {
+    if (!govuk.isAllowedBranchBuild(env.BRANCH_NAME)) {
```
suzannehamilton: I suspect this check is unnecessary because it's only required by applications that are used to test changes to the content schema, which I don't think is the case for Signonotron? Almost all projects are configured in Puppet so that they skip builds for release branches, which makes this check redundant.

The `DEFAULT_SCHEMA_BRANCH` variable above is unnecessary for the same reason.

Jenkinsfile
```diff
@@ -0,0 +1,78 @@
...
+        sh("RAILS_ENV=test SIGNONOTRON2_DB_ADAPTER=${adapter} bundle exec rake --trace")
+      }
+    }
+
+    if (env.BRANCH_NAME == 'master') {
+      stage("Push release tag") {
+        govuk.pushTag(REPOSITORY, env.BRANCH_NAME, 'release_' + env.BUILD_NUMBER)
+      }
+
+      govuk.deployIntegration(REPOSITORY, env.BRANCH_NAME, 'release', 'deploy')
```
cbaines: It would be good to put this in a stage, just for visibility in the Jenkins interface.

Jenkinsfile
steventux: Thanks @suzannehamilton the `DEFAULT_SCHEMA_BRANCH` is redundant, I'm not so sure about the throttling config as old CI has configuration like this:
screenshot from 2017-01-05 15 35 44

suzannehamilton: Sorry, I was just referring to lines 27-29. The throttling config is correct, as far as I know.

steventux: Ah! OK thanks :bulb:

PR #539, opened 2017-02-16
Allow retirement of applications 

tijmenb:
This adds a checkbox to the application edit page that allows users to mark the application as "retired". It's intentionally non-destructive and reversible.

It's an alternative to the soft-deletion https://github.com/alphagov/signon/pull/527 (reverted in https://github.com/alphagov/signon/pull/530 because it didn't work as expected).

## Applications list

screen shot 2017-02-16 at 12 22 36

## Application edit form

screen shot 2017-02-16 at 12 24 45

Part of https://trello.com/c/dGPNmVub/482-retire-panopticon

### DIFF
--- a/db/migrate/20170216105512_add_retired_to_applications.rb
+++ b/db/migrate/20170216105512_add_retired_to_applications.rb
@@ -1,0 +1,5 @@
+class AddRetiredToApplications < ActiveRecord::Migration
+  def change
+    add_column :oauth_applications, :retired, :boolean, default: false
+  end
+end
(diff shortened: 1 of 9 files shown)


db/migrate/20170216105512_add_retired_to_applications.rb
```diff
@@ -0,0 +1,6 @@
+class AddRetiredToApplications < ActiveRecord::Migration
+  def change
+    add_column :oauth_applications, :retired, :boolean, default: false
+    add_index :oauth_applications, :retired
```
gpeng: Do we need an index on a boolean column?

tijmenb: Good point, updated!

Issue #524, opened 2016-12-12
Consider changing the password rotation requirement


matthewford:
The ncsc website is down so I cannot link to the guideance 🙁 , but from their twitter https://twitter.com/ncsc/status/808258353554132992

https://www.ncsc.gov.uk/articles/problems-forcing-regular-password-expiry (for when the website is back live).

fofr:
A good quote from the article:

> It’s one of those counter-intuitive security scenarios; the more often users are forced to change passwords, the greater the overall vulnerability to attack. What appeared to be a perfectly sensible, long-established piece of advice doesn’t, it turns out, stand up to a rigorous, whole-system analysis.

fofr:
We also don't do the following very well, meaning regular password resets are more burdensome:
> Allow users to reset password easily, quickly and cheaply.


We uncovered these problems during 2-step verification research but didn't fix them.

matthewford:
Having reset my passphrase again - I'm reminded of this old issue, please consider it.

boffbowsh:
This is on our backlog of work for Signon, but we don't currently have any work on it planned in our [Roadmap](https://app.productplan.com/p/bUmH4fHC0hOivX-E2LYMu2hg9uEhkWp_)

matthewford:
Hey @boffbowsh, need any help with this one? 

I could see if we can carve out some time from Tariff as we use our own signon instance on dev.

boffbowsh:
Pull Requests are very welcome :)

PR #540, opened 2017-02-23
remove password expired logic

artembiserov:
https://github.com/alphagov/signon/issues/524


cbaines:
Thanks for this @artembiserov :) . Could you add some more information to you commit, particularly about the motivation for making this change (this doesn't have to be complicated)?

app/models/event_log.rb
```diff
@@ -9,7 +9,6 @@ class EventLog < ActiveRecord::Base
     ACCOUNT_UNSUSPENDED                       = LogEntry.new(id: 3, description: "Account unsuspended", require_initiator: true),
     ACCOUNT_AUTOSUSPENDED                     = LogEntry.new(id: 4, description: "Account auto-suspended"),
     MANUAL_ACCOUNT_UNLOCK                     = LogEntry.new(id: 5, description: "Manual account unlock", require_initiator: true),
-    PASSPHRASE_EXPIRED                        = LogEntry.new(id: 6, description: "Passphrase expired"),
```
benlovell: This ought to stay around otherwise these historical audit entries will be orphaned.

matthewford:
@cbaines the primary motivation was a pair programming test, where I thought I might as well scratch an itch at the same time.

matthewford:
It brings the application inline with the new NCC guidance around requiring password rotation.

PR #545, opened 2017-04-07
Add script to prepare training environment

rubenarakelyan:
This commit adds a script that prepares the development VM to act as a training environment by creating applications, organisations, users and permissions.

### DIFF
--- a/script/prepare_training_environment
+++ b/script/prepare_training_environment
@@ -1,0 +1,180 @@
+#!/usr/bin/env ruby
+require_relative "../config/environment"
+
+# This script creates applications and users in the signon database to be
+# used for training purposes. Each application has its OAuth credentials set
+# from either its environment variables (set by puppet) or its gds-sso config
+# file. Users are also created with identical usernames and passwords.
+# All users have full access to all applications. Any existing data in the
+# signon database will be deleted.
+
+unless ARGV[0] == "--really-setup-training-environment"
+  puts "This script must be run with the --really-setup-training-environment argument"
+  exit
+end
+
+NUMBER_OF_USERS = 15.freeze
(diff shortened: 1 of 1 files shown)


script/prepare_training_environment
```diff
@@ -0,0 +1,188 @@
...
+  user.save!
+
+  application = Doorkeeper::Application.where(name: api[:name]).first
+  permissions = SupportedPermission.where(application_id: application.id)
+  permissions.each do |permission|
+    UserApplicationPermission.create!(
+      user_id: user.id,
+      application_id: application.id,
+      supported_permission_id: permission.id
+    )
```
cbaines: There is `grant_application_permissions` to do this more concisely (e.g. https://github.com/alphagov/govuk-guix/blob/c70c9bf4984669cd24aa4c08a921ec6a49f470a6/gds/services/govuk/signon.scm#L262 )

script/prepare_training_environment
```diff
@@ -0,0 +1,188 @@
...
+  end
+
+  token = "oauth-bearer-token-#{api[:slug]}"
+  ActiveRecord::Base.connection.execute(<<-EOQ)
+    INSERT INTO oauth_access_tokens
```
cbaines: I think this might be the equivalent code from the similar script in the govuk-guix repository https://github.com/alphagov/govuk-guix/blob/c70c9bf4984669cd24aa4c08a921ec6a49f470a6/gds/services/govuk/signon.scm#L264-L271 so you might be able to do this in Ruby.

script/prepare_training_environment
rubenarakelyan: Nice, I knew there must be a better way!

script/prepare_training_environment
rubenarakelyan: Likewise, much more concise!

rubenarakelyan:
I was actually thinking about how to prevent this being run accidentally but seems like I forgot to do anything about it. I've added the command-line argument.

PR #549, opened 2017-05-16
Add basic Dockerfile

surminus:
The image is big, but it builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM ruby:2.2.3-slim
+MAINTAINER "govuk-role-platform-accounts-members@digital.cabinet-office.gov.uk"
+
+RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM ruby:2.2.3-slim
+MAINTAINER "govuk-role-platform-accounts-members@digital.cabinet-office.gov.uk"
+
+RUN apt-get update && \
+    apt-get upgrade -y && \
```
thomasleese: One thing we've found for the publishing platform is that splitting out the "update & upgrade" step and the "install packages" step into two separate `RUN`s means the upgraded Docker image can be cached and used as a base by other images.

surminus: Thanks! That's useful. I'll amend.

PR #553, opened 2017-05-31
[DISCUSS] Add signin on support to all users

h-lame:
For: https://trello.com/c/FMBZdTvk/45-support-app-should-probably-require-signon-permission-to-access

Related to: https://github.com/alphagov/support/pull/315

We noticed that the support app, unlike all other signon apps didn't require 'signin' access to allow users to use the app.  This likely leads us into a false sense of security as when we uncheck a users' 'Has access?' checkbox for 'support' in the UI we'd reasonably expect that this would stop them from using the support app.  It doesn't though.  We've made [a change to support to require 'signin' permission](https://github.com/alphagov/support/pull/315) and this PR does the work in signon to make sure that this change doesn't revoke any access that people would have had.

1. A rake task to add a permission for an app to all users (we'll run this with "signin" on "support").
2. Adding a mechanism for a default set of permissions to be added to all users (a true MVP of this, no UI, requires dev effort to change)
3. Make the 3 places where the app creates users (`rake users:create`, batch invite, `users/new`) add these default set of permissions to the users they create

For discussion: is this the right thing to do?  Should we take the default stuff further and add UI on it, or just some docs?

### DIFF
--- a/lib/user_permissions_bulk_granter.rb
+++ b/lib/user_permissions_bulk_granter.rb
@@ -1,0 +1,17 @@
+class UserPermissionsBulkGranter
+  attr_reader :application
+  def initialize(application_name)
+    @application = ::Doorkeeper::Application.find_by(name: application_name)
+    raise "No such application: '#{application_name}'" if @application.nil?
+  end
+
+  def grant(permission_name)
+    permission = application.supported_permissions.find_by(name: permission_name)
+    raise "No such permission: '#{permission_name}' for application: '#{application.name}'" if permission.nil?
+    User.includes(application_permissions: :supported_permission).find_each do |user|
+      unless user.eager_loaded_permission_for(application).include? permission.name
+        user.application_permissions.create!(supported_permission_id: permission.id)
+      end
+    end
+  end
+end
--- a/app/controllers/batch_invitations_controller.rb
+++ b/app/controllers/batch_invitations_controller.rb
@@ -73,9 +73,8 @@ class BatchInvitationsController < ApplicationController
   end
 
   def grant_default_permissions(batch_invitation)
-    support_app = ::Doorkeeper::Application.find_by(name: 'support')
-    if support_app.present?
-      batch_invitation.supported_permissions << support_app.signin_permission
+    Signon.default_permissions_for_all_users.each do |default_permission|
+      batch_invitation.supported_permissions << default_permission
     end
   end
 end
--- a/app/controllers/batch_invitations_controller.rb
+++ b/app/controllers/batch_invitations_controller.rb
@@ -15,6 +15,7 @@ class BatchInvitationsController < ApplicationController
   def create
     @batch_invitation = BatchInvitation.new(user: current_user, organisation_id: params[:batch_invitation][:organisation_id])
     @batch_invitation.supported_permission_ids = params[:user][:supported_permission_ids] if params[:user]
+    grant_default_permissions(@batch_invitation)
     authorize @batch_invitation
 
     unless file_uploaded?
@@ -68,6 +69,13 @@ class BatchInvitationsController < ApplicationController
       false
     else
       true
+    end
+  end
+
+  def grant_default_permissions(batch_invitation)
+    support_app = ::Doorkeeper::Application.find_by(name: 'support')
+    if support_app.present?
+      batch_invitation.supported_permissions << support_app.signin_permission
     end
   end
 end
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -105,9 +105,8 @@ class InvitationsController < Devise::InvitationsController
   end
 
   def grant_default_permissions(user)
-    support_application = ::Doorkeeper::Application.find_by(name: 'support')
-    if support_application.present?
-      user.grant_application_permission(support_application, 'signin')
+    Signon.default_permissions_for_all_users.each do |default_permission|
+      user.grant_permission(default_permission)
     end
   end
 end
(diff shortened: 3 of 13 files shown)


cbaines:
It's unclear to me why the default permissions are needed? Is it something to do with the requirement to remove access from users even when the the criteria in the application is just to have an active account within Signon?

h-lame:
@cbaines I added the default permissions to try and maintain the status quo.  Currently if you create a user without any permissions they will be able to login to the support app by virtue of having a signon account.  After this change (and the change in support), if we don't automatically add signin on support permission to a new user they won't be able to get in.  We could drop the last two commits and leave it up to admins to make sure new users are given support access explicitly.

It's possible that the support app should just require a signon account rather than explicit signin permission on support, but that would make it special and unlike all other apps.  I also think it might be a failure in the API design that the UI suggests you need 'signin' permission to be able to log into an app, and yet this is not enforced by signon, but by the apps themselves.  Arguably it should be up to signon itself to make sure that a user trying to log into a given app has permission to do so (do they have an account, does that account have signin permission for the app they're trying to get into).

cbaines:
> @cbaines I added the default permissions to try and maintain the status quo. Currently if you create a user without any permissions they will be able to login to the support app by virtue of having a signon account. After this change (and the change in support), if we don't automatically add signin on support permission to a new user they won't be able to get in. We could drop the last two commits and leave it up to admins to make sure new users are given support access explicitly.

Ok.

> It's possible that the support app should just require a signon account rather than explicit signin permission on support, but that would make it special and unlike all other apps.

I think there are other apps that just require a Signon account, e.g. by using `authenticate_user!` here [1]. I know the Content Performance Manager does, and I think there are a number of other applications as well.

1: https://github.com/alphagov/gds-sso#securing-your-application

> I also think it might be a failure in the API design that the UI suggests you need 'signin' permission to be able to log into an app, and yet this is not enforced by signon, but by the apps themselves. Arguably it should be up to signon itself to make sure that a user trying to log into a given app has permission to do so (do they have an account, does that account have signin permission for the app they're trying to get into).

I'm uncertain about these changes as they are quite broad, and it's not clear to me what the underlying need is. Going back to:

> This likely leads us into a false sense of security as when we uncheck a users' 'Has access?' checkbox for 'support' in the UI we'd reasonably expect that this would stop them from using the support app. It doesn't though.

I get what you say about the disconnect between Signon and applications, Signon assumes all applications use the `signon` permission, but this is intentionally not the case in a number of apps. If this is just a UI problem in Signon, maybe that can be improved, but I can also see how this might be a problem for applications, as its not possible currently to have an app with access by default, but where its possible to explicitly remove access from individual accounts (without disabling the whole account). How closely do either of these issues match what you are trying to improve with these changes?

h-lame:
> I'm uncertain about these changes as they are quite broad, and it's not clear to me what the underlying need is. Going back to:

The need in the trello story is that we thought it was weird that the support app exists in signon with "has access to" permissions that you can toggle for a user that have no effect on their ability to sign into the support app.  Our acceptance criteria are to make the support app require 'signin' but also maintain the status-quo so that existing users who would have access right now still have access once we make this change and the people that create new users don't have to do anything special to add access to support.

>I think there are other apps that just require a Signon account, e.g. by using authenticate_user! here [1]. I know the Content Performance Manager does, and I think there are a number of other applications as well.

I didn't realise there were lots of apps explicitly built this way.  We [changed 'short-url-manager' a while back](https://github.com/alphagov/short-url-manager/pull/75) to require 'signin' when it previously didn't because we assumed it was a mistake; you could get in without signin but the app required other permissions to let you do anything.  I hadn't seen the ggds-sso docs before, so that might have changed my thinking, but they're in clear conflict with the message the signon UI gives you.

> I get what you say about the disconnect between Signon and applications, Signon assumes all applications use the signon permission, but this is intentionally not the case in a number of apps. If this is just a UI problem in Signon, maybe that can be improved, but I can also see how this might be a problem for applications, as its not possible currently to have an app with access by default, but where its possible to explicitly remove access from individual accounts (without disabling the whole account). How closely do either of these issues match what you are trying to improve with these changes?

Almost exactly.  I think we should be able to control individual user accounts access to individual apps (the signon UI certainly suggests we can) and if we do this I think it would be useful to provide a default set of permissions for new users (that can be toggled later).  I hadn't considered the alternative of fixing the signon UI, but that could work too.  So that the UI can be made clearer I think we'd need to add something to each app to tell signon that this app requires only an account, not explicit "signin on that app" permissions.  In either case I think it would be useful for signon itself to handle the "do I need signin or do I just need an account" aspect as part of the oauth dance and so apps themselves just need to `authenticate_user!`.

I think this sort of fix is probably out of scope for this PR though and would benefit from a wider discussion (an RFC perhaps?).

lib/user_permissions_bulk_granter.rb
```diff
@@ -0,0 +1,17 @@
...
+    raise "No such application: '#{application_name}'" if @application.nil?
+  end
+
+  def grant(permission_name)
+    permission = application.supported_permissions.find_by(name: permission_name)
+    raise "No such permission: '#{permission_name}' for application: '#{application.name}'" if permission.nil?
+    User.includes(application_permissions: :supported_permission).find_each do |user|
+      unless user.eager_loaded_permission_for(application).include? permission.name
+        user.application_permissions.create!(supported_permission_id: permission.id)
+      end
```
h-lame: I think we'd get the same error message (it'd raise ActiveRecord::RecordInvalid on the `find_or_create!`).  I chose this re-implementation to take advantage of eagerloading and avoid n+1 while iterating through all the users.  I admit it's a little unclear though, so maybe it'd be better to optimise for readability.  What do you think?

h-lame:
Closing this for now, in preference to implementing the recommendations in [RFC 78](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-078-re-architect-signin-permissions-in-signon.md), which we're tracking in [this trello card](https://trello.com/c/c4bgclDp/171-implement-rfc-78-make-signon-handle-signin-permission-during-oauth)

PR #554, opened 2017-06-02
Add hint about locked accounts to signin failure warning

h-lame:
For security reasons we don't tell the person entering credentials that an
account has been locked when accessing signon. That would reveal that the
details entered are for an actual account.

We can however suggest to the user that after X attempts the account _may_
now be locked, so that's what we do here.

There are 3 possible error messages you'd get for failing a login: locked,
invalid, and not_found_in_database.  We've changed all of these to the
same generic "failure, maybe locked?" message.  We are running devise in
paranoid mode so it will only use the invalid message, but we change all
of them just in case.

For: https://trello.com/c/2InrPGTy/163-know-when-you%27ve-been-locked-out


config/locales/devise.en.yml
```diff
@@ -17,9 +17,9 @@ en:
       already_authenticated: 'You are already signed in.'
       unauthenticated: 'You need to sign in before continuing.'
       unconfirmed: 'You have to confirm your account before continuing.'
-      locked: 'Your account is temporarily locked.'
-      invalid: 'Invalid email or passphrase.'
-      not_found_in_database: 'Invalid email or passphrase.'
+      locked: 'Invalid email or passphrase. Entering incorrect passphrase 6 times will lock your account and you will be sent an email to reset it.'
+      invalid: 'Invalid email or passphrase. Entering incorrect passphrase 6 times will lock your account and you will be sent an email to reset it.'
+      not_found_in_database: 'Invalid email or passphrase. Entering incorrect passphrase 6 times will lock your account and you will be sent an email to reset it.'
```
fofr: Copy should read "Entering an incorrect…"

PR #555, opened 2017-06-12
Change what organisation admins can do

h-lame:
For: https://trello.com/c/B4Y7e22o/164-self-service-or-devolved-unlock-spike

Our goal is to increase the number of org admins there are so we can increase the number of people that can do simple user support requests.  Before this change: Org admins are basically admins who are limited in scope to just users in their org (or sometimes their org and it's children, grandchildre, etc..).  After this change: org admins can no longer create new users (directly or via batch invite), see a list of organisations, affect users in their subtree - they can only affect users that belong directly to their org.

The existing org admins (there are 7) have been spoken to and none have an issue with the restriction of their permissions.

### DIFF
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -12,9 +12,9 @@ class UserPolicy < BasePolicy
     when 'admin'
       !record.superadmin?
     when 'organisation_admin'
-      current_user.id == record.id ||
+      allow_self_only ||
         (record.normal? && belong_to_same_organisation_subtree?(current_user, record))
-    when 'normal'
+    else # 'normal'
       false
     end
   end
@@ -23,29 +23,20 @@ class UserPolicy < BasePolicy
   alias_method :unlock?, :edit?
   alias_method :suspension?, :edit?
   alias_method :resend?, :edit?
+  alias_method :event_logs?, :edit?
 
   def edit_email_or_passphrase?
-    current_user.id == record.id
+    allow_self_only
   end
-
-  def update_email?
-    current_user.id == record.id
-  end
-
-  def update_passphrase?
-    current_user.id == record.id
-  end
+  alias_method :update_email?, :edit_email_or_passphrase?
+  alias_method :update_passphrase?, :edit_email_or_passphrase?
 
   def cancel_email_change?
-    (current_user.id == record.id) || edit?
+    allow_self_only || edit?
   end
 
   def resend_email_change?
-    (current_user.id == record.id) || edit?
-  end
-
-  def event_logs?
-    current_user.normal? ? false : edit?
+    allow_self_only || edit?
   end
 
   def assign_role?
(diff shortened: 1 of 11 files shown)


test/controllers/users_controller_test.rb
```diff
@@ -438,35 +438,53 @@ def change_user_password(user_factory, new_password)
         assert_match(/You do not have permission to perform this action./, flash[:alert])
       end
 
+      should "can give permissions to all applications even where signin is not delegatable or they don't have access to it" do
```
elenatanasoiu: Is it worth making the distinction between the admin and the person being granted rights by saying:
`can give permissions to a user for all applications in signon, even when signin is not delegatable or the admin does not have access to those applications`
?

spec/policies/organisation_policy_spec.rb
```diff
@@ -19,15 +19,14 @@
       expect(subject).to permit(create(:user_in_organisation, role: 'admin'), build(:organisation))
     end
 
-    it "allows organisation admins to assign a user only to organisations within their organisation subtree" do
+    it "is forbidden for organisation admins even when assigning a user to organisations within their organisation subtree" do
```
elenatanasoiu: Still nitpicking on descriptions, but it's only because I'm hoping it's clear what this does to whomever looks at it for the first time without having much context. Perhaps saying:
`is forbidden for organisation admins to assign a user to organisations within their organisation subtree`
would be a bit clearer?

h-lame: Thanks for picking up on it.  On reflection maybe the suffix isn't needed at all.  Just "is forbidden for organisation admins" feels like it's enough.  The details of "with their org" or "an org in their subtree" can live in the spec itself (and are really only relevant when we think about the permissions org admins *used* to have.

app/policies/user_policy.rb
```diff
@@ -13,12 +20,11 @@ def edit?
       !record.superadmin?
     when 'organisation_admin'
       allow_self_only ||
-        (record.normal? && belong_to_same_organisation_subtree?(current_user, record))
+        (record.normal? && (record.organisation && record.organisation.id == current_user.organisation.id))
```
h-lame: I don't know that `allow_organisation_admin` feels correct for the name for the whole condition, but I do like `own_organisation`.  I'll use that.

In a (unpushed) version of this change I tried to name all the conditionals, and had `allow_superadmin` or `allow_admin` etc... but in the end I found the names didn't really express the concern.  For example my `allow_admin` checked that the current user was an admin, but also that the record being acted on wasn't a superadmin and that didn't feel like the right name for that condition.  I ended up backing all that out and making just the `allow_self_only` change.  I think `own_organisation` works though.

test/controllers/users_controller_test.rb
h-lame: Like `is forbidden for organisation admins even when assigning a user to organisations within their organisation subtree` below I decided that in this case we can also drop the suffix.

h-lame:
Addressed comments in a rebase, thanks @elenatanasoiu!

app/policies/user_policy.rb
elenatanasoiu: 👍

PR #557, opened 2017-06-15
Send custom dimension including org slug to ga

h-lame:
For: https://trello.com/c/svo2ZN6s/131-custom-dimension-send-user-email-domain-to-ga-for-signon-apps

We want to know which departments are using our publishing apps and we've agreed with the performance team that the best way to do this is to send the slug of the users organisation to GA as a custom dimension.  If they don't have an organisation we'll send `(not set)` instead.  Lots of users don't have orgs in signon, but they are mostly local government users and unlikely to be users of the apps the perf team are interested in tracking at the moment.  As a test we'll be doing this to signon and whitehall to see what kinds of results we get.

This relies on https://github.com/alphagov/govuk_admin_template/pull/152 and will need the first and final commits squashing once that PR is available in a released version of govuk_admin_template.


h-lame:
Redone to use org slug instead of email domain, after comments on https://github.com/alphagov/whitehall/pull/3293

cbaines:
Looks good :+1: . The CI issue seems to be Docker related, something about not having Git available, so I think that might go away once you switch to a released version of the `govuk_admin_template` gem.

test/helpers/analytics_helpers.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+  end
+
+  def refute_dimension_is_set(dimension)
+    refute_match(/#{Regexp.escape("GOVUKAdmin.setDimension(#{dimension}")}/, page.body)
+  end
+
+  def assert_dimension_is_set(dimension, with_value: nil)
+    dimension_set_js_code = "GOVUKAdmin.setDimension(#{dimension}"
+    dimension_set_js_code += ", \"#{with_value}\")" if with_value.present?
```
elenatanasoiu: Same as in https://github.com/alphagov/whitehall/pull/3293, a closing bracket is missing. 

I'm going to repeat what I said there: 

```
So if there's no with_value present, we're missing a closing bracket for the previous line ?

Would a better solution be:
dimension_set_js_code += with_value.present? ? ", \"#{with_value}\")" : ")"
?

It hasn't failed because we do regex matching
```

PR #558, opened 2017-07-10
Fix URLs for training environment

rubenarakelyan:
This commit changes the URLs used for the `prepare_training_environment` script from `dev.gov.uk` to `training.publishing.service.gov.uk`.


cbaines:
Having to change the Signon app suggests that this functionality might better sit elsewhere.

rubenarakelyan:
Could be a candidate to be moved into the `training-vm` directory in `govuk-puppet`.

PR #559, opened 2017-07-12
Stop org admin editing a user clearing non delegatable permissions

h-lame:
For: https://trello.com/c/uT9YI5mR/188-devolve-unlock-resetting-permissions

After rolling out org-admin permissions to more people (see #555) we had some reports that saving a user as an org admin was resetting their permissions.  Turns out this is a long-standing bug that we didn't know about because we didn't have many org admins.  When updating a user the `UserUpdater` sets their `supported_permission_ids` to whatever is passed in.  When presenting a user to a superadmin or admin user we show the complete set of permissions and submit the selected ones.  Meaning that the `UserUpdater` is given the full set of permissions that the user should be granted.  When presenting a user to an org admin however, we show a limited set of permissions based on the apps that the org admin has access to that allow delegation of permissions.  This means the `UserUpdater` is only given a subset of permissions but doesn't know there's a difference and will assume it is the complete set.

In this PR we change the `UserUpdater` to filter the permission ids so that it will only add or remove permissions from the user being managed that the user doing the managing is allowed to manage.  Meaning that if a super admin gives you access to whitehall (an app that does not allow delegation of permissions) then an org admin editing you will not clear that permission as they aren't allowed to change it.

There are some more details in the individual commits so it's worth looking at those.

An unanswered question is around permissions that are not grantable from the ui - these are filtered out in the view so they can never be presented to any admin user, but are not currently filtered out by the code we've added to make sure we don't add/remove permissions we're not allowed to change.  This means any user with a permission that is not grantable from the ui will have that permission stripped if they are edited and saved via the UI by an admin.  I don't know that it's a huge problem as the only permissions that are not grantable from the UI at the moment are the `user_update_permission` ones which are only granted to the `Signonotron API Client (permission and suspension updater)` user.   That said, it's a weird edge-case that we should probably fix, I'm juust not really sure what the intention really is for these permissions that you can't grant from the UI.

### DIFF
--- a/app/policies/supported_permission_policy.rb
+++ b/app/policies/supported_permission_policy.rb
@@ -1,0 +1,16 @@
+class SupportedPermissionPolicy < BasePolicy
+  class Scope < ::BasePolicy::Scope
+    def resolve
+      if current_user.superadmin?
+        scope.all
+      elsif current_user.admin?
+        scope.all
+      elsif current_user.organisation_admin?
+        app_ids = Pundit.policy_scope(current_user, :user_permission_manageable_application).pluck(:id)
+        scope.joins(:application).where(oauth_applications: { id: app_ids })
+      else
+        scope.none
+      end
+    end
+  end
+end
--- a/app/services/user_update.rb
+++ b/app/services/user_update.rb
@@ -10,7 +10,7 @@ class UserUpdate
   def update
     user.skip_reconfirmation!
     old_permissions = fetch_user_permissions
-    return unless user.update_attributes(user_params)
+    return unless update_user
     user.application_permissions.reload
 
     record_update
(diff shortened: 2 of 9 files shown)


app/services/user_update.rb
```diff
@@ -24,6 +24,49 @@ def update
 
 private
 
+  class SupportedPermissionsSanitizer
```
elenatanasoiu: Wouldn't this be better housed in the `lib` folder? Similar to:

https://github.com/alphagov/signon/blob/master/lib/user_parameter_sanitiser.rb
and
https://github.com/alphagov/signon/blob/master/lib/unsupported_permission_error.rb

app/policies/supported_permission_policy.rb
```diff
@@ -0,0 +1,16 @@
+class SupportedPermissionPolicy < BasePolicy
+  class Scope < ::BasePolicy::Scope
+    def resolve
+      if current_user.superadmin?
+        scope.all
+      elsif current_user.admin?
+        scope.all
+      elsif current_user.organisation_admin?
+        app_ids = Pundit.policy_scope(current_user, :manage_supported_permissions_for_application).pluck(:id)
```
elenatanasoiu: Don't `SupportPermissionsPolicy` and `ManageSupportedPermissionsForApplicationPolicy` do essentially the same thing? 
What's the added benefit of keeping the logic for `ManageSupportedPermissionsForApplicationPolicy` separated from `SupportedPermissionPolicy`?

h-lame: One returns a list of `Doorkeeper::Application`s (`ManageSupportedPermissionsForApplicationPolicy`), the other returns a list of `SupportedPermission` (`SupportedPermissionPolicy`).  The name for the former is far from ideal, so I'm open to suggestions.  I don't think it can just be `ApplicationPolicy` because it's not the list of applications the user can see, but the list of applications the user can see in a given context (arguably maybe `SupportedPermissionPolicy` should be given a more explicit name too.

app/services/user_update.rb
h-lame: Oh, interesting, I didn't spot the other sanitiser.  I'm not sure sanitizer is even the right name for this thing (it has connotations of rails own strong parameters stuff which is key focussed, when in fact we're focussed on values in this thing).  I'll move it though.

app/policies/supported_permission_policy.rb
elenatanasoiu: You can't really tell from the name that it returns a list of apps, not of permissions for the apps. 
How about `ApplicationAccessPolicy` ?

app/services/user_update.rb
elenatanasoiu: What about `SupportedPermissionsFilter` or `SupportPermissionsGuard` ?

h-lame: Oooh! `<blah>Filter` is good. :+1:

app/policies/supported_permission_policy.rb
h-lame: I think `Access` is wrong that word is used elsewhere to mean the applications that the user has been granted `signin` permission to, and that's not what this list is for, it's the list of applications the user is allowed to grant permissions to other users for.  One candidate therefor might be `ApplicationsUserCanGrantPermissionsToOtherUsersForPolicy`; it's cumbersome for sure but it is more obvious than what I had before.  On the other hand this logic did come from a method called `visible_applications` so we could use that `VisibleApplicationsPolicy`, but it's not intention revealing at all.  I'll have another think.

h-lame: Maybe `PermissionGrantableApplicationsPolicy` ?

elenatanasoiu: I guess that's the closest to brevity + clarity that we can get :/

app/services/user_update.rb
h-lame: Done in 9de35b38f04cfa15b60541e38a46bf6e32cf1e91

app/policies/supported_permission_policy.rb
h-lame: Gone with `UserPermissionManageableApplicationPolicy`.  `UserPermission` because I wanted to be clear that it wasn't the ability to add/remove permissions from the application (e.g. you shouldn't can't use this scope to get the list of applications that you are allowed to edit).  `Manageable` rather than `Grantable` because a) we can revoke permissions as well as grant them, b) there's a method `manage_supported_permissions?` on the `UserPolicy` that is used to say if a user is allowed to do this kind of thing at all, and I hoped that the `Managable` would act as a link to the `manage` part of that.

h-lame: Done in a rebase.

PR #560, opened 2017-08-15
Handle signin permission during oauth

h-lame:
For: https://trello.com/c/c4bgclDp/195-implement-rfc-78-make-signon-handle-signin-permission-during-oauth

Implements: https://github.com/alphagov/govuk-rfcs/blob/8cbb2a0de86de02f54ae37b245e79b46ad62cb6a/rfc-078-re-architect-signin-permissions-in-signon.md

This is a work-in-progress that I'm pushing up to get some extra eyes on it earlier rather than later.  In particular, I'm interested in any thoughts people have on the 2nd commit [`Add signin permission requirement to doorkeeper_authorize!`](8820988eefbd9258abe7f6d4f75630159c3ce90a) as I'm not 100% sure it's a required change.  I would definitely be grateful for any advice from people who know a bunch of stuff about oauth.

### Todo

- [x] Signon MUST check for 'signin' permission during the oauth handshake with an application and reject users that do not have this permission. (commits 1 and 2 in this PR)
- [x] Signon MUST allow a super admin to grant permissions to all users in one go. (commits 3-9 in this PR)
- [x] Signon SHOULD allow super admins to mark permissions for an application as 'default' meaning it is added to all new users. (commits 10-13 - mostly repurposed from #553)
- [x] gds-sso SHOULD deprecate require_signin_permission!. (https://github.com/alphagov/gds-sso/pull/119)
- [x] applications using signon SHOULD be audited to make sure they will continue to function after these changes are made - see https://docs.google.com/spreadsheets/d/1ypHdDHaHJZELkdQC3xQprMxJpkPCLTfnJPsEtPhhENI/edit#gid=0
- [x] applications using signon SHOULD be audited to collect the set of default permissions that should be created and bulk granted to all existing users.
- [x] ~~omniauth-gds SHOULD deprecate whatever it does with signin permissions~~ omniauth-gds doesn't do anything wrt signin permissions, each app that uses it does this manually

Some of these will be completed by doing work in other repos, I'll link to the relevant PRs when they're done.

This will be easiest reviewed commit-by-commit as it's a big change, however, github has messed up the commit order.  To review in the correct order do them in this sequence (it's the git graph order, not the date order that github shows):

1. 1656e10c6243c4cb8d01c19c3ac58c3bf3d7f4d5 Require signin permission to complete oauth authorisation
2. 9c1c76057636a21beb250fee844debeff2986ba1 Add signin permission requirement to doorkeeper_authorize!
3. 3ab1b9f644e25ea9a290c5262ce7f2d27c705cf1 Add name to supported_permissions factory
4. 718077d77c971a944e06d5b614924e441d1372d7 Introduce BulkGrantPermissionSet model and job
5. 55f7246417d8922cb3ce05a9edfe9a4c7bda8913 Rename status_message helper to batch_invite_status_message
6. cdf24b88d582cf4acc8ad2b1af90fa29fb7d0582 Add processed and total user counts to bulk grant permission set
7. f63e48d1e23336b6728eb68e6489caab0bc6f498 Restrict bulk grant permission set to admins and superadmins
8. 03e56b135e7f4b2211a1f7c3045e509d600a3f60 Add UI to allow bulk granting permissions
9. 03eda8786c0923b0facd84027883f2a21f344e37 Represent the bulk grant form if there are errors
10. 9c40ce14e0d4efb583b4bd5fd63b51130fa4805c Add default flag to supported permissions
11. 6dc5b58e7a3072cced651dc6f7bdd2498cab8a40 Remove unsupported permissions check from rake users:create
12. 4f5f0aa14cc60bacb7eecf4f56fbf669c0b0d5ef Extract internals of rake users:create to an object in lib
13. 04ef7198baa54067cd0da4f8f9e8542decc331c6 New users are granted default permissions
14. 30ce041ada2aeab6c012619c182469c865564a21 Remove special case for support app on dashboard
15. f4d92b7381786e484796a0aadbd11a60030c35d6 Record an EventLog when bulk granting permissions

## Screenshots

1.  UI for granting permissions to all users

creating-bulk-grant-permission-set

2. UI for watching that bulk grant being processed

processing-bulk-grant-permission-set

These take "inspiration" (e.g. are copy+paste jobs pretty much) from the existing UI for batch inviting a CSV file of users and granting them permissions.

### DIFF
--- a/app/controllers/signin_required_authorizations_controller.rb
+++ b/app/controllers/signin_required_authorizations_controller.rb
@@ -1,0 +1,40 @@
+class SigninRequiredAuthorizationsController < Doorkeeper::AuthorizationsController
+  def new
+    if pre_auth.authorizable?
+      if skip_authorization? || matching_token?
+        if user_has_signin_permission_to_application?
+          auth = authorization.authorize
+          redirect_to auth.redirect_uri
+        else
+          session[:signin_missing_for_application] = application.try(:id)
+          redirect_to signin_required_path
+        end
+      else
+        render :new
+      end
+    else
+      render :error
+    end
+  end
+
+  def create
+    if user_has_signin_permission_to_application?
+      redirect_or_render authorization.authorize
+    else
+      session[:signin_missing_for_application] = application.try(:id)
+      redirect_to signin_required_path
+    end
+  end
+
+private
+
+  def user_has_signin_permission_to_application?
+    return false if application.nil?
+    return false if current_resource_owner.nil?
+    current_resource_owner.has_access_to?(application)
+  end
+
+  def application
+    pre_auth.try(:client).try(:application)
+  end
+end
--- a/app/controllers/signin_required_authorizations_controller.rb
+++ b/app/controllers/signin_required_authorizations_controller.rb
@@ -1,4 +1,6 @@
 class SigninRequiredAuthorizationsController < Doorkeeper::AuthorizationsController
+  EXPECTED_DOORKEEPER_VERSION = '4.2.0'
+
   def new
     if pre_auth.authorizable?
       if skip_authorization? || matching_token?
--- a/app/controllers/signin_required_authorizations_controller.rb
+++ b/app/controllers/signin_required_authorizations_controller.rb
@@ -1,4 +1,5 @@
 class SigninRequiredAuthorizationsController < Doorkeeper::AuthorizationsController
+  include Pundit
   EXPECTED_DOORKEEPER_VERSION = '4.2.0'
 
   def new
--- a/app/models/bulk_grant_permission_set.rb
+++ b/app/models/bulk_grant_permission_set.rb
@@ -21,9 +21,21 @@ class BulkGrantPermissionSet < ActiveRecord::Base
 
   def perform(_options = {})
     self.update_column(:total_users, User.count)
-    User.find_each do |user|
-      supported_permissions.each do |permission|
-        user.application_permissions.where(supported_permission_id: permission.id).first_or_create!
+    User.find_each do |user_to_change|
+      permissions_granted = supported_permissions.select do |permission|
+        granted_permission = user_to_change.application_permissions.where(supported_permission_id: permission.id).first_or_create!
+        # if 'id' changed then it was a new permission, otherwise it
+        # already existed
+        granted_permission.previous_changes.has_key? 'id'
+      end
+      permissions_granted.group_by(&:application_id).each do |application_id, permissions|
+        EventLog.record_event(
+          user_to_change,
+          EventLog::PERMISSIONS_ADDED,
+          initiator: user,
+          application_id: application_id,
+          trailing_message: "(#{permissions.map(&:name).join(', ')})",
+        )
       end
       self.class.increment_counter(:processed_users, self.id)
     end
--- a/app/models/bulk_grant_permission_set.rb
+++ b/app/models/bulk_grant_permission_set.rb
@@ -36,6 +36,6 @@ class BulkGrantPermissionSet < ActiveRecord::Base
 private
 
   def must_have_at_least_one_supported_permission
-    errors.add(:supported_permissions, 'must add at least one permission to grant to all users') if bulk_grant_permission_set_application_permissions.size.zero?
+    errors.add(:supported_permissions, 'must not be blank. Choose at least one permission to grant to all users.') if bulk_grant_permission_set_application_permissions.size.zero?
   end
 end
--- a/app/models/bulk_grant_permission_set.rb
+++ b/app/models/bulk_grant_permission_set.rb
@@ -7,18 +7,27 @@ class BulkGrantPermissionSet < ActiveRecord::Base
   validates :outcome, inclusion: { in: %w(success fail), allow_nil: true }
   validate :must_have_at_least_one_supported_permission
 
+  def in_progress?
+    outcome.nil?
+  end
+
+  def successful?
+    outcome == 'success'
+  end
+
   def enqueue
     BulkGrantPermissionSetJob.perform_later(self.id)
   end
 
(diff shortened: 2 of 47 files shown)


h-lame:
Removed WIP tag as this is now feature complete and ready for review.  Gonna work on the outstanding audit tasks separately.  There *may* be additional commits to add seed data based on those audits, but it's unlikely they'd change the rest of the code in the PR.

boffbowsh:
This all looks reasonable to me, but I've only had time to skim it. I don't see any obvious security holes. It'd still be good to get a review from someone else. Perhaps a willing volunteer could pair with you on a review?

app/models/bulk_grant_permission_set.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def enqueue
+    BulkGrantPermissionSetJob.perform_later(self.id)
+  end
+
+  def perform(_options = {})
+    self.update_column(:total_users, User.count)
+    User.find_each do |user|
+      supported_permissions.each do |permission|
```
tijmenb: Could we audit-log the permission changes here like we do for manual updates?

See https://github.com/alphagov/signon/pull/541:

https://github.com/alphagov/signon/blob/6208df339d20f0aa5c5bdb51f1b4b171d042f28c/app/services/user_update.rb#L34-L40

h-lame: oh, yes, absolutely.  I'll do a quick check to make sure that we're always adding event logs for all the places where we create users and grant permissions as I suspect this won't be the only place where we've slipped through the net.

fofr:
Are there some screenshots of any new UI?

h-lame:
@fofr - good point.  Added some to the PR description.

Davidslv:
This should probably be tested on integration and staging before merging

app/controllers/signin_required_authorizations_controller.rb
```diff
@@ -0,0 +1,40 @@
+class SigninRequiredAuthorizationsController < Doorkeeper::AuthorizationsController
+  def new
+    if pre_auth.authorizable?
+      if skip_authorization? || matching_token?
+        if user_has_signin_permission_to_application?
+          auth = authorization.authorize
```
h-lame: Lost comment due to rebase:

> @Davidslv: 
>> you might want to call super here instead of repeating code.

> @h-lame: 
>> If I do that then it'll run these lines of code again:
>>
>> ```ruby
>> if pre_auth.authorizable?
>>   if skip_authorization? || matching_token?
>> ```
>>
>> And I don't know if those are "cheap" or have no side-effects, so while it's annoying to duplicate the code, I'd rather keep this as-is.

app/policies/bulk_grant_permission_set_policy.rb
```diff
@@ -0,0 +1,9 @@
+class BulkGrantPermissionSetPolicy < BasePolicy
+  def new?
+    return true if current_user.superadmin? || current_user.admin?
+
+    false
+  end
+  alias_method :create?, :new?
```
h-lame: Lost comment due to rebase:

> @Davidslv:
>> What's the reason for these aliases?

> @h-lame:
>> The policy stuff in pundit works at the controller level by asking if the user is allowed to perform the given controller action on the instance (or class). We're following the pattern established by some of the other policies by defining our rules for one action, and then aliasing the other actions onto it so they're all the same.

app/models/bulk_grant_permission_set.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def successful?
+    outcome == 'success'
+  end
+
+  def enqueue
+    BulkGrantPermissionSetJob.perform_later(self.id)
+  end
+
+  def perform(_options = {})
```
h-lame: Lost comment due to rebase:

> @Davidslv:
>> what's the reason to pass options here if none is being used?

> @h-lame:
>> Mostly so it looks the same as `BatchInvitation` which has a similar "create an instance, queue up a job, run the job by finding the instance and calling perform on it". `BatchInivitation#perform` takes `options` and so I though this should too.  I think the reason for it is that the job takes `options` by default and we pass them through (even if they're ignored). I guess it makes it easy to pass `options` in later if we need to? Not strongly wedded to this, but I'd rather this and `BatchInvitation` look similar, so I could remove the options from both.

db/migrate/20170818134716_add_default_flag_to_supported_permissions.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultFlagToSupportedPermissions < ActiveRecord::Migration
+  def change
+    add_column :supported_permissions, :default, :boolean, default: false, null: false
```
h-lame: Lost comment due to rebase:

> @davidslv:
>> How big is this table? Adding a default value will lock the table until all records are written

> @h-lame:
>> At the time this migration is run it'll be empty because we only just created it in the previous migration.

h-lame: Originally this was a question on the changes to the `bulk_grant_permission_sets` table which is split over two commits (but ultimately only one migration).  However, it's a question that's relevant here on this commit and migration.  Unlike I suggest in my previous answer, the `supported_permissions` isn't empty and hasn't just been created, but it does have only 725 rows, so I think we're safe enough to add the column inline.

sihugh:
I approve this message

Davidslv:
I finally found them on the notifications page. All good from me too

h-lame:
We've tested this branch on integration and staging and encountered no blocking problems so we're gonna merge this.  We have written up [a card for some enhancements about communication of permission changes between signon and apps](https://trello.com/c/URuaiaCl/247-make-sure-signon-always-tells-applications-about-permissions-changes) - but we don't consider it a blocker to progress.

PR #561, opened 2017-08-23
Allow org admins to edit other org admins

elenatanasoiu:
For: https://trello.com/c/Cvttl8Sx/209-can-org-admins-unsuspend-other-org-admins

Following a request from an org admin to be able to unsuspend other org admins (from their own org), we are now increasing the level at which org admins are allowed to operate.

Previously, they were only allowed control on `normal` users, without being able to manage other org admins. This is not true for superadmins, where they can edit users with the same role.

### DIFF
--- a/lib/roles/organisation_admin.rb
+++ b/lib/roles/organisation_admin.rb
@@ -21,7 +21,7 @@ module Roles
     def self.level; 2; end
 
     def self.manageable_roles
-      ['normal']
+      %w{normal organisation_admin}
     end
   end
 end
--- a/lib/roles/organisation_admin.rb
+++ b/lib/roles/organisation_admin.rb
@@ -1,5 +1,5 @@
 module Roles
-  class OrganisationAdmin
+  class OrganisationAdmin < Base
     def self.permitted_user_params
       [
         :uid,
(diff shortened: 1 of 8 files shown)


spec/policies/user_policy_spec.rb
```diff
@@ -2,164 +2,200 @@
...
-      expect(subject).not_to permit(create(:organisation_admin), User)
-      expect(subject).not_to permit(create(:user), User)
+    # Users shouldn't be able to do admin-only things to themselves
+    admin_actions.each do |permission_name|
+      permissions permission_name do
+        it "is not allowed for normal users accessing their own record" do
+          normal_user = create(:user)
+          expect(subject).not_to permit(normal_user, normal_user)
+        end
+      end
```
h-lame: These new specs are really nice.  I like how explicit it all is.  Good stuff.

lib/roles/organisation_admin.rb
```diff
@@ -21,7 +21,11 @@ def self.role_name
     def self.level; 2; end
 
     def self.manageable_roles
-      ['normal']
+      %w{normal organisation_admin}
+    end
+
+    def self.can_manage?(role)
+      manageable_roles.include?(role)
```
h-lame: I don't think we need this in here as it's in `Roles::Base` though.

h-lame: We don't, so I removed it in a rebase in order to merge and deploy this.

PR #562, opened 2017-08-29
Add access permissions list on edit user page for org admins

elenatanasoiu:
For: https://trello.com/c/oEW7PsNs/192-devolve-unlock-ability-to-see-permissions

This will allow organisation admins to see all the permissions for a user, instead of seeing only the applications and permissions that the organisation admin has access to.

### DIFF
--- a/app/views/users/_app_permissions.html.erb
+++ b/app/views/users/_app_permissions.html.erb
@@ -1,0 +1,34 @@
+<table id="all-permissions" class="table table-bordered table-striped table-on-white">
+  <thead>
+    <tr class="table-header">
+      <th>Application</th>
+      <% unless user_object.api_user? %>
+          <th>Has access?</th>
+      <% end %>
+      <th><%= "Other" unless user_object.api_user? %> Permissions</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @application_permissions.each do |application, permissions|
+      permission_names = permissions.map(&:name)
+    %>
+      <tr>
+        <td>
+          <% if application.retired? %>
+              <del><%= application.name %></del>
+          <% else %>
+              <%= application.name %>
+          <% end %>
+        </td>
+        <% unless user_object.api_user? %>
+          <td>
+            <%= permission_names.include?('signin') ? 'Yes' : 'No' %>
+          </td>
+        <% end %>
+        <td>
+          <%= permission_names.reject{|p| p == 'signin'}.to_sentence %>
+        </td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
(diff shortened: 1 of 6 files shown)


lib/user_permissions_controller_methods.rb
```diff
@@ -24,4 +24,16 @@ def zip_permissions(applications, user)
...
   end
+
+  def all_applications_and_permissions_for(user)
+    permission_names = Hash[SupportedPermission.pluck(:id, :name)]
+    app_permissions = {}
+
+    UserApplicationPermission.where(user: user).includes(:application).find_each do |permission|
+      app_permissions[permission.application] ||= []
+      app_permissions[permission.application] << permission_names[permission.supported_permission_id]
+    end
```
h-lame: If we did `user.supported_permissions.includes(:application).group_by(&:application)` here then I think we could avoid the `permission_names` hash, and let `group_by` do the aggregation for us.  At least I think so, anyway.  The reason I suggest this is that it took me a moment to work out why we had `permission_names` and what the ultimate state of `app_permissions` would be (a hash of application instances as keys, with arrays of permission names as values), and I think leveraging AR methods might help with comprehension.

app/views/users/_app_permissions.html.erb
```diff
@@ -0,0 +1,36 @@
...
+        <tr>
+          <td>
+            <% if application.retired? %>
+                <del><%= application.name %></del>
+            <% else %>
+                <%= application.name %>
+            <% end %>
+          </td>
+          <% if user_object.api_user? %>
+            <td></td>
```
h-lame: We don't render the `Has access?` column header for API users so we dont' need to render an empty cell for it here.

app/views/users/_app_permissions.html.erb
```diff
@@ -0,0 +1,36 @@
...
+          <% else %>
+            <td>
+              <%= user_object.has_access_to?(application) ? "Yes" : "No" %>
```
h-lame: I like using the API here as it's easier to understand and I like working with the domain model, but, that said - we've got an array of all the permissions already fetched so we could do `permissions.include? 'signin' ? 'Yes' : 'No'` here to avoid the SQL call that `has_access_to?` will run.

app/views/users/_app_permissions.html.erb
```diff
@@ -0,0 +1,36 @@
...
+            </td>
+          <% end %>
+          <td>
+            <%= permissions.join(', ') %>
```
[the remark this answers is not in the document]
h-lame: We probably want to remove `signin` from here as we're handling it in the previous column.  If we leave it in then we have a disconnect between the editable UI where `signin` won't appear under "Other permissions" and this list where it will.

elenatanasoiu:
Added small changes

PR #563, opened 2017-09-01
Add extra info in account access logs

DennisTsiang:
For: https://trello.com/c/Lf1TWggV

Ip address and browser of user is now recorded on each successful
sign in. 'browser' gem is used to retrieve the browser that the
user is using to sign in.

When a new user account is created by an existing admin, the new
account has a record of who it was invited by.

These extra logging should make it easier to retroactively
trace user behaviour in the case of an incident.

### DIFF
--- a/app/models/event_log.rb
+++ b/app/models/event_log.rb
@@ -40,6 +40,7 @@ class EventLog < ActiveRecord::Base
     ACCOUNT_UPDATED                           = LogEntry.new(id: 34, description: "Account updated", require_initiator: true),
     PERMISSIONS_ADDED                         = LogEntry.new(id: 35, description: "Permissions added", require_initiator: true),
     PERMISSIONS_REMOVED                       = LogEntry.new(id: 36, description: "Permissions removed", require_initiator: true),
+    ACCOUNT_INVITED                           = LogEntry.new(id: 37, description: "Account was invited", require_initiator: true),
   ]
 
   EVENTS_REQUIRING_INITIATOR   = EVENTS.select(&:require_initiator?)
@@ -80,6 +81,10 @@ class EventLog < ActiveRecord::Base
 
   def self.record_role_change(user, previous_role, new_role, initiator)
     record_event(user, ROLE_CHANGED, initiator: initiator, trailing_message: "from #{previous_role} to #{new_role}")
+  end
+
+  def self.record_account_invitation(user, initiator)
+    record_event(user, ACCOUNT_INVITED, initiator: initiator)
   end
 
   def self.for(user)
--- a/app/models/event_log.rb
+++ b/app/models/event_log.rb
@@ -1,3 +1,4 @@
+require "resolv"
 class EventLog < ActiveRecord::Base
   deprecated_columns :event
 
@@ -46,7 +47,7 @@ class EventLog < ActiveRecord::Base
   EVENTS_REQUIRING_INITIATOR   = EVENTS.select(&:require_initiator?)
   EVENTS_REQUIRING_APPLICATION = EVENTS.select(&:require_application?)
 
-  VALID_OPTIONS = [:initiator, :application, :application_id, :trailing_message]
+  VALID_OPTIONS = [:initiator, :application, :application_id, :trailing_message, :ip_address, :user_agent_id]
 
   validates :uid, presence: true
   validates_presence_of :event_id
@@ -56,6 +57,9 @@ class EventLog < ActiveRecord::Base
 
   belongs_to :initiator, class_name: "User"
   belongs_to :application, class_name: "Doorkeeper::Application"
+  belongs_to :user_agent
+
+  delegate :user_agent_string, to: :user_agent
 
   def event
     entry.description
@@ -65,7 +69,19 @@ class EventLog < ActiveRecord::Base
     EVENTS.detect { |event| event.id == event_id }
   end
 
+  def ip_address_string
+    convert_integer_to_ip_address(self.ip_address)
+  end
+
   def self.record_event(user, event, options = {})
+    if options[:ip_address]
+      if options[:ip_address] =~ Resolv::IPv6::Regex
+        Raven.capture_message("Received IP address: #{options[:ip_address]}. IPv6 logging is not supported yet")
+        options[:ip_address] = nil
+      else
+        options[:ip_address] = convert_ip_address_to_integer(options[:ip_address])
+      end
+    end
     attributes = {
       uid: user.uid,
       event_id: event.id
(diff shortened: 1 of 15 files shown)


test/integration/event_log_test.rb
```diff
@@ -237,6 +237,34 @@ class EventLogIntegrationTest < ActionDispatch::IntegrationTest
     assert page.has_content?("You do not have permission to perform this action")
   end
 
+  test "record user's login ip address" do
+    visit root_path
+    signin_with(@user)
+
+    ip_address = @user.event_logs.last.trailing_message.split.first
+    assert ip_address.match(/\A[[:alnum:].:]+$/)
+  end
```
h-lame: It would be good if this test could assert on an actual ip address value, but I don't know if it's possible to stub the request ip address somehow so we have a known value to check against.

test/integration/event_log_test.rb
```diff
@@ -237,6 +237,34 @@ class EventLogIntegrationTest < ActionDispatch::IntegrationTest
...
+
+  test "record user's login browser" do
+    visit root_path
+    signin_with(@user)
+
+    browser = @user.event_logs.last.trailing_message.split.last
+    assert browser.match(/\A[[:alpha:]]+/)
+  end
```
h-lame: As above, it'd be good to assert against an actual browser value here instead of just saying the 2nd part of the trailing message is all word characters.

test/integration/event_log_test.rb
```diff
@@ -237,6 +237,34 @@ class EventLogIntegrationTest < ActionDispatch::IntegrationTest
...
+  test "record who the account was created by" do
+    visit root_path
+    signin_with(@admin)
+    visit users_path
+    click_on "Create user"
+    fill_in 'Name', with: 'New User'
+    fill_in 'Email', with: 'test@test.com'
+    click_on "Create user and send email"
+
+    assert_equal EventLog::ACCOUNT_INVITED, User.last.event_logs.first.entry
```
h-lame: Could we also assert that the event log created has the`@admin` user associated with it as the initiator, not just that we've got an `ACCOUNT_INVITED` event.  Knowing who invited them is one of the key things we want to know after all, so we should check it if we can.

test/models/event_log_test.rb
```diff
@@ -104,4 +104,12 @@ class EventLogTest < ActiveSupport::TestCase
     assert_equal log.entry, EventLog::PASSPHRASE_RESET_REQUEST
     assert_not_nil log.created_at
   end
+
+  test "can record an account invitation" do
+    user = create(:user)
+    admin = create(:admin_user)
+
+    event_log = EventLog.record_account_invitation(user, admin)
+    assert_equal EventLog::ACCOUNT_INVITED, event_log.entry
```
h-lame: Again it'd be nice to assert that the `initiator` is set properly with the 2nd argument to the helper.

sihugh:
Added a DNM banner until we've checked this through with currently unavailable people :-)

DennisTsiang:
Updated the commit with the changes @h-lame mentioned

db/migrate/20170905104611_create_ip_addresses.rb
```diff
@@ -0,0 +1,10 @@
+class CreateIpAddresses < ActiveRecord::Migration
```
boffbowsh: In thinking about this a bit more, I think it makes more sense to store the [IP addresses as integers](http://www.boffbowsh.co.uk/2012/05/22/serving-memcached-pages-from-nginx.html) on the `EventLog` record. We shouldn't be storing them as `VARCHAR`s anyway, so once you have an integer pointing to another table of integers we should just cut out the middleperson.

db/migrate/20170905111153_create_user_agents.rb
```diff
@@ -0,0 +1,10 @@
+class CreateUserAgents < ActiveRecord::Migration
+  def change
+    create_table :user_agents do |t|
+      t.string :user_agent,           default: "", :limit => 1000
```
boffbowsh: I don't think this makes sense to be not nullable with an empty default. Either there is a user agent string or there isn't.

db/migrate/20170905111153_create_user_agents.rb
```diff
@@ -0,0 +1,10 @@
...
+      t.timestamps null: false
```
boffbowsh: I don't think this table needs timestamps. All we're doing is recording the time the user agent was first seen, which we can find from the `EventLog`.

app/models/user.rb
```diff
@@ -95,6 +95,9 @@ def prompt_for_2sv?
 
   def event_logs
     EventLog.where(uid: uid).order(created_at: :desc)
+      .joins("LEFT JOIN ip_addresses ON ip_addresses.id = event_logs.ip_address_id")
+      .joins("LEFT JOIN user_agents ON user_agents.id = event_logs.user_agent_id")
```
boffbowsh: These joins should be doable using just ActiveRecord, you shouldn't need to write SQL for it

app/models/user.rb
```diff
@@ -95,6 +95,9 @@ def prompt_for_2sv?
...
+      .select('event_logs.*, ip_addresses.ip_address as ip_address, user_agents.user_agent as user_agent')
```
boffbowsh: It'd be better to use [`Forwardable`](http://ruby-doc.org/stdlib-2.0.0/libdoc/forwardable/rdoc/Forwardable.html) to delegate the `user_agent` attribute to the joined object.

db/migrate/20170905104611_create_ip_addresses.rb
h-lame: Do we need to worry about ipv6 addresses?

boffbowsh: I... don't believe so. What does `remote_ip` return.

DennisTsiang: When I was testing it locally, it could return either ipv4 and ipv6

DennisTsiang:
Updated the commit with the comments you left @boffbowsh , and with the latest changes from master. Does this look better?

db/migrate/20170905104611_create_ip_addresses.rb
boffbowsh: I don't know how long it'll be until we're supporting IPv6 in our stack. I think it makes sense to only handle the IPv4 case for now. If we get an IPv6 address, record it as a NULL and log an error to Sentry.

app/models/event_log.rb
```diff
@@ -92,4 +107,12 @@ def validate_event_mappable
       errors.add(:event_id, "must have a corresponding `LogEntry` for #{event_id}")
     end
   end
+
+  def self.inet_aton ip
```
boffbowsh: Could give these functions friendlier names.

app/models/event_log.rb
```diff
@@ -64,7 +68,14 @@ def entry
     EVENTS.detect { |event| event.id == event_id }
   end
 
+  def ip_address_string
+    inet_ntoa self.ip_address
+  end
+
   def self.record_event(user, event, options = {})
+    if options[:ip_address]
+      options[:ip_address] = inet_aton options[:ip_address]
```
boffbowsh: We prefer to call functions in this way with parentheses.
```ruby
options[:ip_address] = inet_aton(options[:ip_address])
```

We should probably add this to the styleguide and Rubocop.

DennisTsiang:
Thanks for reviewing that @boffbowsh , I attempted the changes you requested. Let me know what you think.

PR #566, opened 2017-09-11
Add user agent table and new columns to event logs table

DennisTsiang:
Event logs table has has two new columns for storing ip addresses
and foreign key user_agent_id, which references a new user agent
table. These two tables will be used together to display extra
logging information of a user signing in. Gem 'lhm' is used to
perform the migration without locking the database, which is
necessary as the event logs table has over 18M records.

Follows on from #565

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -159,6 +159,7 @@ GEM
       actionpack (>= 3.0.0)
       activesupport (>= 3.0.0)
     kgio (2.11.0)
+    lhm (2.2.0)
     link_header (0.0.8)
     logstash-event (1.1.5)
     logstasher (0.4.8)
@@ -381,6 +382,7 @@ DEPENDENCIES
   jasmine (= 2.1.0)
   json (= 1.8.3)
   kaminari (~> 0.16.3)
+  lhm
   logstasher (= 0.4.8)
   minitest (~> 5.8.0)
   mocha (= 1.1.0)
--- a/db/migrate/20170905111153_create_user_agents.rb
+++ b/db/migrate/20170905111153_create_user_agents.rb
@@ -1,0 +1,7 @@
+class CreateUserAgents < ActiveRecord::Migration
+  def change
+    create_table :user_agents do |t|
+      t.string :user_agent_string,    index: true, null: false, :limit => 1000
+    end
+  end
+end
--- a/db/migrate/20170908160100_add_ip_address_and_user_agent_id_to_event_logs.rb
+++ b/db/migrate/20170908160100_add_ip_address_and_user_agent_id_to_event_logs.rb
@@ -1,0 +1,21 @@
+require 'lhm'
+
+class AddIpAddressAndUserAgentIdToEventLogs < ActiveRecord::Migration
+  def self.up
+    Lhm.cleanup(:run)
+    Lhm.change_table :event_logs do |m|
+      m.add_column :ip_address, "BIGINT"
+      m.add_column :user_agent_id, "INT"
+      m.ddl("ALTER TABLE %s ADD CONSTRAINT event_logs_user_agent_id_fk FOREIGN KEY (user_agent_id) REFERENCES user_agents(id)" % m.name)
+    end
+  end
+
+  def self.down
+    remove_foreign_key :event_logs, name: :event_logs_user_agent_id_fk
+    Lhm.cleanup(:run)
+    Lhm.change_table :event_logs do |m|
+      m.remove_column :user_agent_id
+      m.remove_column :ip_address
+    end
+  end
+end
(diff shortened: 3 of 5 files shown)


db/migrate/20170908145201_add_index_to_user_agent.rb
```diff
@@ -0,0 +1,15 @@
...
+      m.add_index [:user_agent_string]
+    end
+  end
+
+  def self.down
+    Lhm.change_table :user_agents do |m|
+      m.remove_index [:user_agent_string]
+    end
+  end
+end
```
h-lame: We won't need to use `Lhm` for this migration, as it's adding an index to user_ages which is an empty table at this point.  We can even do it in the  `CreateUserAgents` migration too.

db/migrate/20170905111153_create_user_agents.rb
```diff
@@ -0,0 +1,7 @@
+class CreateUserAgents < ActiveRecord::Migration
+  def change
+    create_table :user_agents do |t|
+      t.string :user_agent_string,           :limit => 1000
```
h-lame: Following on from the comment on the `AddIndexToUserAgent` migration, we can add the index here directly with the `index: true` option to the column definition.  I think it'll be useful to mark the column as not nullable by passing `null: false` as well.

h-lame: http://edgeguides.rubyonrails.org/active_record_migrations.html#column-modifiers is a useful resource for migrations and the specific options you can pass when creating columns.  (The whole guide is useful, but that section in particular is useful here).

DennisTsiang:
Updated that commit with the minor changes requested

db/migrate/20170908160100_add_ip_address_and_user_agent_id_to_event_logs.rb
```diff
@@ -0,0 +1,21 @@
...
+      m.add_column :ip_address, "BIGINT"
+      m.add_column :user_agent_id, "INT"
+      m.ddl("ALTER TABLE %s ADD CONSTRAINT event_logs_user_agent_id_fk FOREIGN KEY (user_agent_id) REFERENCES user_agents(id)" % m.name)
+    end
+  end
+
+  def self.down
+    remove_foreign_key :event_logs, name: :event_logs_user_agent_id_fk
+    Lhm.change_table :event_logs do |m|
+      Lhm.cleanup(:run)
```
h-lame: Does the cleanup need to happen outside the change_table block, or inside?  In the `up` it's outside, but in the `down` it's inside.  Probably best for it to be consistent.

db/schema.rb
```diff
@@ -51,9 +51,23 @@
...
+  create_table "lhma_2017_09_11_10_42_28_471_event_logs", force: :cascade do |t|
+    t.string   "uid",              limit: 255, null: false
+    t.datetime "created_at",                   null: false
+    t.integer  "initiator_id",     limit: 4
+    t.integer  "application_id",   limit: 4
+    t.string   "trailing_message", limit: 255
+    t.integer  "event_id",         limit: 4
+  end
+
+  add_index "lhma_2017_09_11_10_42_28_471_event_logs", ["uid", "created_at"], name: "lhm_index_event_logs_on_uid_and_created_at", using: :btree
```
h-lame: We shouldn't be adding this to the schema.  My guess is that you've run this migration locally and rails has picked it up from your dev DB when it generated the schema.  It's not unusual for a dev DB to have a few local changes that we don't want to commit, so we need to be careful when adding `schema.rb` to git to make sure that we don't add more than we expect.

Gemfile.lock
```diff
@@ -411,4 +413,4 @@ DEPENDENCIES
   zeroclipboard-rails (>= 0.1.2)
 
 BUNDLED WITH
-   1.15.1
+   1.15.4
```
h-lame: I don't think we're using bundler 1.15.4 yet (looks like we're on 1.15.1 on all versions of ruby according to puppet: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_rbenv/manifests/all.pp) so we shouldn't bump this.  It's not a problem if you've locally upgraded your bundler, but it does mean you're not using the same version as the rest of the devs, or the ones on our deployed environments so you should take care when committing the `Gemfile.lock` to check that if BUNDLED_WITH is changed, it's in keeping with the version of bundler that is used on our servers.

db/migrate/20170908160100_add_ip_address_and_user_agent_id_to_event_logs.rb
DennisTsiang: I don't think it matters but I'll put them both outside for consistency

DennisTsiang:
Removed the unwanted auto-generated LHM table in db/schema.rb and bundled-with change in Gemfile.lock

PR #568, opened 2017-09-22
Add super organisation admin role

elenatanasoiu:
For: https://trello.com/c/QXKlEEYa/226-create-super-org-admin-permissions-in-signon

## WHAT
This will create a `super_organisation_admin` role which will have the same permissions as an `organisation_admin`, but in addition will also be granted access these permissions for child organisations. 

## WHY
Building on the work of devolving unsuspend permissions to publishers, we want to extend this to being able to unsuspend their child accounts.

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -10,7 +10,7 @@ module UsersHelper
   end
 
   def organisation_select_options
-    { include_blank: is_org_admin? ? false : 'None' }
+    { include_blank: (is_org_admin? || is_super_org_admin?) ? false : 'None' }
   end
 
   def user_email_tokens(user = current_user)
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -26,7 +26,11 @@ module UsersHelper
   end
 
   def is_org_admin?
-    current_user.role == 'organisation_admin'
+    current_user.organisation_admin?
+  end
+
+  def is_super_org_admin?
+    current_user.super_organisation_admin?
   end
 
   def sync_needed?(permissions)
--- a/app/policies/base_policy.rb
+++ b/app/policies/base_policy.rb
@@ -12,7 +12,11 @@ class BasePolicy
 
   protected
 
-  def belong_to_same_organisation_subtree?
+  def record_in_own_organisation?
+    record.organisation && (record.organisation_id == current_user.organisation_id)
+  end
+
+  def record_in_child_organisation?
     current_user.organisation.subtree.include?(record.organisation)
   end
 
--- a/app/policies/base_policy.rb
+++ b/app/policies/base_policy.rb
@@ -12,8 +12,8 @@ class BasePolicy
 
   protected
 
-  def belong_to_same_organisation_subtree?(current_user, record)
-    current_user.organisation.subtree.pluck(:id).include?(record.organisation_id.to_i)
+  def belong_to_same_organisation_subtree?
+    current_user.organisation.subtree.include?(record.organisation)
   end
 
   class Scope
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -19,7 +19,7 @@ class UserPolicy < BasePolicy
     when 'admin'
       can_manage?
     when 'super_organisation_admin'
-      allow_self_only || (can_manage? && record_in_own_organisation?)
+      allow_self_only || (can_manage? && (record_in_own_organisation? || record_in_child_organisation?))
     when 'organisation_admin'
       allow_self_only || (can_manage? && record_in_own_organisation?)
     else # 'normal'
@@ -69,21 +69,15 @@ private
     Roles.const_get(current_user.role.classify).can_manage?(record.role)
   end
 
-  def record_in_own_organisation?
-    record.organisation && (record.organisation_id == current_user.organisation_id)
-  end
-
-  def record_in_child_organisation?
-    current_user.organisation.subtree_ids.include?(record.organisation_id)
-  end
-
   class Scope < ::BasePolicy::Scope
     def resolve
       if current_user.superadmin?
         scope.web_users
       elsif current_user.admin?
         scope.web_users.where(role: current_user.manageable_roles)
-      elsif current_user.super_organisation_admin? || current_user.organisation_admin?
+      elsif current_user.super_organisation_admin?
+        scope.web_users.where(role: current_user.manageable_roles).where(organisation: current_user.organisation.subtree)
+      elsif current_user.organisation_admin?
         scope.web_users.where(role: current_user.manageable_roles).where(organisation: current_user.organisation)
       else
         scope.none
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -1,10 +1,10 @@
 class UserPolicy < BasePolicy
   def index?
-    current_user.superadmin? || current_user.admin? || current_user.organisation_admin?
+    %w(superadmin admin super_organisation_admin organisation_admin).include? current_user.role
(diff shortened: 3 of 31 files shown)


app/helpers/users_helper.rb
```diff
@@ -26,7 +26,11 @@ def edit_user_path_by_user_type(user)
   end
 
   def is_org_admin?
-    current_user.role == 'organisation_admin'
+    current_user.organisation_admin?
+  end
+
+  def is_super_org_admin?
+    current_user.super_organisation_admin?
```
h-lame: This method is used in some of the views in the previous commit so we probably want to add it there, not here.

app/helpers/users_helper.rb
```diff
@@ -26,7 +26,11 @@ def edit_user_path_by_user_type(user)
   end
 
   def is_org_admin?
-    current_user.role == 'organisation_admin'
+    current_user.organisation_admin?
```
h-lame: :+1: for using the predicate instead of string comparison.

app/policies/user_policy.rb
```diff
@@ -76,7 +78,9 @@ def resolve
       if current_user.superadmin?
         scope.web_users
       elsif current_user.admin?
-        scope.web_users.where(role: %w(admin organisation_admin normal))
+        scope.web_users.where(role: %w(admin super_organisation_admin organisation_admin normal))
```
h-lame: Could we use `where(role: current_user.role.manageable_roles)` here (and in other branches of the condition).  I don't think we're making much use of `manageable_roles` and it makes sense to try and keep that stuff as centralised as possible if we can.

spec/policies/organisation_policy_spec.rb
```diff
@@ -4,10 +4,11 @@
   subject { described_class }
 
   permissions :index? do
-    it "is forbidden to organisation admins and normal users" do
+    it "is forbidden to super org admins, organisation admins and normal users" do
```
h-lame: nitpick: `org` => `organisation` for consistency (I'd probably have done the same thing though).

app/policies/user_policy.rb
```diff
@@ -80,7 +80,7 @@ def resolve
       elsif current_user.admin?
         scope.web_users.where(role: %w(admin super_organisation_admin organisation_admin normal))
       elsif current_user.super_organisation_admin?
-        scope.web_users.where(role: %w(super_organisation_admin organisation_admin normal)).where(organisation_id: current_user.organisation_id)
+        scope.web_users.where(role: %w(super_organisation_admin organisation_admin normal)).where(organisation_id: child_organisation_ids)
```
h-lame: Hyperoptimisation: if you did `where(organisation: current_user.organisation.subtree)` you'd avoid the SQL query from `child_organisation_ids`:

`User.where(organisation_id: organisation.subtree.pluck(:id))` #=> `SELECT organisations.id FROM organisations WHERE (organisations.id = 6 or organisations.ancestry like '6/%' or organisations.ancestry = '6')` and
`SELECT users.* FROM users WHERE users.organisation_id IN (2, 20, [SNIP: A BUNCH OF IDS], 1100, 6)`

vs.

`User.where(organisation: organisation.subtree)` #=> `SELECT users.* FROM users WHERE users.organisation_id IN (SELECT organisations.id FROM organisations WHERE (organisations.id = 6 or organisations.ancestry like '6/%' or organisations.ancestry = '6'))`

Rails knows how to merge ActiveRelation objects into queries without resolving them to their results first so you can avoid queries.

test/models/user_test.rb
```diff
@@ -298,14 +298,14 @@ def setup
...
-    assert_equal "can't be 'None' for an Organisation admin", user.errors[:organisation_id].first
+    assert_equal "can't be 'None' for Organisation Admin", user.errors[:organisation_id].first
   end
 
   test "super organisation admin must belong to an organisation" do
     user = build(:user, role: 'super_organisation_admin', organisation_id: nil)
 
     refute user.valid?
-    assert_equal "can't be 'None' for a Super Organisation admin", user.errors[:organisation_id].first
+    assert_equal "can't be 'None' for Super Organisation Admin", user.errors[:organisation_id].first
```
h-lame: This change belongs in a previous commit as you didn't change the error message in this commit.

spec/policies/organisation_policy_spec.rb
```diff
@@ -86,8 +86,13 @@
     context 'for super org admins' do
       let(:user) { create(:super_org_admin, organisation: org_root_one) }
 
-      it 'is empty' do
-        expect(resolved_scope).to be_empty
+      it 'includes own org and child orgs' do
```
h-lame: I don't know if this is true.  The `index?` method for OrganisationPolicy suggests that only super admins and admins can see a list of organisations, so I think the scope should match that and return nothing.  Or! the `index?` method  should return true for super org admins to match the change you've made here.

I suspect however, changing the `index?` to return `true` will have more far reaching repercussions than changing the scope to return `[]`, as I suspect it's used to control what controller and view actions you're allowed to see, and I don't think we want to let super org admins see a list of orgs (but then again, maybe we do - they can't edit those orgs, but it might be useful for them to see the list of orgs they're able to manage users for).

spec/policies/organisation_policy_spec.rb
h-lame: You might want to check the rest of the PR for similar `org` vs. `organisation` changes too.  FWIW: I think `org` is fine in variable names, but we should probably use `organisation` in test names.

spec/policies/user_policy_spec.rb
```diff
@@ -86,6 +92,72 @@
...
+        it "is forbidden for super org admins to access org admins from other organisations" do
+          expect(subject).to permit(super_org_admin, build(:organisation_admin, organisation: super_org_admin.organisation))
+        end
+
+        it "is forbidden for super org admins to access superadmins and admin users from either their own or other organisations" do
+          expect(subject).not_to permit(super_org_admin, build(:superadmin_user))
+          expect(subject).not_to permit(super_org_admin, build(:admin_user))
+        end
+      end
+    end
```
h-lame: I think we're missing a test that shows a super org admin is allowed to do things to users within their org's subtree.

h-lame: And maybe a corresponding test in the org admin tests that make sure they can't do things to users within their org's subtree.  It has "forbidden to access other orgs" but we could do "forbidden to access child orgs" to be explicit.

spec/policies/user_policy_spec.rb
```diff
@@ -259,6 +343,41 @@
       end
     end
 
+    context 'for super org admins' do
+      let(:user) { create(:super_org_admin, organisation: org_root_one) }
+
+      it 'includes only super org admins and users belonging to their org' do
```
h-lame: Maybe combine or reword this test and the one following it.  `includes only...` suggests this test tells us the complete set of users that are included, but the next test says `does include...` which sort of breaks that.

spec/policies/organisation_policy_spec.rb
elenatanasoiu: I've made the scope match the `index?` method (returns `scope.none`)

elenatanasoiu:
Added requested changes

app/policies/base_policy.rb
```diff
@@ -12,7 +12,11 @@ def scope
 
   protected
 
-  def belong_to_same_organisation_subtree?
+  def record_in_own_organisation?
+    record.organisation && (record.organisation_id == current_user.organisation_id)
+  end
+
+  def record_in_child_organisation?
```
h-lame: Nice rename here!

PR #571, opened 2017-09-26
Model HMCTS correctly within Signon

elenatanasoiu:
For: https://trello.com/c/c3d52nTH/228-model-hmcts-correctly-within-signon

## WHAT
Checks that the ancestry for HMCTS is modeled correctly. If it finds that some of HMCTS' children are not properly assigned, it re-assigns them.

| Parent                         | Children                             |
| ------------------------------ | ------------------------------------ |
| HM Courts & Tribunals Service  | Bankruptcy Court                     |
|                                | First-tier Tribunal (Mental Health)  |
|                                | ...                                  |

Also, "HM Courts Service" was a closed organisation with no parent. It will now be assigned to "HM Courts & Tribunals Service" together will all its other children. 

## WHY 

Building on the work of devolving responsibilities to Organisation Admins and Super Organisation Admins, we want to extend this to being able to have responsibilities on all org's child accounts.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170908160100) do
+ActiveRecord::Schema.define(version: 20170926101024) do
 
   create_table "batch_invitation_application_permissions", force: :cascade do |t|
     t.integer  "batch_invitation_id",     limit: 4, null: false
--- a/db/migrate/20170926101024_model_hmcts_org_structure.rb
+++ b/db/migrate/20170926101024_model_hmcts_org_structure.rb
@@ -1,0 +1,98 @@
+class ModelHmctsOrgStructure < ActiveRecord::Migration
+  def up
+    # Make sure parent of HMCTS is MOJ
+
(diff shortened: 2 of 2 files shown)


db/schema.rb
```diff
@@ -11,7 +11,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170908160100) do
+ActiveRecord::Schema.define(version: 20170926101024) do
+
+  add_index "_event_logs_old", ["uid", "created_at"], name: "index_event_logs_on_uid_and_created_at", using: :btree
```
h-lame: I don't think any of these other changes should be in the schema.rb.  I don't know why they're in the production dump, but they're not expected as part of the app, and certainly weren't introduced by this PR.  We might want to write a separate card to explore where these non-migration changes came from and either introduce them to schema.rb, or remove them from the production db.

elenatanasoiu: The reason I didn't remove them was because they were necessary to have a complete schema so keeping them and committing them seemed like the most useful thing to do. However, these changes can also be added later as part of a different PR. 

I've added a card here: https://trello.com/c/WzntS0j7/213-signon-fix-schema

elenatanasoiu:
Added check for MOJ to be parent of HMCTS

PR #575, opened 2017-10-06
Adds migrations to clean up schema

elenatanasoiu:
For: https://trello.com/c/WzntS0j7/251-signon-fix-schema

Whenever a new migration gets created and run, 2 extra tables get added in the schema: 

1. A table starting with `lhm_<date>`. A migration was run using the LHM gem which is meant to be used for large databases with millions of rows. However, after using it, `lhm.cleanup` needs to be run in order to drop any temporary tables that it might have created in order to ensure zero downtime.

2. An `_old_event_logs` table which was not dropped properly

In addition to that, we also get a "deleted_at" field in the `oauth_applications` table, as well as an index on that column. This is a residue from when it was attempted to implement soft delete on doorkeeper tables using the `acts_as_paranoid` gem. This work was reverted, but the column was not dropped. We're correcting that in this PR.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171020132253) do
+ActiveRecord::Schema.define(version: 20171023161055) do
 
   create_table "batch_invitation_application_permissions", force: :cascade do |t|
     t.integer  "batch_invitation_id",     limit: 4, null: false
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -88,17 +87,19 @@
...
-    t.datetime "updated_at",                                       null: false
+    t.string   "uid",                   limit: 255,                 null: false
+    t.string   "secret",                limit: 255,                 null: false
+    t.string   "redirect_uri",          limit: 255,                 null: false
+    t.datetime "created_at",                                        null: false
+    t.datetime "updated_at",                                        null: false
     t.string   "home_uri",              limit: 255
     t.string   "description",           limit: 255
     t.boolean  "supports_push_updates",             default: true
+    t.time     "deleted_at"
```
h-lame: We've still got this column, and the index on it, being added - we should get to the bottom of why these are still here and work out if they should be or shouldn't.  The right thing might be that we want these fields in the schema because they're in the production DB, but it'd be nice to investigate and have a separate commit that adds them with a message that explicitly addresses where they came from and why we're adding them without a migration.

PR #576, opened 2017-10-18
Add new short-url-manager permission.

danielroseman:
The functionality to receive notifications when a new short URL is
requested was split off into the new `receive_notifications`
permission in https://github.com/alphagov/short-url-manager/pull/99.

This migration creates the permission and assigns it to
all users who have the `manage_short_urls` permission.

### DIFF
--- a/db/migrate/20171017152038_add_short_url_manager_permission.rb
+++ b/db/migrate/20171017152038_add_short_url_manager_permission.rb
@@ -1,0 +1,15 @@
+class AddShortUrlManagerPermission < ActiveRecord::Migration
+  def up
+    app = Doorkeeper::Application.find_by(name: "Short URL Manager")
+    manage_short_urls = SupportedPermission.find_by(application: app, name: "manage_short_urls")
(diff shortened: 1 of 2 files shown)


db/migrate/20171017152038_add_short_url_manager_permission.rb
```diff
@@ -0,0 +1,15 @@
...
+    manage_short_urls = SupportedPermission.find_by(application: app, name: "manage_short_urls")
+    receive_notifications = SupportedPermission.create!(application: app, name: "receive_notifications")
+    manage_short_urls.user_application_permissions.each do |uap|
+      UserApplicationPermission.create!(user: uap.user, supported_permission: receive_notifications)
+    end
+  end
+
+  def down
+    app = Doorkeeper::Application.find_by(name: "Short URL Manager")
+    SupportedPermission.find_by(application: app, name: "receive_notifications").destroy
```
cbaines: I checked, as I worry about referential integrity, and it looks like this will cause the deletion of the related UserApplicationPermissions, which seems like what you'd want to happen 👍

PR #577, opened 2017-10-20
Fix logging of password reset failures

suzannehamilton:
Always log the validation errors when a password reset fails. Before, these were only being logged if the user's details had been persisted to the database. I suspect this was never true, because the record had just failed validation and so was never in a persisted state when this check was run.

We're currently trying to debug why a user cannot log into Signon, and these event logs are missing even though the user has reported validation errors and we can see that they have visited the password reset page.

It's not clear from the original commit (fe7ff13) why this check was necessary.

### DIFF
--- a/app/controllers/passwords_controller.rb
+++ b/app/controllers/passwords_controller.rb
@@ -27,7 +27,7 @@ class PasswordsController < Devise::PasswordsController
       if resource.valid?
         record_password_reset_success(resource)
       else
-        record_password_reset_failure(resource) if resource.persisted?
+        record_password_reset_failure(resource)
       end
     end
   end


suzannehamilton:
Here's what it looks like in the logs:

screen shot 2017-10-20 at 13 03 38

suzannehamilton:
I'm going to merge this to see if it helps us debug the problem for this user. I've tested valid and invalid password resets and haven't seen any problems so I think it's safe to deploy.

@rubenarakelyan - let me know if you have any suggestions or ideas about why that line was there.

rubenarakelyan:
I have no idea why that condition was there but it makes sense to remove it since invalid data will never be persisted anyway. I wonder if it was an attempt to filter out errors other than ones to do with invalid data...

rubenarakelyan:
I have no idea why the condition was there but it makes sense to remove it since invalid data will never be persisted anyway. I wonder if it was an attempt to filter out errors other than ones to do with invalid data...

PR #589, opened 2017-11-20
Upgrade to Rails 5.1.4

danielburnley:
WIP for now but I have a few questions: 

- In https://github.com/alphagov/signon/pull/589/commits/a13eefa88cbc9a116e4ca264b52fd20e5b694a76 you are required to set a `fallback_location` - what should this be?
- There's a lot of converting ActionController::Parameters to hashes and back again when dealing with the `UserParameterSanitiser`- Would it be better to get  the sanitiser to handle the conversion, keep it within controllers, or add another helper for this?


app/controllers/invitations_controller.rb
```diff
@@ -80,7 +80,7 @@ def resource_params
   # https://github.com/plataformatec/devise/blob/v2.2/app/controllers/devise_controller.rb#L99
   # for details :)
   def unsanitised_user_params
-    params.fetch(:user, {})
+    params.require(:user).permit(:name, :email, :organisation_id, :invitation_token, :password, :password_confirmation, :require_2sv, :role).to_h
```
RoryMacDonald: Worth checking the GDS style guide for this. Imagine this will cause a linting error

app/controllers/users_controller.rb
```diff
@@ -58,7 +58,7 @@ def unlock
     EventLog.record_event(@user, EventLog::MANUAL_ACCOUNT_UNLOCK, initiator: current_user)
     @user.unlock_access!
     flash[:notice] = "Unlocked #{@user.email}"
-    redirect_to :back
+    redirect_back(fallback_location: root_path)
```
RoryMacDonald: What is root_path for signon? Imagine this should be the GOV.UK Homepage?

app/controllers/users_controller.rb
```diff
@@ -79,7 +79,7 @@ def cancel_email_change
     @user.unconfirmed_email = nil
     @user.confirmation_token = nil
     @user.save(validate: false)
-    redirect_to :back
+    redirect_back(fallback_location: root_path)
```
RoryMacDonald: As per above

app/controllers/users_controller.rb
```diff
@@ -199,7 +199,7 @@ def allow_no_application_access
 
   def user_params
     UserParameterSanitiser.new(
-      user_params: params.require(:user),
+      user_params: params.require(:user).permit(:user, :name, :email, :organisation_id, :role, { supported_permission_ids: [] }).to_h,
```
RoryMacDonald: As per above

test/integration/api_authentication_test.rb
```diff
@@ -5,7 +5,8 @@ class ApiAuthenticationTest < ActionDispatch::IntegrationTest
   include TokenAuthSupport
 
   def access_user_endpoint(token = nil, params = {})
-    get "/user.json", params, token.nil? ? {} : {"HTTP_AUTHORIZATION" => "Bearer #{token}"}
+    headers = token.nil? ? {} : {"HTTP_AUTHORIZATION" => "Bearer #{token}"}
+    get "/user.json", params: params, headers: headers
```
RoryMacDonald: Prefer this approach, but is this related to the upgrade?

test/models/lib/user_parameter_sanitiser_test.rb
```diff
@@ -24,31 +24,6 @@ class UserParameterSanitiserTest < ActiveSupport::TestCase
     end
   end
 
-  context "with ActionController params object" do
```
RoryMacDonald: How come these tests have been removed?

app/controllers/api_users_controller.rb
```diff
@@ -54,8 +54,13 @@ def load_and_authorize_api_user
 
   def api_user_params
     UserParameterSanitiser.new(
-      user_params: params.require(:api_user),
+      user_params: permitted_user_params(params.require(:api_user)).to_h,
```
RoryMacDonald: Unsure about the `to_h` here and in a couple of other places in this changeset. I'd be tempted to handle this type conversion elsewhere. Could potentially get `UserParameterSanitiser` to support the new Params type or introduce an object to go from params to hash and name it something, so it's clearer what is happening.

test/integration/api_authentication_test.rb
danielburnley: Headers had to be passed in as a hash - didn't think youd be allowed:

```get "user.json", params: params, headers: token.nil? {} : {"HTTP...etcetc"}```

File: test/models/lib/user_parameter_sanitiser_test.rb (hunk 6 above)
Comment (danielburnley) [reply test/models/lib/user_parameter_sanitiser_test.rb]: Mentioned in the commit message which removes them: https://github.com/alphagov/signon/pull/589/commits/a4212964fa9c83793a2281956892b1670dabb83c

File: app/controllers/api_users_controller.rb (hunk 7 above)
Comment (danielburnley) [reply app/controllers/api_users_controller.rb]: To convert to a hash you need to be sure the parameters have been permitted already - which is why i decided to do it in the controller. You could permit them but still have `UserParameterSanitiser` handle the conversion, but as that conversion removes non-permitted parameters it feels like it could be some hidden behavior not immediately obvious from within the controller

File: Gemfile [hunk 8]
```diff
@@ -71,17 +70,17 @@ gem 'logstasher', '0.4.8'
...
 group :test do
-  gem 'rspec-rails', '~> 3.3.3'
-  gem 'capybara', '~> 2.5.0'
+  gem 'rspec-rails', '~> 3.7.1'
+  gem 'rails-controller-testing'
+  gem 'capybara', '~> 2.7.0'
   gem 'capybara-email', '~> 2.4.0'
   gem 'poltergeist', '1.6.0'
-  gem 'database_cleaner', '1.5.0'
+  gem 'database_cleaner', '1.6.2'
```
NARKOZ: We probably can get rid of this gem. Rails 5.1 introduced system tests that share the same database connection: http://edgeguides.rubyonrails.org/5_1_release_notes.html#system-tests

RSpec 3.7 also supports system tests: http://rspec.info/blog/2017/10/rspec-3-7-has-been-released/

app/controllers/users_controller.rb
```diff
@@ -170,6 +170,10 @@ def any_filter?
       params[:two_step_status].present?
   end
 
+  def filter_params
+    params.permit(:filer, :role, :status, :organisation, :two_step_status)
+  end
```
NARKOZ: Why we need this method? I can't see it being used anywhere.

Gemfile
danielburnley: I can look into removing that, but potentially worth doing as a separate PR after this one is merged?

NARKOZ: Sure 👍

app/controllers/users_controller.rb
rubenarakelyan: `root_path` is the signon homepage

app/controllers/invitations_controller.rb
rubenarakelyan: Would be nice to break this into multiple lines for readability

rubenarakelyan:
LGTM

1. `root_path` is probably the best place to fall back to for `redirect_back` since there's no other more relevant place.
2. I like keeping the hash conversions within the controllers since that way it's explicit what's going on.

PR #591, opened 2017-11-23
Update migration

danielburnley:
This migration was introduced to fix an error when trying to load the
schema - but as it was done on foreign keys it didn't run correctly.

### DIFF
--- a/db/migrate/20171122121841_change_event_log_user_agent_id_to_integer.rb
+++ b/db/migrate/20171122121841_change_event_log_user_agent_id_to_integer.rb
@@ -1,9 +1,15 @@
 class ChangeEventLogUserAgentIdToInteger < ActiveRecord::Migration[5.1]
   def up
+    remove_foreign_key :event_logs, :user_agents
+    change_column :user_agents, :id, "BIGINT"
     change_column :event_logs, :user_agent_id, "BIGINT"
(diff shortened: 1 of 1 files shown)


db/migrate/20171122121841_change_event_log_user_agent_id_to_integer.rb
```diff
@@ -1,9 +1,15 @@
 class ChangeEventLogUserAgentIdToInteger < ActiveRecord::Migration[5.1]
   def up
+    remove_foreign_key :event_logs, :user_agents
+    change_column :user_agents, :id, "BIGINT"
     change_column :event_logs, :user_agent_id, "BIGINT"
+    add_foreign_key :event_logs, :user_agents
```
h-lame: This table is massive in production (18M rows) and running this migration will lock it for several minutes.  In the past when making changes to it we've used [lhm](https://github.com/soundcloud/lhm/) to avoid locking the tables (see: https://github.com/alphagov/signon/pull/566 and https://github.com/alphagov/signon/pull/567).

thomasleese:
I also wonder if we should revert #591 and get this commit into a new PR with all the upgrade changes because it means Signon is currently in an undeployable state.

danielburnley:
Yep makes sense to me :+1: probably better to make sure the changes are all in one merge too

PR #592, opened 2017-11-24
Revert "Upgrade to Rails 5.1.4"

danielburnley:
Reverts alphagov/signon#589

The migration in this PR was incorrect, and needed to be updated. To keep master in a deployable state I am reverting this for now and will create a new merge with the correct migration (currently in #591 )


cbaines:
> The migration in this PR was incorrect, and needed to be updated. To keep master in a deployable state I am reverting this for now and will create a new merge with the correct migration (currently in #591 )

I don't know much about the context of this, but please could you put this information in the commit.

kevindew:
@danielburnley Has this migration already run on Integration, will merging this leave integration in an inconsistent state?

Similarly have you checked this works ok on Integration?

danielburnley:
The original migration failed on staging so it was rolledback - as this is just a revert it will just put integration back to what it was yesterday prior to the merge

Jenkinsfile
```diff
@@ -47,7 +47,6 @@ node {
 
     for (adapter in DB_ADAPTERS) {
       stage("Set up the DB") {
-        sh("RAILS_ENV=test SIGNONOTRON2_DB_ADAPTER=${adapter} bin/rails db:environment:set")
```
kevindew: FYI you shouldn't reintroduce this. It will error if db doesn't exist and can be better skipped with an env var: https://github.com/alphagov/govuk-puppet/blob/master/modules/govuk_jenkins/files/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy#L154

db/migrate/20170908160100_add_ip_address_and_user_agent_id_to_event_logs.rb
```diff
@@ -5,7 +5,7 @@ def self.up
     Lhm.cleanup(:run)
     Lhm.change_table :event_logs do |m|
       m.add_column :ip_address, "BIGINT"
-      m.add_column :user_agent_id, "BIGINT"
+      m.add_column :user_agent_id, "INT"
```
kevindew: How come we've been editing a past migration?

danielburnley: This migration will fail under rails 5 - the default types for id fields have been changed to BIGINTS so the field types are mismatched. I can avoid editing it if you'd prefer that, but it will mean the migration is broken from Rails 5 onwards

Jenkinsfile
danielburnley: :+1: I'll remove that change in the updated PR

db/migrate/20170908160100_add_ip_address_and_user_agent_id_to_event_logs.rb
RoryMacDonald: IMO, we should change this via a new migration and also drop old migrations from the repository (don't see any value in having them in here if they have hit production already).

h-lame: In general I agree with removing old migrations, but in this case we have an AWS env that's not being deployed to as often so might need these migrations for longer than usual.  Also I think signon is special in that it doesn't get restored from staging/prod over night so we might have to do a thing to make sure the migration change actually goes through properly.

PR #593, opened 2017-11-24
Upgrade to Rails 5.1.4

danielburnley:
This re-introduces the Rails 5 upgrade which was reverted due to an change being required in the migration

https://trello.com/c/YN2Nu8cO/32-rails-upgrade-for-alphagov-signon


db/migrate/20171122121841_change_event_log_user_agent_id_to_integer.rb
```diff
@@ -1,9 +1,21 @@
 class ChangeEventLogUserAgentIdToInteger < ActiveRecord::Migration[5.1]
   def up
-    change_column :event_logs, :user_agent_id, "BIGINT"
+    Lhm.cleanup(:run)
+    Lhm.change_table :event_logs do |m|
+      m.remove_foreign_key :event_logs, :user_agents
+      m.change_column :user_agents, :id, "BIGINT"
+      m.change_column :event_logs, :user_agent_id, "BIGINT"
+      m.add_foreign_key :event_logs, :user_agents
+    end
```
kevindew: How come we need to do this? and not change any of the other id types / foreign keys?

danielburnley: When originally adding that foreign key, it was added by adding a column then specifying the key, so the type was explicitly set to `INT` ( https://github.com/alphagov/signon/blob/master/db/migrate/20170908160100_add_ip_address_and_user_agent_id_to_event_logs.rb ) - this needs to change now due to the default value of IDs changing to BIGINT - so db:schema:load fails due to the mismatched data type. 

This is the only place this error occurs

app/controllers/passwords_controller.rb
```diff
@@ -24,7 +24,7 @@ def require_no_authentication
 
   def update
     super do |resource|
-      if resource.valid?
+      if resource.errors.empty?
```
gpeng: Why was this change required?

config/application.rb
```diff
@@ -59,7 +61,5 @@ class Application < Rails::Application
     config.active_job.queue_adapter = :sidekiq
```
gpeng: You'll need to add `lib` to `eager_load_paths` in here to ensure `lib` is loaded when in production env (as per https://github.com/alphagov/manuals-frontend/pull/247) because https://blog.bigbinary.com/2016/08/29/rails-5-disables-autoloading-after-booting-the-app-in-production.html

app/controllers/passwords_controller.rb
danielburnley: (copied from https://github.com/alphagov/signon/pull/589/commits/be807797240debf08cee35d78bbb75b0c5640c36)

Due to a change when updating devise (https://github.com/plataformatec/devise/pull/4456/files)
blank passwords no longer get set on the resource when trying to reset
the password with a blank string. This change meant that the call to
resource.valid? now returned true rather than false, as the password was
nil - which is valid.

To fix this, the update method now checks for the presence of errors
rather than checking if the resource is valid.

Additionally, the error regarding password strength is no longer
displayed when submitting a blank password so that test has been removed

config/application.rb
danielburnley: :+1: Updated

app/controllers/passwords_controller.rb
gpeng: Eeuw... that's a bit unfortunate that it's not valid but it is. Not much else you can do I don't suppose.

db/migrate/20171122121841_change_event_log_user_agent_id_to_integer.rb
kevindew: Ah I see, yes - as that is the only foreign key.

Ok so I don't think we should be doing that via the method you have there. As that ends us in up a situation where running rake `db:setup` has a different database schema to what is on production and you'll hit same problems with any future foreign keys.

Rails actually has a system for this where it specifies id type as part of the create table argument: `create_table "batch_invitation_users", id: :integer`

If I run db:schema:dump with rails 5 on this I get this https://gist.github.com/kevindew/4de510ce86eff45622d1aa1d2d11ce09 which will load up without any need for migrations. Does however create a "lhma_2017_09_18_16_13_31_585_event_logs" table that may not be desirable.

kevindew:
Also, having a revert of a revert commit is pretty whacky. Could we just not re-introduce the commits that change things? it'd make things easier for us to review and leave a clearer git history

db/migrate/20171122121841_change_event_log_user_agent_id_to_integer.rb
danielburnley: It seems the `create_table, id: :integer` method is the simplest way to go :+1:

config/initializers/wrap_parameters.rb
```diff
@@ -5,10 +5,10 @@
 
 # Enable parameter wrapping for JSON. You can disable this by setting :format to an empty array.
 ActiveSupport.on_load(:action_controller) do
-  wrap_parameters format: [:json] if respond_to?(:wrap_parameters)
```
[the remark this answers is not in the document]
danielburnley: Probably not, it was one of the changes generated by rails app:update so I wasn't sure if the `if respond_to` was added by us or if that used to be the old default

PR #595, opened 2017-12-05
Upgrade to Ruby 2.4.2 & Bundle update

danielburnley:
Ahead of enabling Dependabot on the repository, we want to be using the latest
dependancies, to minimise the onslaught of incremental PRs.

https://trello.com/c/vX8qwh2X/46-ruby-upgrade-of-all-apps


h-lame:
Looks like there's no change in the Doorkeeper::AuthorizationsController between 4.2.0 and 4.2.6 as it doesn't appear in the files changed list on https://github.com/doorkeeper-gem/doorkeeper/compare/v4.2.0...v4.2.6.  There is a change in the specs for it (see: https://github.com/doorkeeper-gem/doorkeeper/compare/v4.2.0...v4.2.6#diff-71675d21b066882b49c6a7cb64f4507e) - so maybe we should import those to our `signin_required_authorizations_controller_test.rb` to make sure we don't miss anything.  Especially as they include rails 5 changes.

PR #647, opened 2018-01-11
Regenerate schema.rb using modern rails

h-lame:
Modern rails has changed the format of schema.rb so it's useful to
regenerate it as a separate commit/PR so that the next PR with a migration
doesn't incur a large diff-cost simply for adding a column.

This regenerated schema has revealed that we some of our tables still use
latin1 instead of utf8, but we'll fix that separately.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,6 @@
 #!/usr/bin/env groovy
 
 REPOSITORY = 'signon'
-DB_ADAPTERS = ['mysql', 'postgresql']
 repoName = JOB_NAME.split('/')[0]
 
 node {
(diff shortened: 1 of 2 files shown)


boffbowsh:
Let's stop testing against Postgres for now.

boffbowsh:
I wouldn't bother commenting it out, just remove it.

h-lame:
@boffbowsh of course - we won't go back down this path will we?  I'll excise the array stuff fully from the Jenkinsfile.  I'll leave the SIGNONOTRON2_DB_ADAPTER env var in the app though, as I suspect it'll be useful later when we try to migrate (and might be how some externals make their signon run against postgres).

PR #649, opened 2018-01-12
Allow super admins to batch create users for multiple orgs

h-lame:
For: https://trello.com/c/62ShH5HQ/298-spike-batch-creation-of-new-signon-accounts

This PR allows super admins to upload a CSV of users for batch inviting that contains organisation slugs for each user, letting them invite a load of users who belong to different organisations all in one go.

### DIFF
--- a/test/integration/batch_inviting_users_test.rb
+++ b/test/integration/batch_inviting_users_test.rb
@@ -3,50 +3,83 @@ require 'test_helper'
 class BatchInvitingUsersTest < ActionDispatch::IntegrationTest
   include ActiveJob::TestHelper
 
-  should "superadmin user can create users whose details are specified in a CSV file" do
-    application = create(:application)
-    user = create(:superadmin_user)
-
-    perform_batch_invite_with_user(user, application)
+  setup do
+    @cabinet_office = create(:organisation, slug: 'cabinet-office', name: 'Cabinet Office')
+    @department_of_hats = create(:organisation, slug: 'department-of-hats', name: 'Department of Hats')
+    @application = create(:application)
   end
 
-  should "admin user can create users whose details are specified in a CSV file" do
-    application = create(:application)
-    user = create(:admin_user)
+  context 'for superadmin users' do
+    setup do
+      @user = create(:superadmin_user)
+    end
 
-    perform_batch_invite_with_user(user, application)
+    should "allow creating users whose details are specified in a CSV file, assigning them all to one org" do
+      perform_batch_invite_with_user(@user, @application, organisation: @cabinet_office)
+      assert_user_created_and_invited('fred@example.com', @application, organisation: @cabinet_office)
+    end
+
+    should "allow creating users with org details specified in a CSV file, overriding the org selected in the form" do
+      perform_batch_invite_with_user(@user, @application, fixture_file: 'users_with_orgs.csv', user_count: 3, organisation: @cabinet_office)
+      assert_user_created_and_invited('fred@example.com', @application, organisation: @department_of_hats)
+      assert_user_created_and_invited('lara@example.com', @application, organisation: @cabinet_office)
+      assert_user_not_created('emma@example.com')
+    end
   end
 
-  should "not allow organisation admin user to create users whose details are specified in a CSV file" do
-    application = create(:application)
-    user = create(:user_in_organisation, role: 'organisation_admin')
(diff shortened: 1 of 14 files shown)


test/integration/batch_inviting_users_test.rb
```diff
@@ -3,50 +3,83 @@
...
+      @user = create(:admin_user)
+    end
+
+    should "allow creating users whose details are specified in a CSV file, assigning them all to one org" do
+      perform_batch_invite_with_user(@user, @application, organisation: @cabinet_office)
+      assert_user_created_and_invited('fred@example.com', @application, organisation: @cabinet_office)
+    end
 
-    perform_batch_invite_with_user(user, application)
+    should "ignore org details specified in a CSV file, creating all the users assigned to one org" do
```
suzannehamilton: I wonder if it will confuse people if we override the organisations specified in the CSV with the org selected in the UI. In particular, it seems odd that a superadmin would see an error if they tried to add emma@example.com, but a regular admin wouldn't.

Would it be less surprising to raise an error if the CSV lists a user with an organisation other than the admin's organisation?

h-lame: Yeah, I had that thought too; an admin can upload a file and process all the entries that a superadmin can't because the system ignores the org info for the admin, but not for the superadmin.  Maybe it should error before it tires to process the upload if there is any org info in the CSV when you're an admin.  Or, as you say, if the org info is there, but it's not the org you selected for the entire batch from the drop down on the upload a CSV page.

In theory however, an admin wouldn't know that they can upload a CSV with organisation info in it, because the UI doesn't explain that to them.  That's only explained in the UI for superadmins, so maybe it's fine?

Can't say I feel particularly strongly either way - do you have a preference?

h-lame:
Rebased now that #647 is merged to tidy up the `db/schema.rb` diff (and updated the PR description, now that paragraph is no longer relevant)

test/integration/batch_inviting_users_test.rb
elenatanasoiu: I think I'd prefer if you exhaustively mentioned it for admins as well and explained that only superadmins are allowed to add org info in the CSV. This should be visible to superadmins as well. 

We would then have a clear written down sentence about it, since they can't be digging in the code to see how it behaves. It's also good for us as documentation, because if we ever test it, we won't have to remember this particular exception.

h-lame: I like the idea of making it more explicit in the UI what is going to happen, but I feel like it's weird to tell someone about a feature and then say "but you can't do this".  Maybe the best solution is to let admins do this as well?  I'll ask Liz (who wrote this card originally) and Martin.

If it is for super-admins only perhaps a solution is to make the hint-text on the form better.  For the CSV we could explain to superadmins what happens if the organisation slug is wrong (something like: "If organisation is present in the CSV it should be the slug for the organisation (something something how to find this?) and if this doesn't match an organisation the user will not be invited.") and add some text to both explaining that all other fields in the CSV are ignored (something like "Only 'Name' and 'Email' fields are used from the CSV, all others will be ignored" for admins and "Only 'Name', 'Email', and 'Organisation' fields are used from the CSV, all others will be ignored" for superadmins).  For the organisation drop down we can make it clearer what it does for admins (something like: "All users are added to this organisation", and what it does for super admins (something like: "In the absence of an organisation slug in the CSV for a user, they are added to this organisation.").

elenatanasoiu: Sounds good to me

h-lame: @suzannehamilton / @elenatanasoiu - Confirmed that it is definitely for superadmins only so I have tweaked the UI help text.  Maybe I should get a content-pass on the words though (more than happy for any editorial advice from y'all though).  WDYT?

suzannehamilton: Thanks for adding the extra help text. I think that will be useful. 🙂 

I still wonder if it's confusing to just ignore the organisation fields for admin users. The scenario I'm imagining is that someone gives an admin user a CSV to use as a template. If the template contains organisations, the admin user will think they'll be able to set the organisation themselves and be confused when it doesn't work. But that might not be realistic, and I don't feel very strongly about it so I think this is fine to merge! 👍

PR #650, opened 2018-01-12
Fix dep warnings

h-lame:


Gemfile.lock
```diff
@@ -114,10 +114,10 @@ GEM
       railties (>= 4.2)
     erubi (1.7.0)
     execjs (2.7.0)
-    factory_girl (4.9.0)
+    factory_bot (4.8.2)
```
cbaines: The version going down is a bit odd to me...

h-lame: Yeah, I was surprised too, but as best I can tell 4.8.1 was the last feature release of factory_girl and 4.8.2 is the first real release of factory_bot.  4.9.0 of factory_girl is the one with the dep warnings to move from factory_girl to factory_bot, so there's no 4.9.0 of factory_bot (yet anyway, I guess if there ever is it'll be a new feature release).

PR #660, opened 2018-01-23
Deploy with the release_... tag, not release branch

cbaines:
Switch to deploying release_... instead of release, as this means it's
possible to more easily work out what was released. The Release app
also depends on the tag being used, rather than the branch for
displaying releases against commits.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,7 +80,7 @@ node {
       }
 
       stage("Deploy to integration") {
-        govuk.deployIntegration(REPOSITORY, env.BRANCH_NAME, 'release', 'deploy')
+        govuk.deployIntegration(REPOSITORY, env.BRANCH_NAME, "release_${env.BUILD_NUMBER}", 'deploy')
       }
     }
 


tijmenb:
Could you try replacing the Jenkinsfile with `buildProject`? I have a feeling it might just work after the complexity with postgres was removed (https://github.com/alphagov/signon/commit/8efa6b4142ab1da6976ac8bbc8e005c671c9a4d5#diff-58231b16fdee45a03a4ee3cf94a9f2c3). `buildProject` does the release tag deploy too.

cbaines:
> Could you try replacing the Jenkinsfile with buildProject? I have a feeling it might just work after the complexity with postgres was removed (8efa6b4#diff-58231b16fdee45a03a4ee3cf94a9f2c3). buildProject does the release tag deploy too.

Sure, I'll add it to my TODO list :)

tijmenb:
@cbaines never mind, there's still some weirdness going on preventing the use of `buildProject()` (https://github.com/alphagov/signon/pull/662).

cbaines:
> @cbaines never mind, there's still some weirdness going on preventing the use of buildProject() (#662).

Hmm, yeah, looks like the asset precompilation fails, and the current Jenkinsfile doesn't precompile the assets and this doesn't look to currently be configurable with the buildProject function. I'll merge this for now, but I'll still have a look at this later.

PR #672, opened 2018-01-31
Ensure all users have signin permission for Asset Manager

floehopper:
This database migration is heavily based on [this one][1] which did the same thing for the Support and Content Preview apps.

The migration does two things:

1. Makes `signin` a default permission for the Asset Manager app. This means that new users will automatically be granted this permission.

2. Queues up jobs to grant the permission to all existing users.

We need this because we're using SSO in Asset Manager to protect draft assets. In the same way that all users are given `signin` permission for the Content Preview app in order to be able to view draft content, all users will need `signin` permission in order to be able to view draft assets.

Ideally we'd create a new Assets Preview app in Signon for this permission to mirror the Content Preview app. However, until we split Asset Manager into frontend & backend apps (or at least two instances of the same app with different configs), we're going to need use the existing Asset Manager `Doorkeeper::Application`.  This is because the `gds-sso` gem only supports authenticating against a single Signon app and the Asset Manager app is already using `gds-sso` to protect its API.

We're OK with this, because the API routes are not publicly accessible, even via the new `draft assets` host. However, it might be worth considering changing Asset Manager so it checks that the authenticated user for API access matches ones of the Signon API users for one of the publishing apps.

### Questions

1. As far as I can tell the `BulkGrantPermissionSet` only ever *adds* permissions, i.e. the fact that I've only listed the `signin` permission for the Asset Manager app shouldn't mean that the same permissions for Support & Content Preview apps are removed. But it would be good if someone could confirm that.

2. It looks as if the [`users: grant_content_preview_access Rake task`][5] scopes the users to update using [`web_users` & `not_suspended`][6]. This Rake task looks considerably older than the migration I based this PR on, but it would be good to be sure we don't need to do this.

3. I've moved the application lookup in the `down` method of the migration out of the `say_with_time` block to match the `up` method. It would be good to be sure that this isn't going to cause any problems.

4. Although not directly related to this PR, I notice [`script/prepare_training_environment`][2] hard-codes the [`home_uri`][3] and the [`redirect_uri`][4] for the Asset Manager app. We will be changing the bottom-level domain for these URIs from `asset-manager` to `draft-assets` soon, but I believe the changes to integration, staging & production will be made manually. I'm unsure whether/when I should update the values in this script.

**Note that this PR should not be merged until we're happy that the Asset Manager authentication for draft assets is working correctly**

[1]:
https://github.com/alphagov/signon/blob/c4b928af9ac85069362f7e332f0285fae1086467/db/migrate/20171107153427_add_default_permissions_and_bulk_grant_them_to_all_users.rb
[2]: https://github.com/alphagov/signon/blob/c4b928af9ac85069362f7e332f0285fae1086467/script/prepare_training_environment
[3]: https://github.com/alphagov/signon/blob/c4b928af9ac85069362f7e332f0285fae1086467/script/prepare_training_environment#L25
[4]: https://github.com/alphagov/signon/blob/c4b928af9ac85069362f7e332f0285fae1086467/script/prepare_training_environment#L24
[5]: https://github.com/alphagov/signon/blob/c4b928af9ac85069362f7e332f0285fae1086467/lib/tasks/users.rake#L73-L90
[6]: https://github.com/alphagov/signon/blob/c4b928af9ac85069362f7e332f0285fae1086467/lib/tasks/users.rake#L76

### DIFF
--- a/db/migrate/20180131120100_set_signin_as_default_permission_for_asset_manager_and_grant_to_all_users.rb
+++ b/db/migrate/20180131120100_set_signin_as_default_permission_for_asset_manager_and_grant_to_all_users.rb
@@ -1,0 +1,31 @@
+require 'enhancements/application'
+
+class SetSigninAsDefaultPermissionForAssetManagerAndGrantToAllUsers < ActiveRecord::Migration[5.1]
+  def up
+    asset_manager_app = Doorkeeper::Application.find_by!(name: 'Asset Manager')
+
+    say_with_time 'Marking signin on Asset Manager as default permission' do
+      asset_manager_app.signin_permission.update_attributes(default: true)
+    end
+
+    say_with_time 'Enqueuing bulk grant permissions job as a super admin to make sure all existing users have the default permissions' do
+      superadmin = User.with_status('active').where(role: 'superadmin').first
+      bulk_grant = BulkGrantPermissionSet.create!(
+        user: superadmin,
+        supported_permission_ids: [asset_manager_app.signin_permission.id]
+      )
+      bulk_grant.enqueue
+    end
+  end
+
+  def down
+    # We can't undo the bulk grant because we don't know who had these
+    # permissions explicitly before we bulk granted them
+
+    asset_manager_app = Doorkeeper::Application.find_by!(name: 'Asset Manager')
+
+    say_with_time 'Removing default permission status from signin on Asset Manager' do
+      asset_manager_app.signin_permission.update_attributes(default: false)
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180111145345) do
+ActiveRecord::Schema.define(version: 20180131120100) do
 
   create_table "batch_invitation_application_permissions", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
     t.integer "batch_invitation_id", null: false


floehopper:
I tested this migration in my development environment albeit with a very small set of custom data, not data replicated from production.

### Asset Manager permissions

#### Before migration

asset-manager-permissions-before

#### After migration

asset-manager-permissions-after

### Permissions on existing user

#### Before migration

user-permissions-before

#### After migration

user-permissions-after

### Permissions on new user

#### After migration

new-user-permissions-after

floehopper:
@h-lame Thanks! Would you have time to answer the questions at the bottom of the PR description as best you can?

h-lame:
@floehopper sure:

1. correct - [this test](https://github.com/alphagov/signon/blob/master/test/models/bulk_grant_permission_set_test.rb#L96-L105) cover what you're worried about
2. this rake task should probably be removed, content-preview is now a default permission so we don't need it
3. seems fine
4. I don't know anything about this specifically.  However, I'm not sure there is a training environment anymore (if environment here refers to a deployment environment like integration, staging, etc).  To the best of my knowledge it was a firebreak project that never got finished.

floehopper:
@h-lame Thanks - that's really helpful 😄

floehopper:
@chrisroos [has pointed out][1] that the `draft-assets` host is now working as expected in staging & production and so it's now OK to merge this. So I'm going to remove the "DO NOT MERGE" prefix.

[1]: https://github.com/alphagov/asset-manager/issues/434#issuecomment-363507198

floehopper:
Rebasing against `master` and force-pushing in preparation for merging.

floehopper:
This was deployed to integration at ~11:06 today, 12 Feb. Unfortunately things didn't go as planned. As far as I can tell, the `BulkGrantPermissionSet` instance was created in the database and processing of the corresponding `BulkGrantPermissionSetJob` was started by Sidekiq. However, a couple of seconds later the Sidekiq workers were restarted as part of the deployment. This seems to have killed the job without putting it back on the queue.

```
irb> set = BulkGrantPermissionSet.last
=> #<BulkGrantPermissionSet id: 2, user_id: 109, outcome: nil, processed_users: 346, total_users: 9549, created_at: "2018-02-12 11:06:04", updated_at: "2018-02-12 11:06:04">
```

Thus only 346 of 9549 users had been processed.

Unfortunately I didn't notice that this problem had occurred and asked @dwhenry to deploy Signon to staging at ~12:27 also today, 12 Feb. So the same thing happened again.

```
irb> set = BulkGrantPermissionSet.last
=> #<BulkGrantPermissionSet id: 2, user_id: 200, outcome: nil, processed_users: 271, total_users: 14187, created_at: "2018-02-12 12:27:40", updated_at: "2018-02-12 12:27:40">
```

Thus 271 of 14187 users had been processed.

Having diagnosed the problem with help from @h-lame, I fixed it in integration by running the following commands (i.e. re-running the whole thing synchronously):

```
irb> set = BulkGrantPermissionSet.find(2)
=> #<BulkGrantPermissionSet id: 2, user_id: 109, outcome: nil, processed_users: 346, total_users: 9549, created_at: "2018-02-12 11:06:04", updated_at: "2018-02-12 11:06:04">
irb> set.perform
```

It took ~3 mins to run. After it had finished, the stats were as follows:

```
irb> set = BulkGrantPermissionSet.find(2)
=> #<BulkGrantPermissionSet id: 2, user_id: 109, outcome: "success", processed_users: 9895, total_users: 9549, created_at: "2018-02-12 11:06:04", updated_at: "2018-02-12 11:06:04">
```

Note that `processed_users` was 346 more than `total_users` as expected.

@dwhenry then ran the following commands on staging to fix the problem (i.e. re-run the whole thing asychronously):

```
irb> set = BulkGrantPermissionSet.find(2)
=> #<BulkGrantPermissionSet id: 2, user_id: 200, outcome: nil, processed_users: 271, total_users: 14187, created_at: "2018-02-12 12:27:40", updated_at: "2018-02-12 12:27:40">
irb> set.enqueue
```

~15 mins later we saw the following output:

```
irb> set.reload
=> #<BulkGrantPermissionSet id: 2, user_id: 200, outcome: "success", processed_users: 14458, total_users: 14187, created_at: "2018-02-12 12:27:40", updated_at: "2018-02-12 12:27:40">
```

Note that `processed_users` was 271 more than `total_users` as expected.

@dwhenry then carried out the following checks for me:

### Check 1

1. Sign in at https://signon.staging.publishing.service.gov.uk
2. Click on your name at top right
3. Scroll down to see the permissions you have for Asset Manager
4. See that the "Has access?" checkbox is checked for Asset Manager

### Check 2

1. Sign in at https://signon.staging.publishing.service.gov.uk
2. Click on "Applications" link (assuming you have permission)
3. Click on the "Edit" link for Asset Manager
4. Scroll down and lick on "Edit supported permissions" link
5. See that "Is a default permission given to all new users?" is "Yes" for "signin" permission

Both of these were OK.

At ~16:33 @h-lame deployed Signon to production for me, but this time **without migrations** in order to avoid Sidekiq killing the job. He then ran the migration after the deployment:

```
16:33:11 *** [err :: backend-1.backend.publishing.service.gov.uk] Migrating to SetSigninAsDefaultPermissionForAssetManagerAndGrantToAllUsers (20180131120100)
16:33:11 *** [err :: backend-1.backend.publishing.service.gov.uk] == 20180131120100 SetSigninAsDefaultPermissionForAssetManagerAndGrantToAllUsers: migrating
16:33:11 *** [err :: backend-1.backend.publishing.service.gov.uk] -- Marking signin on Asset Manager as default permission
16:33:11 *** [err :: backend-1.backend.publishing.service.gov.uk] -> 0.2031s
16:33:11 *** [err :: backend-1.backend.publishing.service.gov.uk] -- Enqueuing bulk grant permissions job as a super admin to make sure all existing users have the default permissions
16:33:11 *** [err :: backend-1.backend.publishing.service.gov.uk] -> 0.2134s
16:33:11 *** [err :: backend-1.backend.publishing.service.gov.uk] == 20180131120100 SetSigninAsDefaultPermissionForAssetManagerAndGrantToAllUsers: migrated (0.4360s)
```

So I think we're now all sorted.

Many thanks to @dwhenry and @h-lame for all their help.

floehopper:
@dwhenry Just ran the following command on a backend box in production:

```
$ govuk_app_console signon

irb> BulkGrantPermissionSet.last
#<BulkGrantPermissionSet id: 2, user_id: 200, outcome: "success", processed_users: 14196, total_users: 14196, created_at: "2018-02-12 16:33:11", updated_at: "2018-02-12 16:33:11">
```

So it looks as if the job ran successfully in production yesterday afternoon. 🎉

PR #710, opened 2018-03-07
Disallow search engine indexing

tijmenb:
This disallows all search engines from indexing the pages by this app. There's nothing for the public to see here, and service domains shouldn't be indexed anyway:

https://www.gov.uk/service-manual/technology/get-a-domain-name#using-robotstxt-and-root-level-redirects

Thanks @dashouse and @edwardhorsford

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,5 +1,2 @@
-# See http://www.robotstxt.org/wc/norobots.html for documentation on how to use the robots.txt file
-#
-# To ban all spiders from the entire site uncomment the next two lines:
-# User-Agent: *
-# Disallow: /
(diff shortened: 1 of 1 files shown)


36degrees:
I believe this will only prevent prevents robots from *crawling* the page. It will not prevent the page from appearing in the index.

https://developers.google.com/search/reference/robots_meta_tag#combining-crawling-with-indexing--serving-directives

Which means you *should* allow the robots to crawl the page, but use the noindex meta tag / header to tell it not to index it.

tijmenb:
Thanks @36degrees! We're going to try to add a header to signon.

PR #784, opened 2018-07-19
Rake task to export a list of users created / auto-suspended since a given date

barrucadu:
I used environment variables to be consistent with the other rake tasks.

---

[Trello card](https://trello.com/c/71LDXfs0/307-export-of-users-who-have-had-accounts-suspended)

### DIFF
--- a/lib/user_suspensions_exporter.rb
+++ b/lib/user_suspensions_exporter.rb
@@ -1,0 +1,51 @@
+require 'csv'
+
+class UserSuspensionsExporter
+  attr_reader :applications, :export_dir, :suspensions_since_date, :logger
+
+  def initialize(export_dir, suspensions_since_date, logger = Rails.logger)
+    @export_dir = export_dir
+    @suspensions_since_date = suspensions_since_date
+    @logger = logger
+  end
+
+  def export_suspensions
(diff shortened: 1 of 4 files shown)


lib/user_suspensions_exporter.rb
```diff
@@ -0,0 +1,52 @@
...
+  attr_reader :applications, :export_dir, :users_since_date, :suspensions_since_date, :logger
+
+  def initialize(export_dir, users_since_date, suspensions_since_date, logger = Rails.logger)
+    @export_dir = export_dir
+    @users_since_date = users_since_date
+    @suspensions_since_date = suspensions_since_date
+    @logger = logger
+  end
+
+  def export_suspensions
```
thomasleese: Since the class does only one thing it might be nice to make this a `call` method, then the `attr_reader` and `file_path` could be private.

barrucadu: Would the `file_path` being private affect the stubbing in the tests?

thomasleese: I believe Rspec is able to handle that.

barrucadu:
Ok, just waiting for content to confirm it's the sort of report they were expecting.

PR #831, opened 2018-10-01
Redesign the signin page

tijmenb:
This uses the new design for the sign in page.

- Most of the things in the layout come from content publisher
- We've had to include jQuery (from jquery-rails, a dependency of govuk_admin_template) to get the current functionality to work.
- The header & footer will be updated to reflect new designs, I've not copied over these from content publisher
- This also removes a bunch of code that will never be run, but is part of Devise - specifically `if devise_mapping.rememberable?` and the other things in the links.

## Before

screen shot 2018-10-01 at 11 43 07

## After

screen shot 2018-10-01 at 11 42 53

https://trello.com/c/05wMNlxg

### DIFF
--- a/app/views/devise/sessions/new.html.erb
+++ b/app/views/devise/sessions/new.html.erb
@@ -1,37 +1,31 @@
-<% content_for :title, "Sign in" %>
-<% content_for :thin_form, true %>
+<% content_for :title, "Sign in to GOV.UK" %>
 
-<div class="page-title">
-  <h1>Sign in</h1>
-</div>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-one-half">
+    <%= form_for(resource, as: resource_name, url: session_path(resource_name), html: { autocomplete: 'on' }) do |f| %>
+      <%= render "govuk_publishing_components/components/input", {
+        label: {
+          text: "Email"
+        },
+        name: "user[email]",
+        type: "email"
+      } %>
 
-<!--[if lte IE 7]>
-  <div class="alert alert-warning add-bottom-margin">
-    <h4>Your browser is not supported</h4>
-    <p>Pages might not work as expected. You should <a href="https://www.gov.uk/help/browsers" class="link-inherit">upgrade your browser</a> – you might need to ask your IT administrator to do this for you.</p>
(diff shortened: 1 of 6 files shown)


soniaturcotte:
What do we think of `password` instead of `passphrase` ?

app/views/devise/sessions/new.html.erb
```diff
@@ -1,37 +1,31 @@
...
-  <div class="alert alert-warning add-bottom-margin">
-    <h4>Your browser is not supported</h4>
-    <p>Pages might not work as expected. You should <a href="https://www.gov.uk/help/browsers" class="link-inherit">upgrade your browser</a> – you might need to ask your IT administrator to do this for you.</p>
-  </div>
-<![endif]-->
-<%= form_for(resource, as: resource_name, url: session_path(resource_name), html: { class: 'panel panel-default', autocomplete: 'on' }) do |f| %>
-  <div class="panel-body">
-    <div class="form-group">
-      <%= f.label :email %>
-      <%= f.email_field :email, class: 'form-control input-lg', tabindex: 1, autofocus: 'autofocus' %>
```
alex-ju: A reminder for us to update the input component to allow `tabindex` and `autofocus`. I'll try and do it later today.

app/views/devise/sessions/new.html.erb
```diff
@@ -1,37 +1,31 @@
-<% content_for :title, "Sign in" %>
-<% content_for :thin_form, true %>
+<% content_for :title, "Sign in to GOV.UK" %>
 
-<div class="page-title">
-  <h1>Sign in</h1>
-</div>
-
-<!--[if lte IE 7]>
```
alex-ju: Can we still keep this alert wrapped in a conditional IE7 tag?

tijmenb:
@soniaturcotte yeah, good idea to rename. I can do that in a separate PR to make sure we've renamed it everywhere consistently.

app/views/devise/sessions/new.html.erb
tijmenb: IRL conversation: let's move this up to the layout or components gem.

PR #848, opened 2018-10-03
Remove unused partial

tijmenb:
And associated translations.

https://trello.com/c/05wMNlxg


tijmenb:
Turns out this is still used!

PR #855, opened 2018-10-04
Update email and password

alex-ju:
This PR updates the email and password page to use the new admin layout:
- updates the form_errors to be used with both layouts (keeps `govuk-error-summary` and `alert alert-danger)
- does validation on blur (which means we don't need jquery-debounce anymore)
- removes the unnecessary bits from `password-strength-indicator.js` which were updating the UI and does inline validation using form field error message
- removes the password-guidance.scss and keeps two declarations in `new_admin_layout.scss` which can be removed when the JavaScript is refactored

This PR doesn't split the edit email and edit password in two separate pages; this work can be part of separate PR.

[Trello card](https://trello.com/c/05wMNlxg)

### Before
edit your account _ gov uk signon

### After
edit your account - gov uk


tijmenb:
I've added a commit to use the same layout on the page that the user sees after clicking the password reset link:

screen shot 2018-10-04 at 16 37 23

app/assets/javascripts/password-strength-indicator.js
```diff
@@ -17,7 +16,7 @@
             passwordConfirmation = $(options["password_confirmation_field"]).val();
         instance.updateIndicator(password, passwordConfirmation, options);
       };
-      $(password_field).debounce("keyup", update, 50);
+      $(password_field).on("blur", update);
```
tijmenb: I think we might want to keep this - I had trouble resetting my password. I first entered a weak password and I tried to update it, but the error message persisted no matter my changes. Eventually I got something working, but it seemed quite random why it accepted the password. Only after seeing this change I realised that the validation happens on blur.

PR #858, opened 2018-10-05
Add missing classes to links

tijmenb:
@alex-ju

### DIFF
--- a/app/views/root/index.html.erb
+++ b/app/views/root/index.html.erb
@@ -24,13 +24,13 @@
   <div class="govuk-grid-column-one-third">
     <ul class="govuk-list app-dashboard-side">
       <li>
-        <%= link_to "Change your email or password", edit_email_or_password_user_path(current_user) %>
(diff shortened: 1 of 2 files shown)


app/views/root/index.html.erb
```diff
@@ -24,13 +24,13 @@
   <div class="govuk-grid-column-one-third">
     <ul class="govuk-list app-dashboard-side">
       <li>
-        <%= link_to "Change your email or password", edit_email_or_password_user_path(current_user) %>
+        <%= link_to "Change your email or password", edit_email_or_password_user_path(current_user), class: "govuk-link" %>
       </li>
 
       <% if current_user.has_2sv? %>
-        <li><%= link_to "Change your 2-step verification phone", two_step_verification_path, class: "js-track" %></li>
```
alex-ju: We don't use these `js-track` selectors anywhere?

tijmenb: No, I think they were used during the initial user research.

PR #863, opened 2018-10-12
Encourage people to set up 2-step verification

tijmenb:
This updates the 2-step verification prompt page. This page is seen by user when they've been required to set up 2-step verification.

## Before

screen shot 2018-10-12 at 12 05 55

## After

screen shot 2018-10-12 at 11 58 52

Also adds a permanent warning to the dashboard if 2-step verification hasn't been set up:

screen shot 2018-10-12 at 12 03 50

### DIFF
--- a/app/assets/stylesheets/new_admin_layout.scss
+++ b/app/assets/stylesheets/new_admin_layout.scss
@@ -51,6 +51,10 @@
   }
 }
 
+a.gem-c-button.govuk-button {
+  color: #fff;
+}
+
 // TODO: replace IDs with `js-` prefixed classes as JavaScript selectors
 .not-strong-enough #password-guidance,
 .not-strong-enough #password-entropy,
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/new_admin_layout.scss
```diff
@@ -51,6 +51,10 @@
   }
 }
 
+a.gem-c-button.govuk-button {
```
alex-ju: Why do we need this?

tijmenb: Otherwise the button text color is the link color. What's a good fix for that?

alex-ju: Raised this PR https://github.com/alphagov/govuk_publishing_components/pull/582 to fix the issue in govuk_publishing_components. Once that's release I'll add a commit to this branch to remove this line and update the the new version of govuk_publishing_components.

PR #889, opened 2018-11-05
Fix user filters and CSV export

deborahchua:
Trello card: https://trello.com/c/9WFo2NqK/589-when-exporting-users-from-signon-as-csv-filtering-is-not-actually-applied-to-the-results-set


test/controllers/users_controller_test.rb
```diff
@@ -290,9 +290,21 @@ def change_user_password(user_factory, new_password)
...
+        should "export filtered users by all fields" do
+          organisation = create(:organisation, name: "Cookies Of Other Lands", abbreviation: 'COOL')
+
+          create(:suspended_user, name: "Special cookie", email: "specialcookie@email.com", role: 'normal', organisation: organisation)
+          create(:suspended_user, name: "Normal cookie", email: "normalcookie@email.com", role: 'normal', organisation: organisation)
+
+          get :index, params: { role: 'normal', status: 'suspended', organisation: organisation, two_step_status: false, filter: 'special', format: :csv }
+
           lines = @response.body.lines
           assert_equal(2, lines.length)
```
steventux: When I first read this I wondered if any users were being filtered out, it's not that clear the signed in user is omitted from the export.
Perhaps before this assertion something like:
```ruby
assert_equal(3, User.count)
```
Would make this more obvious.

deborahchua: That's a good point - thanks @steventux! I have added this in now 👍

PR #903, opened 2018-11-19
Make error messages more helpful

erino:
Show who the publisher should contact if they can't access a publishing application.

https://trello.com/c/bOxIvf1L

### DIFF
--- a/app/helpers/root_helper.rb
+++ b/app/helpers/root_helper.rb
@@ -1,9 +1,13 @@
 module RootHelper
-  def application_name
-    if @application.present?
-      @application.name
+  def gds_only_application_and_non_gds_user?
+    @application.gds_only? && !current_user.belongs_to_gds?
+  end
+
+  def title
+    if @application.blank? || gds_only_application_and_non_gds_user?
+      "You don’t have permission to use this app."
     else
-      "this application"
+      "You don’t have permission to sign in to #{@application.name}."
     end
   end
 end
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -3,6 +3,8 @@
 class Organisation < ActiveRecord::Base
   include ActiveModel::ForbiddenAttributesProtection
 
+  GDS_ORG_CONTENT_ID = "af07d5a5-df63-4ddc-9383-6a666845ebe9".freeze
+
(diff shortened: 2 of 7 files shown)


app/models/organisation.rb
```diff
@@ -3,6 +3,8 @@
 class Organisation < ActiveRecord::Base
   include ActiveModel::ForbiddenAttributesProtection
 
+  GDS_CONTENT_ID = "af07d5a5-df63-4ddc-9383-6a666845ebe9".freeze
```
cbaines: Maybe `GDS_ORG_CONTENT_ID`?

cbaines: Just because I got confused for a second.

app/views/root/signin_required.html.erb
```diff
@@ -1,16 +1,24 @@
-<% content_for :title, "Sorry, you don't have permission to access " + application_name %>
+<% content_for :title, title %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+    <% if gds_only_application_and_non_gds_user? %>
+    <% else %>
```
cbaines: Is the above part of the `if` meant to be empty?

app/helpers/root_helper.rb
```diff
@@ -1,9 +1,13 @@
...
+  def gds_only_application_and_non_gds_user?
+    @application.gds_only? && !current_user.belongs_to_gds?
+  end
+
+  def title
+    if @application.blank? || gds_only_application_and_non_gds_user?
+      "You don’t have permission to use this app."
     else
-      "this application"
+      "You don’t have permission to sign in to " + @application.name + "."
```
cbaines: `in to #{@application.name}."` might be neater and more consistent .

erino: 👍

app/views/root/signin_required.html.erb
erino: Yes, fixed in 9cae5e4b1cbe9786deb7b44c7a594904900ef85a

app/models/organisation.rb
erino: 👍

PR #943, opened 2019-01-03
Add extra healthcheck for expiring API User tokens

benthorner:
https://trello.com/c/pvSF3FTe/5-rotate-gds-sso-tokens

This will help us to proactively rotate tokens by alerting us as reach
their expiry. We can't currently automate cycling of tokens, so the
check includes details of each API User and Application pair that need
attention, which is enough to find them in the UI.

### DIFF
--- a/lib/healthcheck/api_tokens.rb
+++ b/lib/healthcheck/api_tokens.rb
@@ -1,0 +1,47 @@
+module Healthcheck
+  class ApiTokens
+    WARNING_THRESHOLD = 1.year.to_i
+    CRITICAL_THRESHOLD = 1.month.to_i
+
+    QUERY = <<-SQL.freeze
+      SELECT users.name, oauth_applications.name FROM oauth_access_tokens
+      INNER JOIN users ON users.id = resource_owner_id
+      INNER JOIN oauth_applications ON oauth_applications.id = application_id
+      WHERE oauth_access_tokens.revoked_at IS NULL
+      AND users.api_user = TRUE
+      AND (oauth_access_tokens.expires_in -
+          (unix_timestamp(now()) -
+           unix_timestamp(oauth_access_tokens.created_at))) < %<threshold>s
+    SQL
+
+    def name
+      :api_tokens
+    end
+
+    def details
+      { warnings: warnings - criticals, criticals: criticals }
+    end
+
+    def status
+      return GovukHealthcheck::CRITICAL if criticals.any?
+
+      return GovukHealthcheck::WARNING if warnings.any?
+
+      GovukHealthcheck::OK
+    end
+
+  private
+
+    def warnings
+      @warnings ||= connection.execute(QUERY % { threshold: WARNING_THRESHOLD }).to_a
+    end
+
+    def criticals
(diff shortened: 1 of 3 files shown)


lib/healthcheck/api_tokens.rb
```diff
@@ -0,0 +1,47 @@
...
+    QUERY = <<-SQL.freeze
+      SELECT users.name, oauth_applications.name FROM oauth_access_tokens
+      INNER JOIN users ON users.id = resource_owner_id
+      INNER JOIN oauth_applications ON oauth_applications.id = application_id
+      WHERE oauth_access_tokens.revoked_at IS NULL
+      AND users.api_user = TRUE
+      AND (oauth_access_tokens.expires_in -
+          (unix_timestamp(now()) -
+           unix_timestamp(oauth_access_tokens.created_at))) < %<threshold>s
+    SQL
```
kevindew: Do you know if this sql is ok for both mysql and postgres? I believe (and would like to be wrong) that this app supports both databases.

lib/healthcheck/api_tokens.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def name
+      :api_tokens
+    end
+
+    def details
+      { warnings: warnings - criticals, criticals: criticals }
```
kevindew: It'd be nice if these were a bit more readable and said when the expiry was. E.g. "#{user_name} token for #{publishing_app} expires at #{datetime}" 

I don't think we'd actually need to break it up into warning and critical groups if we have the times there - what do you think?

lib/healthcheck/api_tokens.rb
```diff
@@ -0,0 +1,47 @@
+module Healthcheck
+  class ApiTokens
+    WARNING_THRESHOLD = 1.year.to_i
+    CRITICAL_THRESHOLD = 1.month.to_i
```
kevindew: I wonder if these are a bit too high - I think something that is a year away wouldn't be enough to inspire me to do something soon.

I'd suggest maybe 1 month warning and 1 week critical?

lib/healthcheck/api_tokens.rb
benthorner: @tijmenb thinks we can drop Postgres support - the tech debt card doesn't mention a particular use-case we're aiming for, aside from migrating fully to one database.

https://trello.com/c/Zw8xD510/61-signon-supports-both-postgresql-and-mysql

lib/healthcheck/api_tokens.rb
benthorner: Done :-). Rewritten a bit to avoid calculating the time twice. I also thought it would be clearer to show the number of days, instead of the timestamp.

lib/healthcheck/api_tokens.rb
benthorner: Going for 2 months, 1 month, as discussed.

lib/healthcheck/api_tokens.rb
kevindew: Great - be good to get rid of one!

benthorner:
@kevindew looks good to me...

screen shot 2019-01-07 at 09 54 06

kevindew:
Nice 👍

PR #948, opened 2019-01-04
Rewrite RSpec policy specs as MiniTest tests

benthorner:
https://trello.com/c/pvSF3FTe/5-rotate-gds-sso-tokens
^^^ card where pain was felt - which framework should I write my new test in?

Previously these specs were written in RSpec to take advantage of the
pundit matchers, which are only available for that test framework.
This adds the small amount of helper code necessary to rewrite the
specs as MiniTest tests, so that we can move to just one test framework.

The rewrite is mostly find-and-replace, with a couple of changes:

  * Combining policy specs that individually tested different users
  * Replacing disallowed_actions_* with a single test for 'create'

The latter relates to the User policy test, where we were previously
testing against the 'assign_permissions' action twice.


benthorner:
Thanks :-). Now just some finishing touches to lose RSpec entirely...

PR #955, opened 2019-01-07
Restructure tests

benthorner:
This moves some of the tests so that they match the directory structure in app/ and lib/, as well as separating test helpers from tests of helpers, which wasn't confusing at all ;-).


benthorner:
Ah, I see what you mean. Turns out the AppHelper classes don't have an 's' on the end, so they should be fairly distinct. Renaming them to XTestHelper might be confusing for some of them, as they're not all used in tests called XTest. Are you happy if I leave as-is?

PR #958, opened 2019-01-08
Tweak app for expired invites

tijmenb:
Currently we only display the strike through for accounts that are suspended. This changes that that to include accounts that haven't been activated yet (the user hasn't accepted the invitation mail) and locked accounts (too many incorrect password attempts).

Also includes some tweaks & refactors - see commits for details.


app/models/user.rb
```diff
@@ -150,24 +150,25 @@ def grant_permission(supported_permission)
...
-  #    https://github.com/alphagov/signon/commit/451b89d9
+  # Override Devise::Recoverable behavior to notify suspended users that they
+  # can't reset their password.
   def self.send_reset_password_instructions(attributes = {})
     user = User.find_by_email(attributes[:email])
     if user.present? && user.suspended?
       UserMailer.notify_reset_password_disallowed_due_to_suspension(user).deliver_later
       user
-    elsif user.present? && user.invited_to_sign_up? && !user.invitation_accepted?
+    elsif user.present? && user.invited_but_not_yet_accepted?
```
benthorner: Good find :-).

app/models/user.rb
```diff
@@ -150,24 +150,25 @@ def grant_permission(supported_permission)
     application_permissions.where(supported_permission_id: supported_permission.id).first_or_create!
   end
 
-  # override Devise::Recoverable behavior to:
-  # 1. notify suspended users that they can't reset their password, and
-  # 2. handle emails blacklisted by AWS such that we conceal whether
-  #    or not an account exists for that email. moved from:
-  #    https://github.com/alphagov/signon/commit/451b89d9
+  # Override Devise::Recoverable behavior to notify suspended users that they
```
benthorner: Should this be 'notify unusable users'?

tijmenb: Removed a bit, it's already clear from the code what it does.

tijmenb:
@benthorner thanks, updated the users with access. But the status already accounts for invited users:

https://github.com/alphagov/signon/blob/59a053f5d926b07af157af69b85cebfdfdaeaedb/app/models/user.rb#L74-L75

benthorner:
*when they're not suspended

PR #959, opened 2019-01-08
Stop password expiration

tijmenb:
We currently require people to change their password every 90 days. This isn't a best practice anymore, and contravenes the latest NCSC guidance:

https://www.ncsc.gov.uk/articles/problems-forcing-regular-password-expiry
https://www.ncsc.gov.uk/guidance/password-guidance-simplifying-your-approach

This PR removes the password expiry logic. 

It's likely still useful to know when a user has last changed their password, so I've moved the `update_password_changed` callback into the User model.

Closes https://github.com/alphagov/signon/issues/524. Supersedes an earlier attempt: https://github.com/alphagov/signon/pull/540.

https://trello.com/c/HXSZikzB


boffbowsh:
👏

PR #970, opened 2019-01-15
Use a fixed version of alphabetical_paginate

cbaines:
That doesn't show a deprecation warning. This change hasn't been
released yet but has been pushed upstream here [1].

1: https://github.com/lingz/alphabetical_paginate/pull/55

The patch changes alphabetical_paginate to work with Rails 6, where
non-attribute arguments will be disallowed.

Below is an example of a deprecation warning that shows up without
this:

```
DEPRECATION WARNING: Dangerous query method (method whose arguments
are used as raw SQL) called with non-attribute argument(s): "substr(
users.name, 1 , 1)". Non-attribute arguments will be disallowed in
Rails 6.0. This method should not be called with user-provided values,
such as request parameters or model attributes. Known-safe values can
be passed by wrapping them in Arel.sql().
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,11 @@ gem 'rails', '~> 5'
 
 gem 'activejob-retry'
 gem 'addressable', '~> 2'
-gem 'alphabetical_paginate', '~> 2'
+gem(
+  'alphabetical_paginate',
+  git: 'https://github.com/cbaines/alphabetical_paginate',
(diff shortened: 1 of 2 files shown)


cbaines:
I'm not sure if we actually want to merge this, or wait until a new upstream release, but I though I'd raise a PR for visibility anyway.

PR #975, opened 2019-01-17
Improve security for 2SV cookies

tijmenb:
At present, the 2SV cookie (which determines whether or not the user needs to provide a 2FA code) is not marked as secure or httpOnly. I think we intended to do this in https://github.com/alphagov/signon/pull/507, but I think this was missed.

I've not found a way to actually test this in code. I've tried the `cookies` hash [as described in the Rails docs](https://guides.rubyonrails.org/testing.html#the-three-hashes-of-the-apocalypse), and I've tried to access this cookie via `Capybara.current_session.driver.response.headers["Set-Cookie"]`. Any ideas welcome.
 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Secure_and_HttpOnly_cookies

https://trello.com/c/cDd3UUeg


rubenarakelyan:
With Capybara you can check "secure" but not "httponly" - I found the same with Smokey (https://github.com/alphagov/smokey/blob/master/features/step_definitions/cookie_steps.rb).

PR #1016, opened 2019-02-26
Update govuk-lint to version 3.11.0

cbaines:
And fix all the linting issues that this raises.


app/models/event_log.rb
```diff
@@ -109,19 +109,19 @@ def self.for(user)
...
 
-private
-
-  def validate_event_mappable
-    unless entry
-      errors.add(:event_id, "must have a corresponding `LogEntry` for #{event_id}")
-    end
-  end
-
   def self.convert_ip_address_to_integer(ip_address_string)
```
erino: Do you think the author of these methods expected them to be private?

cbaines: Thanks for taking a look Erin.

I guess maybe, but as pure functions, the only effect I can think of in terms of being private would be to discourage use from outside this class. Ruby seems to have a `private_class_method` thing you can use, but I don't really know what it does. ..

erino: I don't think private class methods are particularly easy in ruby, and as you say there shouldn't be any side effects

PR #1062, opened 2019-04-17
Clones Trade Tariff Admin and Trade Tariff Backend apps and user permissions

oscarwyatt:
We will clone the applications and allow users to use the new one
then change the URI on the new application to the new service.
Once this works, we can retire the old applications
Zendesk: https://govuk.zendesk.com/agent/tickets/3634856

### DIFF
--- a/db/migrate/20190417111131_clone_trade_tariff_admin.rb
+++ b/db/migrate/20190417111131_clone_trade_tariff_admin.rb
@@ -1,0 +1,48 @@
+class CloneTradeTariffAdmin < ActiveRecord::Migration[5.2]
+  def up
+    ActiveRecord::Base.transaction do
+      existing_trade_tariff_admin = Doorkeeper::Application.find(47)
+      cloned_trade_tariff_admin = Doorkeeper::Application.new
+      existing_trade_tariff_admin.attributes.each_pair do |key, value|
+        if !%w(created_at updated_at id uid).include?(key)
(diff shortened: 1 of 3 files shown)


cbaines:
> We will clone the applications and allow users to use the new one
> then change the URI on the new application to the new service.
> Once this works, we can retire the old applications

I'm not quite sure about the motivation behind this change. Is this part of a migration between two domains? And if it is, why can't the relevant settings in Signon just be changed for the existing application(s)?

cbaines:
@oscarwyatt I'd still appreciate continuing the discussion here, as I'd like to know more about what I asked previously.

Also, merging this before the conversation was even started isn't very useful. Even ignoring the questions I raised, the commit message doesn't meet the GOV.UK styleguide https://github.com/alphagov/styleguides/blob/master/git.md#structure .

oscarwyatt:
@cbaines We're cloning the apps and then moving them as (as far as I understand) there needs to be a bit of a dance where the people managing the service (or we) need to change the secret on the new app, somehow communicate the new secret (haven't figured that out yet) and spin up their new instances with that new secret, before dropping the old ones and removing those apps on signon. There has been a fair bit of back and forth on the Zendesk ticket between several departments and this approach seems to be the consensus. Thomas Leese has more of the context than I

I can't figure out a way to amend the commit message (I'm no git guru) but I'll keep a closer eye on the style guide in the future

PR #1063, opened 2019-04-17
Fix issues with the clone_trade_tariff_admin migration

oscarwyatt:
This fixes some bugs in commit c81f453

The previous commit works in a rails console but not when deploying.
This is due to it not having a proper definition of
Doorkeeper::Application and it having the method
Doorkeeper::Application.supported_permissions. By adding the
definition and using a where command instead, this fixes the issue.
Zendesk: https://govuk.zendesk.com/agent/tickets/3634856

### DIFF
--- a/db/migrate/20190417111131_clone_trade_tariff_admin.rb
+++ b/db/migrate/20190417111131_clone_trade_tariff_admin.rb
@@ -1,8 +1,12 @@
 class CloneTradeTariffAdmin < ActiveRecord::Migration[5.2]
+  class SupportedPermission < ActiveRecord::Base
+    belongs_to :application, class_name: 'Doorkeeper::Application'
+  end
+
   def up
     ActiveRecord::Base.transaction do
-      existing_trade_tariff_admin = Doorkeeper::Application.find(47)
-      cloned_trade_tariff_admin = Doorkeeper::Application.new
+      existing_trade_tariff_admin = ::Doorkeeper::Application.find_by_name("Trade Tariff Admin (PaaS)")
+      cloned_trade_tariff_admin = ::Doorkeeper::Application.new
       existing_trade_tariff_admin.attributes.each_pair do |key, value|
         if !%w(created_at updated_at id uid).include?(key)
           cloned_trade_tariff_admin[key] = value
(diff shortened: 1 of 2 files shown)


cbaines:
What does this Pull Request relate to? It seems to be an exact copy of a previous one.

oscarwyatt:
@cbaines I've made some changes since, it fixes some bugs in #1062 which hadn't been a problem when I was writing/testing the original in a rails console but cause problems when deploying. Let me know if you'd like me to revert the original and squash everything into one

cbaines:
> @cbaines I've made some changes since, it fixes some bugs in #1062 which hadn't been a problem when I was writing/testing the original in a rails console but cause problems when deploying. Let me know if you'd like me to revert the original and squash everything into one

While the best that can now be done with #1062 is to update the Pull Request with any information that would have ideally gone in the commit message, lets perhaps try here to write a good, style guide compliant commit message prior to merging the commit to master.

In particular for these changes, the commit message says:
```
This fixes some bugs in commit c81f45305e760dfdb79b5cec7ac4d2901d01f5db

Clones Trade Tariff Admin and Trade Tariff Backend apps and user permissions
We will clone the applications and allow users to use the new one
then change the URI on the new application to the new service.
Once this works, we can retire the old applications
Zendesk: https://govuk.zendesk.com/agent/tickets/3634856
```

The first line is quite a bit over the 50 character limit set out in the styleguide, and also exceeds the 68 characters at which my editor starts complaining. The styleguide also talks about putting the commit message in the present tense, as this makes a bit more sense when you look at the commit individually.

Something like `Fix issues with the clone_trade_tariff_admin migration` is better I think, a little over the limit, and you could always cut out `the` if you want to stick more closely to 50 characters.

As for the rest of the commit message, it would be useful if that described the issues and fixes in more detail. Doesn't have to be much, but at least remove the body carried over from the previous commit.

oscarwyatt:
@cbaines Commit amended 😃

cbaines:
> @cbaines Commit amended smiley

Thanks, it looks good, however the first line is still the same as before.

oscarwyatt:
@cbaines Apologies, missed that

PR #1075, opened 2019-04-30
Stream event logs to Splunk

rubenarakelyan:
This commit adds functionality to stream event logs to Splunk using Sidekiq background jobs. These logs will be used for analysis of Signon usage patterns and anomalies.


app/models/event_log.rb
```diff
@@ -76,6 +76,29 @@ def ip_address_string
...
+
+    event = {
+      id: self.id,
+      uid: self.uid,
+      created_at: self.created_at,
+      initiator_name: self.initiator&.name,
+      application_name: self.application&.name,
+      trailing_message: self.trailing_message,
+      event: self.event,
+      ip_address: (self.ip_address_string unless self.ip_address.nil?),
```
erino: What about `self.ip_address_string if self.ip_address.present?`?

rubenarakelyan: Yup, done

PR #1081, opened 2019-05-03
Set event log retention policy

rubenarakelyan:
This commit enforces a retention policy of 2 years for signon event logs. This is achieved with a scheduled job that deletes all log entries older than 2 years every day.


lib/tasks/event_log.rake
```diff
@@ -0,0 +1,18 @@
...
+
+    path = "#{ENV['GOVUK_APP_ROOT'] || Rails.root}/reports"
+    EventLogPresenter.new(min_id, max_id).write_csv(path)
+
+    puts "Report successfully generated inside #{path}"
+  end
+
+  desc "Delete all events in the event log older than 2 years"
+  task delete_logs_older_than_two_years: :environment do
+    delete_count = EventLog.where("created_at < ?", 2.years.ago).delete_all
```
benlovell: That will delete everything up to two years ago.

timblair: Yup, that's what we're trying to achieve. From a data retention perspective, we're only keeping audit events for the last two years.

benlovell: What am I missing? The query says less than 2 years old.

rubenarakelyan: It's saying anything before 2 years ago...

timblair: "Where the the created at timestamp is less than (further back in time than) the date two years ago."

benlovell: Time for me to go home. 🤦‍♂

PR #1082, opened 2019-05-03
Add nta font

theseanything:
This fixes a bug where the font isn't loading. Font loading is disabled since govuk-frontend v2.9.0 to ensure font file is not required twice for projects that rely on govuk_template. The typeface comes disabled from govuk_publishing_component and needs to be enabled in repos that don't rely govuk_template.


theseanything:
[(1) Fix fonts in Signon](https://trello.com/c/sWn0hRvt/1420-1-fix-fonts-in-signon)

alex-ju:
@theseanything I was preparing to push a PR for this. Thanks!

PR #1085, opened 2019-05-07
Encapsulate Splunk event payload

rubenarakelyan:
This commit wraps the Splunk event payload inside an `event` block as required by the Splunk HTTP event collector.


app/models/event_log.rb
```diff
@@ -80,15 +80,17 @@ def send_to_splunk(*)
     return unless ENV['SPLUNK_EVENT_LOG_ENDPOINT_URL'] && ENV['SPLUNK_EVENT_LOG_ENDPOINT_HEC_TOKEN']
 
     event = {
```
thomasleese: I wonder if it would be clearer to have it below like this instead:

```rb
request.body = { event: event }.to_json
```

Just because then we don't need the extra `event` key.

rubenarakelyan: Yup looks neater, done 👍

PR #1193, opened 2019-09-03
Improve splunk logging of unsuccessful login attempts

barrucadu:
**Record user agent string on unsuccessful login:** done by adding a `user_agent_string` field to the `event_logs` table.  I'm not sure if this approach is better or worse than creating `UserAgent` models, I went for this because I figured it would be pretty straightforward to prune the event log if it grows too big.  If there are too many `UserAgent` values, then both `event_logs` and `user_agents` need pruning.

**Record email address when no such account exists:** done by making the `uid` field of `event_logs` optional (though it's required for all pre-existing event types), adding a `user_email_string` field, and adding a new event type.

The new event is type 38 and passes the email address in the `user` field.

---

https://govuk.zendesk.com/agent/tickets/3783119
https://govuk.zendesk.com/agent/tickets/3788622

### DIFF
--- a/app/models/event_log.rb
+++ b/app/models/event_log.rb
@@ -42,6 +42,7 @@ class EventLog < ActiveRecord::Base
     PERMISSIONS_ADDED = LogEntry.new(id: 35, description: "Permissions added", require_uid: true, require_initiator: true),
     PERMISSIONS_REMOVED = LogEntry.new(id: 36, description: "Permissions removed", require_uid: true, require_initiator: true),
     ACCOUNT_INVITED = LogEntry.new(id: 37, description: "Account was invited", require_uid: true, require_initiator: true),
+    NO_SUCH_ACCOUNT_LOGIN = LogEntry.new(id: 38, description: "Attempted login to nonexistent account"),
 
     # We no longer expire passwords, but we keep this event for history purposes
     PASSWORD_EXPIRED = LogEntry.new(id: 6, description: "Password expired"),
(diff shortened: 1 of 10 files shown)


app/models/event_log.rb
```diff
@@ -88,7 +88,7 @@ def send_to_splunk(*)
       src_ip: (self.ip_address_string if self.ip_address.present?),
       action: self.event,
       result: self.trailing_message,
-      http_user_agent: self.user_agent&.user_agent_string
+      http_user_agent: self.user_agent&.user_agent_string || self.user_agent_string,
```
sihugh: Is this the same thing again or did I miss something? `user_agent_string` appears to be delegated to `user_agent`:  https://github.com/alphagov/signon/blob/5b66fcbf425e35035bbafeee4fafdc3243b235d7/app/models/event_log.rb#L65

barrucadu: Whoops, that would have been a bug.  Good catch

app/models/event_log.rb
```diff
@@ -6,53 +6,55 @@ class EventLog < ActiveRecord::Base
...
-    ACCESS_TOKEN_GENERATED = LogEntry.new(id: 29, description: "Access token generated", require_application: true, require_initiator: true),
-    ACCESS_TOKEN_REVOKED = LogEntry.new(id: 30, description: "Access token revoked", require_application: true, require_initiator: true),
-    PASSWORD_RESET_LOADED_BUT_TOKEN_EXPIRED = LogEntry.new(id: 31, description: "Password reset page loaded but the token has expired"),
-    SUCCESSFUL_PASSWORD_RESET = LogEntry.new(id: 32, description: "Password reset successfully"),
-    ROLE_CHANGED = LogEntry.new(id: 33, description: "Role changed", require_initiator: true),
-    ACCOUNT_UPDATED = LogEntry.new(id: 34, description: "Account updated", require_initiator: true),
-    PERMISSIONS_ADDED = LogEntry.new(id: 35, description: "Permissions added", require_initiator: true),
-    PERMISSIONS_REMOVED = LogEntry.new(id: 36, description: "Permissions removed", require_initiator: true),
-    ACCOUNT_INVITED = LogEntry.new(id: 37, description: "Account was invited", require_initiator: true),
+    ACCOUNT_LOCKED = LogEntry.new(id: 1, description: "Password verification failed too many times, account locked for #{LOCKED_DURATION}", require_uid: true),
```
erino: I wonder if the default for `require_uid ` should be `true` as nearly all of these `LogEntry`s have `require_uid: true`.

barrucadu: I thought about that, but it would be inconsistent with the others, which could be confusing

erino: Fair enough

PR #1225, opened 2019-09-27
Use source version of alphabetical_paginate

sihugh:
[Chris' PR](https://github.com/lingz/alphabetical_paginate/pull/55) was merged into master (and is the only change since v2.3.3 which was the previous version we used).  

We want to use the changes in his PR to ensure that we're ready for Rails 6.


sihugh:
I've raised a PR on the source repo to [bump the gem version](https://github.com/lingz/alphabetical_paginate/pull/56).  If this gets merged then we can revert to using plain old rubygems

PR #1237, opened 2019-10-04
Update to govuk frontend v3

chao-xian:
Signon still uses `govuk_admin_template`, but this bump still brings in some new things from GOV.UK Frontend 3.

## Screenshot

signon dev gov uk_ (1)


chao-xian:
https://trello.com/c/m3F4UCQE

PR #1242, opened 2019-10-07
Add links to the support app to relevant signon emails

Tetrino:
To make sure we redirect to the support app correctly from emails, we now ensure there are relevant links in emails that mention the key phrases "support form" or "support ticket."

In the case of text documents, the relevant link is included in brackets after the phrase.

Trello card: https://trello.com/c/B14NqZXp/1048-add-links-to-the-support-app-to-relevant-signon-emails

### DIFF
--- a/app/views/user_mailer/notify_reset_password_disallowed_due_to_suspension.text.erb
+++ b/app/views/user_mailer/notify_reset_password_disallowed_due_to_suspension.text.erb
@@ -5,4 +5,4 @@ Your <%= t('department.name') %> Signon <%= account_name %>, for <%= @user.email
 <%= account_name.humanize %> suspensions do not suspend production accounts.
 <% end %>
 
-If you believe your account has been suspended in error, please contact a managing <%= t('department.name') %> editor in your organisation (or your parent organisation). They can use the support form to get help from the <%= t('department.name') %> team.
+If you believe your account has been suspended in error, please contact a managing <%= t('department.name') %> editor in your organisation (or your parent organisation). They can use the support form (https://www.gov.uk/support/internal) to get help from the <%= t('department.name') %> team.
--- a/app/views/user_mailer/notify_reset_password_disallowed_due_to_suspension.html.erb
+++ b/app/views/user_mailer/notify_reset_password_disallowed_due_to_suspension.html.erb
@@ -4,4 +4,4 @@
 <p><%= account_name.humanize %> suspensions do not suspend production accounts.</p>
 <% end %>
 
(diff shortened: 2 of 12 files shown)


Tetrino:
@sihugh a bit of a simple approach for the moment while I sort out dynamic linking, just making sure I have everything.

Tetrino:
> I was wondering if it would be best to extract the support URL to the [same config file](https://github.com/alphagov/signon/blob/master/config/locales/department_specific.en.yml) as the other GOV.UK specific bits (like `department.name`) etc.

Ahah, this is what I went looking for - couldn't find it so wen't directly.

> Also, could we add a test? I often forget which way round the `link_to` params go, so it's be great to verify that everything looks right. I don't think there are many tests around the email bodies, so this would be a good thing to add.

I'll throw one in.

app/views/user_mailer/notify_reset_password_disallowed_due_to_suspension.text.erb
```diff
@@ -5,4 +5,4 @@ Your <%= t('department.name') %> Signon <%= account_name %>, for <%= @user.email
 <%= account_name.humanize %> suspensions do not suspend production accounts.
 <% end %>
 
-If you believe your account has been suspended in error, please contact a managing <%= t('department.name') %> editor in your organisation (or your parent organisation). They can use the support form to get help from the <%= t('department.name') %> team.
+If you believe your account has been suspended in error, please contact a managing <%= t('department.name') %> editor in your organisation (or your parent organisation). They can use the support form (<%= t('support.url') %>) to get help from the <%= t('department.name') %> team.
```
bilbof: Should this be wrapped in an `<a>`?

Tetrino: Is that how text is rendered? I couldn't find any clear definition for it. Other links in text.erb don't have the <a> e.g. `<%= link_to t('department.name'), t('department.url') %>` in signon/app/views/user_mailer/notify_reset_password_disallowed_due_to_unaccepted_invitation.text.erb

Tetrino:
@bilbof @sihugh Added some extra tests - ideally I wanted to iterate through and do a deeper pass of the email content, but that proved unreliable (hence the delay, sorry about that). In the mean time I've added tests matching the other patterns for the uncovered scenarios and extended existing tests to include a URL parse check.

Tetrino:
> 🎉

I wouldn't yet, I've got a second test incoming!

Tetrino:
@sihugh @bilbof One final commit adding a new file content test. We can extend this test when needed, but for now it's built with one target in mind: Ensuring link_to isn't malformed.

I've put some comments in the file itself to make it clearer because frankly, regexp is messy at the best of times.

PR #1457, opened 2020-07-07
Specify the Rails version for each migration

issyl0:
- We're trying to bring up Signon (`deployed-to-production` branch) on our Replatforming test environment. But the deploy job is failing due to:

```
16:29:59 *** [err :: ip-10-200-4-67.eu-west-1.compute.internal] StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:
16:29:59 *** [err :: ip-10-200-4-67.eu-west-1.compute.internal]
16:29:59 *** [err :: ip-10-200-4-67.eu-west-1.compute.internal] class DeviseCreateUsers < ActiveRecord::Migration[4.2]
```

- We have to specify Rails versions for all of the migrations in this repo. It says "specify the Rails release the migration was written for", but we have migrations going back to 2012, when Rails 6 didn't exist.  I've set it to `[6.0]` as I don't really want to have to comb through and figure out which exact Rails release was available in 2012, 2014, etc.
- This seems quite noisy. 😞 :nauseated_face: Presumably it doesn't occur in other environments because the data sync means the databases are already populated.

### DIFF
--- a/db/migrate/20120402120458_devise_create_users.rb
+++ b/db/migrate/20120402120458_devise_create_users.rb
@@ -1,4 +1,4 @@
-class DeviseCreateUsers < ActiveRecord::Migration
+class DeviseCreateUsers < ActiveRecord::Migration[6.0]
   def change
     create_table(:users) do |t|
       ## Database authenticatable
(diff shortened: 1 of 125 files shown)


issyl0:
We're not going to take this approach.

PR #1458, opened 2020-07-07
git rm -r db/migrate/

issyl0:
- We're trying to spin up Signon in the `test` env, `pink` stack, as part of having a working base for Replatforming.
- We were facing problems with Signon not deploying because Rails refused to apply migrations, because Rails behaviour has changed between 2012 and now (surprise) and old migrations no longer apply.
- The solution, it seems, is to ignore the migrations by deleting them all, as we're not going to edit every single migration for workingness eight years on. The [Rails guides](https://edgeguides.rubyonrails.org/active_record_migrations.html#old-migrations) say that the `db/schema.rb` file is the canonical source of the current state of the database (what `db:setup` refers to), not the migrations, so the `db/migrate` files can be deleted at any time (once applied).
- Worth noting that this problem does not occur in production, staging or integration, or indeed in GOV.UK dev environments, because we've replicated production data so the database exists when the apps deploy/run, fully populated.

Co-authored-by: Bill Franklin <william.franklin@digital.cabinet-office.gov.uk>


kevindew:
I'm a bit worried this may be quite surprising as you'd expect an ActiveRecord project to have migrations and it's unconventional for our ActiveRecord apps to not have any. The approach we've taken in the past is to nuke just the data migrations that are coupled to application code (😡). I'll just do a quick look to see if that's practical as that seems preferable if it's not a massive job.

kevindew:
I had a bit of a play to see what the situation is like ([branch](https://github.com/alphagov/signon/tree/fix-migrations)) and even then the schema comes out different so I don't think there's anything to salvage here as production and migrations seem to have diverged.

I'm a bit concerned that if this is something that is beyond a one-off for signon we may need to have some community discussions about when is the right time to delete migrations (it seems to be something we've successfully put off for many years) as there's a level of confusion and ambiguity about when/why a migration nuking should occur. Might be worth just checking in tomorrow with #govuk-developers before merging to see if anyone has concerns/opinions.

cbaines:
> * We're trying to spin up Signon in the `test` env, `pink` stack, as part of having a working base for Replatforming.
> 
> * We were facing problems with Signon not deploying because Rails refused to apply migrations, because Rails behaviour has changed between 2012 and now (surprise) and old migrations no longer apply.

So, I'm guessing this is occurring because the deployment process tries to run the migrations.

> * The solution, it seems, is to ignore the migrations by deleting them all, as we're not going to edit every single migration for workingness eight years on. The [Rails guides](https://edgeguides.rubyonrails.org/active_record_migrations.html#old-migrations) say that the `db/schema.rb` file is the canonical source of the current state of the database (what `db:setup` refers to), not the migrations, so the `db/migrate` files can be deleted at any time (once applied).
>
> * Worth noting that this problem does not occur in production, staging or integration, or indeed in GOV.UK dev environments, because we've replicated production data so the database exists when the apps deploy/run, fully populated.

I think the more common (what I'd call proper) approach to this is run `rake/rails db:setup`. If you're deploying the app from scratch, that's what you should do to setup a database. I'm not particularly looking forward to this trend of deleting the migrations from the Git repositories just to work around the fact that the processes for bringing up new environments don't handle the situation properly.

cbaines:
>     * Worth noting that this problem does not occur in production, staging or integration, or indeed in GOV.UK dev environments, because we've replicated production data so the database exists when the apps deploy/run, fully populated.

Ah, I forgot to say, this isn't particularly relevant. For local development, you also run `rake/rails db:setup`, and that's why this issue with the migrations doesn't arise. This functionality is written in to tools like govuk-docker and govuk-guix.

bevanloon:
Glancing through the migrations, there are instances where it looks like we are setting data as well as changing schema, Is there a risk in removing all migrations that we loose some state of data that a fresh application relies on? I realise that currently running migrations on a fresh copy breaks, so perhaps a bit of a moot point.

The other thought I had - are there any classes/code paths in for example the `lib` directory that we only ever call from a migration? If we are throwing away migrations, cleaning those potential things up as well would be a good idea, though of course more tedious.

bilbof:
I'll close this, since on further investigation we found that it's not necessary.

Rails maintains a `schema_migrations` table based on the filenames under the db/migrations directory. This lets it keeps track of which migrations have to be run. E.g.

```
10:33:46 database: signon_production
10:33:46 
10:33:46  Status   Migration ID    Migration Name
10:33:46 --------------------------------------------------
10:33:46    up     20120402120458  Devise create users
10:33:46    up     20120402150534  Make users suspendable
10:33:46    up     20120403120725  Create doorkeeper tables
10:33:46    up     20120410094033  Add name to users
10:33:46    up     20120410094527  Add uid to users
10:33:46    up     20120412131921  Remove rememberable from users
10:33:46    up     20120609094354  Add admin flag to users
10:33:46    up     20120609094747  Add indices for user
10:33:46   down    20120611131259  Create permissions
10:33:46   down    20120612135328  Devise invitable add to users
```

We had deployed a branch **with migrations removed**, set up the database, and then tried deploying the app with migrations. 

This didn’t work, since rails thought the migrations were new/down and tried to run them (as above), which isn’t what we wanted.

The fix was to:

1. deploy the release of signon without migrations
2. re-setup the db: `govuk_setenv signon bundle exec rake db:schema:load DISABLE_DATABASE_ENVIRONMENT_CHECK=1`
3. deploy with migrations (no migrations actually ran)

The result of `db:schema:load` meant that all migrations present in the db/migrations directory at the time of setting up the db are assumed to have already run (marked as `up`), which prevents them from being run again:

```
10:40:02  Status   Migration ID    Migration Name
10:40:02 --------------------------------------------------
10:40:02    up     20120402120458  Devise create users
10:40:02    up     20120402150534  Make users suspendable
10:40:02    up     20120403120725  Create doorkeeper tables
10:40:02    up     20120410094033  Add name to users
10:40:02    up     20120410094527  Add uid to users
10:40:02    up     20120412131921  Remove rememberable from users
10:40:02    up     20120609094354  Add admin flag to users
10:40:02    up     20120609094747  Add indices for user
10:40:02    up     20120611131259  Create permissions
10:40:02    up     20120612135328  Devise invitable add to users
10:40:02    up     20120619152849  Insert permissions data
```

PR #1476, opened 2020-07-29
Switch from SES to Notify in ActionMailer

benjamineskola:
https://trello.com/c/ZB29y9Y0/2050-signon-use-notify-instead-of-amazon-ses-%F0%9F%8D%90


benjamineskola:
Deployed the branch and got the following error:

```OpenSSL::SSL::SSLError: hostname "localhost" does not match the server certificate```

https://sentry.io/organizations/govuk/issues/1813665209/?environment=integration-blue-aws&project=202251&query=is%3Aunresolved#exception

Likely because the delivery method hasn't been properly set for integration. Have reverted to the previous released tag and will investigate in the morning.

PR #1482, opened 2020-08-10
Override reset_password_instructions mailer method for devise to use Notify

benjamineskola:
https://trello.com/c/ZB29y9Y0/2050-signon-use-notify-instead-of-amazon-ses-%F0%9F%8D%90


benjamineskola:
>Does this mean `app/views/devise/mailer/reset_password_instructions.text.erb` and `app/views/devise/mailer/reset_password_instructions.html.erb` can now be removed?

Yes, those are probably unnecessary now, I'll remove those too.

Just working out the test failures at the moment.

PR #1570, opened 2021-01-15
Manually set APHA parent to Defra 

thomasleese:
This updates the organisation fetcher to manually fix APHA to set its parent to Defra. This introduces tech debt, but it was deemed to be the most practical option. Defra is a better parent for APHA than what it ends up being set to (The Welsh Government) as often editors in Defra will need access to content published by APHA.

It's worth noting, we've introduced a migration to make this change on three different occasions, but each time it's wiped when the fetcher runs:

- #574
- #1157
- #1481

At the moment the chosen parent will be the one that is last in the list ordered alphabetically. We wondered whether changing it to be the first in the list as stored in Whitehall was better, but although it fixes the parent for APHA, it introduces some different parents to other organisations that doesn't seem quite right. We felt that this wasn't a decision for us to make.

[Trello Card](https://trello.com/c/7koTf4OD/2293-investigate-prevent-parent-organisation-being-swapped-for-another-in-signon-timebox-2-days)

### DIFF
--- a/lib/organisations_fetcher.rb
+++ b/lib/organisations_fetcher.rb
@@ -5,10 +5,9 @@ require "gds_api/organisations"
 # membership of organisations.
 class OrganisationsFetcher
   def call
-    organisation_relationships = {}
-    organisations.each do |organisation_data|
+    organisation_relationships = organisations.each_with_object({}) do |organisation_data, memo|
       update_or_create_organisation(organisation_data)
-      organisation_relationships[organisation_data["details"]["slug"]] = child_organisation_slugs(organisation_data)
+      memo[organisation_data["details"]["slug"]] = child_organisation_slugs(organisation_data)
     end
     # Now that any new organisations have been created and any slug changes
     # have been applied, we can safely tie together organisations
@@ -39,17 +38,18 @@ private
       abbreviation: organisation_data["details"]["abbreviation"],
       closed: organisation_data["details"]["govuk_status"] == "closed",
     }
+
     organisation.update!(update_data)
   end
 
   def child_organisation_slugs(organisation_data)
-    organisation_data["child_organisations"].collect { |child_organisation| child_organisation["id"].split("/").last }
+    organisation_data["child_organisations"].map { |child_organisation| child_organisation["id"].split("/").last }
   end
 
   def update_ancestry(organisation_relationships)
     organisation_relationships.each do |organisation_slug, child_organisation_slugs|
-      parent = Organisation.find_by(slug: organisation_slug)
-      Organisation.where(slug: child_organisation_slugs).map do |child_organisation|
+      parent = Organisation.find_by!(slug: organisation_slug)
+      Organisation.where(slug: child_organisation_slugs).find_each do |child_organisation|
         # TODO: this ignores that organisations can have multiple parents. I think organisations will
         # end up with the parent that appears last in the API response(s).
         #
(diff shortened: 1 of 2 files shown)


AlanGabbianelli (commented):
I understand this is only needed for APHA at the moment but I wonder if we could be more generic and test that this applies to all organisations in `MANUAL_PARENT_FIXES`. Could the test be `"it applies all manual parent fixes"` and iterate over `MANUAL_PARENT_FIXES`?

thomasleese:
> I understand this is only needed for APHA at the moment but I wonder if we could be more generic and test that this applies to all organisations in `MANUAL_PARENT_FIXES`. Could the test be `"it applies all manual parent fixes"` and iterate over `MANUAL_PARENT_FIXES`?

Hmm, I'm not sure I like the idea of the tests iterating over `MANUAL_PARENT_FIXES` since it's a private member of the class.  I think you're right about the tests though, so I've amended the commit to make it easier for users to add more tests in the future by iterating over a dictionary.

lib/organisations_fetcher.rb
```diff
@@ -50,12 +51,27 @@ def update_ancestry(organisation_relationships)
...
+    "animal-and-plant-health-agency" => "department-for-environment-food-rural-affairs",
+  }.freeze
+
+  def fix_parents_manually
+    # Signon doesn't support multiple parents. Most of the time this is fine, but for
+    # certain organisations this leads to a poor user experience and frequent support
+    # tickets.
+    MANUAL_PARENT_FIXES.each do |child_slug, parent_slug|
+      child = Organisation.find_by(slug: child_slug)
+      parent = Organisation.find_by(slug: parent_slug)
```
ChrisBAshton, line 71: These should both be `find_by!`, I think?

test/lib/organisations_fetcher_test.rb
```diff
@@ -113,4 +113,19 @@ class OrganisationsFetcherTest < ActiveSupport::TestCase
       OrganisationsFetcher.new.call
     end
   end
+
+  {
+    "animal-and-plant-health-agency" => "department-for-environment-food-rural-affairs",
```
ChrisBAshton, line 118: I agree with Alan that it would be nicer to iterate over `MANUAL_PARENT_FIXES` rather than repeat the config in two places.

I think we could either move `MANUAL_PARENT_FIXES` to the top of the class so that it's public (so long as the `.freeze` is on it, it should be safe to do that I think), or we could use [`instance_variable_set`](https://apidock.com/ruby/Object/instance_variable_set) to mock the `MANUAL_PARENT_FIXES` variable with some foobar values. Both would be preferable to duplicating the config IMO.

ChrisBAshton (commented):
Well done for getting to the bottom of this! 👍 Code LGTM, just two small suggestions below.

lib/organisations_fetcher.rb
thomasleese: I left it as `find_by` intentionally so that if an organisation was deleted in the future, this task wouldn't break, it would just ignore the organisations that no longer exist.

test/lib/organisations_fetcher_test.rb
thomasleese: Since you both think we should just iterate over `MANUAL_PARENT_FIXES`, I've amended the commit to do that!

lib/organisations_fetcher.rb
ChrisBAshton: Ah yes - makes sense.

ChrisBAshton (approved):
LGTM 👏

AlanGabbianelli (approved):

PR #1602, opened 2021-03-22
Fix issue with Chromedriver not clicking a button off the viewport

barrucadu:
A recent upgrade to Chromedriver means that `click_button` only works
if the element is visible on the screen.  Which it's not in this test.
So fix the test by scrolling to the button first.


sihugh (approved):
🤦

PR #1615, opened 2021-04-13
Update troubleshooting documentation

brucebolt:
Updating the documentation on manually sending a password reset email to a user:

- offers an alternative method that does not involve logging into a   production console
- accounts for Signon having moved from Carrenza to AWS
- adds instructions to use the gds-cli
- fixed formatting of the main header

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,19 +1,49 @@
-# Troubleshooting
+# Troubleshooting
 
 This is a place to record workarounds to some common problems.
 
 ## User didn't receive password reset email
 
-All the emails do is send users a link to reset their password. We can find this link and give it to them directly.
+When a user account is created, they are sent a link to reset their password. A similar email is sent when users request a password reset email.
 
-First, open a Rails console (Signon is on the backend box in Carrenza).
+In the first instance, a new email should be requested. The route is different depending on the type of password reset email.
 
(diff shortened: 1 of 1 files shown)


bilbof (approved):
It might be worth also highlighting that we can resend a signup email through the app, and that should be the first (perhaps only?) way to do this. What do you think?

Screenshot 2021-04-13 at 16 44 44

brucebolt:
> It might be worth also highlighting that we can resend a signup email through the app, and that should be the first (perhaps only?) way to do this. What do you think?
> 

Thanks, I'll add that in.

I believe these docs were initially added to solve a problem of certain departments filtering our password reset emails (like the NHS email we saw with email subscriptions), which is how I ended up reading them.

docs/troubleshooting.md
```diff
@@ -1,19 +1,49 @@
...
+
+As a superadmin user, in Signon, find the user and press the 'Resend signup email' button.
+
+### Password reset email
+
+The user can request a new password reset link themselves through the [password reset form](https://signon.publishing.service.gov.uk/users/password/new).
+
+### User does not receive the email
+
+If a user does not receive this email after attempting to reteieve it again themselves, you may wish to send it directly to their email address.
```
benlovell, line 21: _retrieve_

PR #1617, opened 2021-04-15
Add API for bearer token creation and rotation

bilbof:
This commit adds an API for the automated creation and rotation of ApiUser bearer tokens

## Motivation

**We don't rotate secrets generated by Signon often**. It is a manual process to generate a new bearer token, add it to our secrets store (hiera-eyaml), deploy Puppet, and then delete the old token.

However, we will soon be moving GOV.UK from EC2 to [ECS](https://aws.amazon.com/ecs/) (the replatforming team's goal).

As a side effect of the migration, we will change our secret store from the git-based hiera-eyaml (puppet/govuk-secrets) to AWS SecretsManager.

This gives us an opportunity to automate the rotation of Signon credentials, since AWS supports rotation of SecretsManager credentials natively.

## Implementation

Use a [SecretsManager Rotation Lambda](https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotating-secrets-lambda-function-overview.html) to create and rotate the bearer tokens (using the API added in this PR) for the cluster. The API is not enabled in the EC2 environment, and uses bearer token auth and IP-restriction in ECS.

We considered implementing a Terraform provider for Signon. However, Terraform would likely store the secret in its statefile so this was a blocker to this approach.

## Benefit

This enables us to rotate bearer tokens automatically.

## Further work

* Implement a rotation Lambda for creation and rotation of bearer tokens in ECS (https://github.com/alphagov/govuk-infrastructure/pull/202)
* Automate deletion of old unused bearer tokens (only those created by the new platform in ECS)
* Implement rotation of OAuth application secret keys using a Lambda
* Remove UI for Applications and ApiUsers, and move all config into code (after replatforming).

### DIFF
--- a/app/controllers/api/v1/authorisations_controller.rb
+++ b/app/controllers/api/v1/authorisations_controller.rb
@@ -1,0 +1,95 @@
+class Api::V1::AuthorisationsController < ApplicationController
+  before_action :authenticate
+  before_action :validate_create_params, only: %w[create]
+  before_action :validate_test_params, only: %w[test]
+  before_action :api_user
+
+  skip_after_action :verify_authorized
+  protect_from_forgery with: :null_session
+
+  rescue_from ActionController::ParameterMissing, with: :missing_params_error
+  rescue_from ActiveRecord::RecordNotFound, with: :not_found_error
+
+  respond_to :json
+
+  def create
+    authorisation = api_user.authorisations.build(expires_in: ApiUser::DEFAULT_TOKEN_LIFE)
+    authorisation.application_id = application.id
+    ActiveRecord::Base.transaction do
+      authorisation.save!
+      grant_app_permissions!(authorisation, params.fetch(:permissions, []))
+    end
+    EventLog.record_event(
+      api_user,
+      EventLog::ACCESS_TOKEN_GENERATED,
+      initiator: api_user,
+      application: authorisation.application,
+      ip_address: request.remote_ip,
+    )
+    render json: { application_name: authorisation.application.name, token: authorisation.token }
+  end
+
+  def test
+    authorisation = api_user.authorisations
+      .find_by!(
+        application: application,
+        token: params.require(:token),
+      )
+
+    render json: { application_name: authorisation.application.name }
+  end
+
+private
+
+  BEARER_TOKEN_VAR = "SIGNON_ADMIN_PASSWORD".freeze
+
+  def authenticate
+    authenticate_or_request_with_http_token do |token, _options|
+      if ENV.key? BEARER_TOKEN_VAR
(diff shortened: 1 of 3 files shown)


bilbof:
If this design is accepted, it would probably be worth moving the config to Terraform / Concourse, rather than having JSON in config in the app. It would enable bootstrapping with a subset of apps / api_users, or a completely different configuration, if this app is used by other departments.

lib/configure/api_users.rb
```diff
@@ -0,0 +1,33 @@
+module Configure
+  class ApiUsers
+    def initialize(namespace:, resource_name_prefix:)
+      @namespace = namespace
+      @name_prefix = resource_name_prefix
+    end
+
+    def configure!(api_users)
+      api_users.each do |api_user|
+        email = [namespace, "#{api_user.fetch('slug')}@alphagov.co.uk"].compact.join("-")
```
richardTowers, line 10: `alphagov.co.uk` - weird to see this domain - is this a legacy thing that signon already uses? Does it actually send emails to or from these addresses?

richardTowers (commented):
A few initial comments after a first pass. I think I might need a fresh brain to review this properly. I might need another pair of eyes too.

The new API we're adding is definitely security-adjacent, so I want to be confident that we're  not missing any edge cases where it could be used For Evil.

config/api_users.json
```diff
@@ -0,0 +1,294 @@
...
+    "slug": "metadata-api"
+  },
+  {
+    "name": "Need API",
+    "slug": "need-api"
+  },
+  {
+    "name": "New London: Trade Tariff Admin (PaaS)",
+    "slug": "trade_tariff_admin"
+  },
```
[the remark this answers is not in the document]
bilbof: Good idea -- I've removed all apps and api users that we don't currently need.

lib/configure/api_users.rb
bilbof: This domain was the original domain for the development of GOV.UK. For legacy reasons, we still use it for a few things, [list here](https://docs.google.com/document/d/1qdYVIU8dI_CqmwYwwxSfORHEPH-9gX8m0cAJOZcwVOs/edit#heading=h.axp1xnscuvsv).

I've changed it to `digital.cabinet-office`, which seems most appropriate. I don't think we intentionally send emails to these addresses.

bilbof:
@richardTowers thanks for your review. I've made some changes, and I've split out the bootstrap task into a separate PR (https://github.com/alphagov/signon/pull/1623).

karlbaker02 (approved):
Looks solid, great work! 💯 All test paths look to be covered, this should make rotation far easier in the future.

PR #1620, opened 2021-04-19
Separate out API token healthcheck & add RFC 141 healthcheck endpoints

barrucadu:
The API token check has value, but not as a readiness healthcheck.  So,
following a pattern from whitehall and a few other apps, I'm giving it
its own route, and it'll get its own Icinga alert.

Once govuk-puppet has been updated, the old endpoint can be removed.

See the RFC for more details.

---

[Trello card](https://trello.com/c/tl6RV1el/723-improve-govuk-healthchecks)

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -1,0 +1,7 @@
+class HealthcheckController < ApplicationController
+  skip_after_action :verify_authorized
+
+  def api_tokens
+    render json: Healthcheck::ApiTokens.new.to_hash
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,6 +6,12 @@ Rails.application.routes.draw do
         Healthcheck::ApiTokens,
       )
 
+  get "/healthcheck/live", to: proc { [200, {}, %w[OK]] }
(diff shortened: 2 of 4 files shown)


huwd (approved):

AlanGabbianelli (approved):

app/controllers/healthcheck_controller.rb
```diff
@@ -0,0 +1,7 @@
+class HealthcheckController < ApplicationController
+  skip_after_action :verify_authorized
+
+  def api_tokens
+    render json: Healthcheck::ApiTokens.new.to_hash
```
huwd, line 5: 🤔 actually how does `.new` work here. I can't see an initalizer in the class and it doesn't inherit?
https://github.com/alphagov/signon/blob/master/lib/healthcheck/api_tokens.rb

barrucadu: It just doesn't have a constructor

huwd: cool cool, just tested it and all looks good - rare I see a class that doesn't have any parameters.
I think more often i've seen this sort of thing done with a

```ruby
module Foo
  class << self
    def call; end
  end
end
```

sort of thing

PR #1623, opened 2021-04-22
Add bootstrap rake task

bilbof:
This adds a rake task `bootstrap:all` which will enable us to configure signon with the OAuth applications and api users required to run in a GOV.UK cluster.

For now I've only added the apps required for the core GOV.UK cluster - that are currently running in ECS. We'll add others as needed.

Motivation
----

**Bootstrapping Signon is slow**. Currently when bringing up GOV.UK in a new environment, one must either copy an existing database to the new environment (making appropriate changes to application URLs, creating new tokens) or manually create all of the configuration required.

We want to automate Signon's bootstrapping process (the creation of Signon resources).

Implementation
----

We will run a new rake task `bootstrap:all[integration.publishing.service.gov.uk]` when deploying Signon to a new environment/namespace. This will create the ApiUsers and Applications necessary for the cluster.

Since the task is safe to re run, we can run it during deploy.

Not handled here is deletion of secrets - it might be preferable in future to expose the application/api_user models through a CRUD api. For now we'll need to continue manually deleting secrets.

This implementation supports creating resources for a namespace, which is useful for running multiple clusters of govuk that use the same Signon database - as will be the case in the ECS test environment.

It might make sense to move the JSON config over to govuk-infrastructure, and provide that as input, but I've included it here for now.

### DIFF
--- a/config/api_users.json
+++ b/config/api_users.json
@@ -1,0 +1,26 @@
+[
+  {
+    "name": "Content Store",
+    "slug": "content-store"
+  },
+  {
+    "name": "Draft Content Store",
+    "slug": "draft-content-store"
+  },
+  {
+    "name": "Frontend App",
+    "slug": "frontend"
+  },
+  {
+    "name": "Publisher App",
+    "slug": "publisher"
+  },
+  {
+    "name": "Publishing API",
+    "slug": "publishing-api"
+  },
+  {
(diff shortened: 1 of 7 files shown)


karlbaker02 (approved):
Once again very comprehensive, looks good 🥇

PR #1631, opened 2021-05-05
Use provided domain for ApiUser emails

bilbof:
The domain `digital.cabinet-office.gov.uk` isn't owned by us, and GDS IT run email on it so there is a chance of a conflict.

This enables us to use workspace/environment specific domains in emails, such as test.govuk.digital where a conflict is less likely.

### DIFF
--- a/test/lib/configure/api_users_test.rb
+++ b/test/lib/configure/api_users_test.rb
@@ -1,14 +1,18 @@
 require "test_helper"
 
 class ConfigureApiUsersTest < ActiveSupport::TestCase
+  public_domain = "example.gov.uk"
+
   test "creates required api users" do
     Configure::ApiUsers.new(
-      namespace: nil, resource_name_prefix: nil,
+      namespace: nil,
+      resource_name_prefix: nil,
+      public_domain: public_domain,
     ).configure!(api_users)
 
     api_users.each do |api_user|
       assert ApiUser.exists?(
-        email: "#{api_user.fetch('slug')}@digital.cabinet-office.gov.uk",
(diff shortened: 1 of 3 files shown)


test/lib/configure/api_users_test.rb
```diff
@@ -1,14 +1,18 @@
...
-      namespace: nil, resource_name_prefix: nil,
+      namespace: nil,
+      resource_name_prefix: nil,
+      public_domain: public_domain,
     ).configure!(api_users)
 
     api_users.each do |api_user|
       assert ApiUser.exists?(
-        email: "#{api_user.fetch('slug')}@digital.cabinet-office.gov.uk",
+        email: "#{api_user.fetch('slug')}@#{public_domain}",
```
sengi, line 15: Minor/optional thing: This might be a Ruby culture thing that I'm unused to, but doing unnecessary substitutions like this in test assertions would be considered a bit iffy in some circles e.g. https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html (It doesn't particularly bother me in this case but thought I'd mention it in passing - I won't be upset if you wanna leave it as is 😄)

sengi (approved):
Cool, thanks for fixing this up!

test/lib/configure/api_users_test.rb
bilbof: Interesting, thanks! I hadn't seen that advice before. I see this pattern regularly in Ruby land, but I can see the advice applies here. I think this approach is okay for now since the interpolation is pretty simple - but I agree that complex and flexible tests are often worse for readability than simple and inflexible tests.

PR #1641, opened 2021-05-12
Add get OAuth application API endpoint

bilbof:
This adds the endpoint `GET /api/v1/applications?name=Publishing+API` which will return the OAuth ID and Secret for a Doorkeeper App.

The endpoint is IP-restricted and requires a bearer token.

This will be used by the Concourse deploy pipeline when deploying apps. It's not sufficient to provide only a create endpoint, since we can be in a state where an application has been created but the generated secrets don't yet exist in SecretsManager. This endpoint enables the deploy pipeline to recover from this state.

Trello: https://trello.com/c/elhnwnaf/454-bootstrapping-signon-oauth-app-secrets

### DIFF
--- a/test/integration/api/applications_test.rb
+++ b/test/integration/api/applications_test.rb
@@ -7,13 +7,49 @@ class ApplicationsTest < ActionDispatch::IntegrationTest
   description = "Herds cats"
   permissions = %w[cat_herder herder_of_cats]
   endpoint = "/api/v1/applications"
-  params = {
+  create_params = {
     "name" => name,
     "home_uri" => home_uri,
     "redirect_uri" => redirect_uri,
     "description" => description,
     "permissions" => permissions,
   }
+
+  test "#show responds with a 401 error when an invalid token is given" do
+    ENV["SIGNON_ADMIN_PASSWORD"] = SecureRandom.uuid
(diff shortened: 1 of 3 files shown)


test/integration/api/applications_test.rb
```diff
@@ -7,14 +7,50 @@ class ApplicationsTest < ActionDispatch::IntegrationTest
...
+  test "#show responds with a 401 error when SIGNON_ADMIN_PASSWORD env var is unset" do
+    ENV["SIGNON_ADMIN_PASSWORD"] = nil
+    get endpoint
+    assert_unauthorized(response)
+  end
+
+  test "#show responds with a 400 when required params are missing" do
+    get_req(endpoint, params: {})
+    assert_equal 400, response.status
+    assert_equal JSON.generate({ error: "param is missing or the value is empty: name" }), response.body
```
sengi, line 33: Ideally I think it'd be better not to assert on the entire message verbatim in cases like these, but if the message is unlikely to ever change then it's fairly harmless. (No need to change them unless you're super keen to - just something to maybe consider for future tests.)

sengi (approved):
LGTM. The level of abstraction in these security-critical tests makes me a bit nervous, but I think it's a case of Normal For Ruby and I didn't spot any logic issues so 🤞😅

PR #1642, opened 2021-05-12
Allow OAuth apps to have no special permissions

bilbof:
When creating an OAuth app one can specify the permissions that the app has. However, some apps don't have any permissions - this change adds support for those apps.

### DIFF
--- a/test/integration/api/applications_test.rb
+++ b/test/integration/api/applications_test.rb
@@ -46,9 +46,9 @@ class ApplicationsTest < ActionDispatch::IntegrationTest
     assert_equal 200, response.status
     body = JSON.parse(response.body)
     assert_equal 43, body.fetch("oauth_id").length
-    assert_match(/[A-Za-z0-9]\w+/, body.fetch("oauth_id"))
+    assert_match(/[a-zA-Z0-9\-_]/, body.fetch("oauth_id"))
     assert_equal 43, body.fetch("oauth_secret").length
-    assert_match(/[A-Za-z0-9]\w+/, body.fetch("oauth_secret"))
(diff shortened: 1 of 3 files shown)


test/integration/api/applications_test.rb
```diff
@@ -92,6 +92,16 @@ class ApplicationsTest < ActionDispatch::IntegrationTest
     assert_match(/[A-Za-z0-9]\w+/, body.fetch("oauth_secret"))
   end
 
+  test "#create with no permissions is successful" do
+    post_req(endpoint, params: create_params.merge("permissions" => []))
+    assert_equal 200, response.status
+    body = JSON.parse(response.body)
+    assert_equal 43, body.fetch("oauth_id").length
+    assert_match(/[A-Za-z0-9]\w+/, body.fetch("oauth_id"))
```
sengi, line 100: Just to double-check the regex here as it seems oddly specific: is it the case that our OAuth client IDs and secrets can contain `_` but not start with `_`?

sengi (approved):

test/integration/api/applications_test.rb
bilbof: Thanks, I've pushed up another commit to correct the regex.

sengi: I think you're now just checking that it contains at least one alpha (or `-` or `_`) character. Did you drop the `+` intentionally? Also I just realised these regexes are unanchored so if you wanted to check that the whole string consists of only those characters you'd need something like `/^[A-Za-z0-9_-]+$/`

bilbof: I merged too soon, sorry - I've put up a separate PR for this https://github.com/alphagov/signon/pull/1643

PR #1644, opened 2021-05-17
Return 409 status on failed recreate OAuth App call

bilbof:
We'll return a 409 Conflict status code and a more appropriate error message when we cannot create an application or its associated resources due to a RecordNotUnique error. This API's only client is the Concourse deploy pipeline, so it's an easy change.


fredericfran-gds (approved):
ok

PR #1645, opened 2021-05-17
Expire bearer tokens in 60 days

bilbof:
This changes the bearer token expiry time from 2 years to 60 days for tokens that are created via the API.

The expectation is that tokens created via the API will be automatically rotated every 30 days. This is currently the case for bearer tokens created for the platform in ECS.

If a rotation fails, it retries daily for a few days - we should be able to fix the issue within the 30 days after a rotation fails.

This does _not_ change anything in the EC2 platform - we'll still have long-lived tokens for apps there, since we must rotate tokens in that platform manually.


fredericfran-gds (approved):
LGTM

PR #1707, opened 2021-08-23
Ignore expiring tokens alert for tokens created by Lambda

bilbof:
This modifies the expiring tokens alert to ignore the expiration
of tokens that are managed by the Lambda rotation process. We're
taking advantage of the difference in email domain to filter the
alert (alphagov is typically used in production, while the Lambda
uses the environment's public domain).

It is OK for these tokens to expire, since the Lambda will create
new tokens, and leave the old tokens to expire without deleting them.

In future we'll create a new alert for monitoring Lambda rotation
failures.

https://trello.com/c/QbkAqyQW/570-signon-resolve-ux-alerting-issue


sengi (approved):
LGTM - thanks for fixing up the tests

PR #1759, opened 2021-12-06
Upgrade CI to use MySQL 8

davidgisbey:
This changes the Jenkins CI configuration to use MySQL 8.

MySQL is [available at port `33068 `](https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-on-jenkins-ci.html#specifying-which-database-to-use) on the CI server. By explicitly setting the `TEST_DATABASE_URL`, we're telling Rails to use that MySQL server.

Previously, Rails implicitly used the default MySQL port `3306` which is a MySQL 5.5 server.

Trello ticket: https://trello.com/c/btcm7JyB/

The database config is already using the correct ENV var 

image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,4 +10,7 @@ node {
   govuk.buildProject(
     brakeman: true,
   )
+
+  // Run against the MySQL 8 Docker instance on GOV.UK CI
+  govuk.setEnvar("TEST_DATABASE_URL", "mysql2://root:root@127.0.0.1:33068/signonotron2_test")
 }


Jenkinsfile
```diff
@@ -10,4 +10,7 @@ node {
   govuk.buildProject(
     brakeman: true,
   )
+
+  // Run against the MySQL 8 Docker instance on GOV.UK CI
+  govuk.setEnvar("TEST_DATABASE_URL", "mysql2://root:root@127.0.0.1:33068/signonotron2_test")
```
ollietreend, line 15: Odd database name 😂 `signonotron2` doesn't seem like a very meaningful name.
But 👍🏻 for being consistent with what's already used in `database.yml`

ollietreend (approved):

Jenkinsfile
davidgisbey: Yea haha. Very whimsical

PR #1768, opened 2021-12-22
Add ApiUsers API endpoint

bilbof:
This adds the `/api/v1/api-users` resource and changes the `/api/v1/authorisations` endpoint to be a sub-resource of api-users: `/api/v1/api-users/:id/authorisations`. This API is currently not in use so we are safe to break the API without modifying the version.

This addition enables the only client ([Signon bootstrap rake task](https://github.com/alphagov/govuk-infrastructure/pull/499)) to create all Signon resources necessary during cluster turnup. Previously ApiUsers were created using a Signon rake task.

Trello: https://trello.com/c/fUXJ6X1U/725-create-signon-resources-when-signon-is-deployed

### DIFF
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -1,0 +1,14 @@
+class Api::V1::ApiController < ApplicationController
+  include AdminApiHelper
+
+  before_action :authenticate
+
+  skip_after_action :verify_authorized
+  protect_from_forgery with: :null_session
+
+  rescue_from ActionController::ParameterMissing, with: :missing_params_error
+  rescue_from ActiveRecord::RecordInvalid, with: :not_valid_error
+  rescue_from ActiveRecord::RecordNotFound, with: :not_found_error
+
+  respond_to :json
+end
--- a/app/controllers/api/v1/api_controller.rb
+++ b/app/controllers/api/v1/api_controller.rb
@@ -9,6 +9,7 @@ class Api::V1::ApiController < ApplicationController
   rescue_from ActionController::ParameterMissing, with: :missing_params_error
   rescue_from ActiveRecord::RecordInvalid, with: :not_valid_error
   rescue_from ActiveRecord::RecordNotFound, with: :not_found_error
+  rescue_from ActiveRecord::RecordNotUnique, with: :already_created_error
 
   respond_to :json
 end
--- a/app/controllers/api/v1/api_users_controller.rb
+++ b/app/controllers/api/v1/api_users_controller.rb
@@ -20,6 +20,11 @@ private
                            password: password, password_confirmation: password)
     api_user.skip_confirmation!
     api_user.api_user = true
+
+    if api_user.invalid? && api_user.errors.where(:email, :taken).any?
+      raise ActiveRecord::RecordNotUnique
+    end
+
     api_user.save!
     api_user
   end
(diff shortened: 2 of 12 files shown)


app/controllers/api/v1/api_users_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Api::V1::ApiUsersController < Api::V1::ApiController
+  before_action :validate_create_params, only: %w[create]
+  before_action :validate_show_params, only: %w[show]
+
+  def create
+    api_user = create_api_user(name: params.fetch(:name), email: params.fetch(:email))
+
```
sengi, line 7: tiny whitespace nit (just so it's consistent with the one below)
```suggestion
```

lib/data_hygiene/bulk_organisation_updater.rb
```diff
@@ -2,8 +2,9 @@
 
 module DataHygiene
   class BulkOrganisationUpdater
-    def initialize(filename)
+    def initialize(filename, logger: Logger.new($stdout))
```
sengi, line 5: Thanks for sorting out the logging!

sengi (approved):
LGTM (with the caveat that I'm not a Rubyist so can't really review for style)

app/controllers/api/v1/api_users_controller.rb
bilbof: Fixed :)

app/controllers/api/v1/api_controller.rb
```diff
@@ -0,0 +1,14 @@
+class Api::V1::ApiController < ApplicationController
```
hannako, line 1: Satisfying refactor :-)

app/helpers/admin_api_helper.rb
```diff
@@ -19,8 +19,8 @@ def not_found_error(_exception)
     render json: { error: "Not found" }, status: :not_found
   end
 
-  def not_valid_error(_exception)
-    render json: { error: "Not valid" }, status: :bad_request
+  def not_valid_error(exception)
+    render json: { error: exception.message }, status: :bad_request
```
hannako, line 27: I'm not sure I follow this commit. What was missing from the exception messages previously to make them un-useful? ie why does using the application id make the difference between them being useful or not?

app/helpers/admin_api_helper.rb
```diff
@@ -11,6 +11,10 @@ def authenticate
     end
   end
 
+  def already_created_error(_exception)
+    render json: { error: "Record not unique" }, status: :conflict
```
hannako, line 15: Very minor comment:

If you've using "Record not unique", might be nice to amend line 23 so that's it's "Record not found".

hannako (approved):
Looks great! Just a few nitpick comments you are welcome to ignore :-)

app/helpers/admin_api_helper.rb
bilbof: Previously the error message was a generic "Not valid", after this change the returned error message will be (e.g. [from test](https://github.com/alphagov/signon/pull/1768/commits/b620dd4d900c1a914946de5feb4282b24107a63d#diff-0e0ce7463121fd2e27cdc242c28a84fa04413d272034d54c78f8019000062b5eR71)) "Validation failed: Redirect URI must be an absolute URI." which clients can use to more easily debug why a request failed.

app/helpers/admin_api_helper.rb
bilbof: Thanks, I've done that

app/helpers/admin_api_helper.rb
hannako: Sure, I just wondered why we were previously setting a response of "Not valid" rather than returning `exception.message`

bilbof: Yeah I'm not sure why I didn't do this previously 🤔

hannako: Ah OK that's fine! I thought there was a clever reason 😂

PR #1771, opened 2022-01-06
Remove lhma_2019_04_29_13_15_20_064_event_logs table

kevindew:
Trello: https://trello.com/c/7Tr2EOzG/67-plan-production-upgrade-for-signon-mysql-database

This table was created in 2019 [1] as part of a migration to the event
logs table (lhm is a reference to Large Hadron Migrator [2]). I don't think
it was actually intended to be ever part of the DB schema and just an
artefact of dev usage.

I've used a `if_exists: true` as this table doesn't seem to exist in all
environments.

As part of this work I've manually removed some tables from the production
database that weren't listed in the DB schema: __event_logs_old,
lhma_2019_04_30_10_36_45_946_event_logs and lhma_2019_06_18_13_31_49_199_event_logs
which were collectively using about 8GB of storage.

[1]: https://github.com/alphagov/signon/pull/1771/commits/cc9bac9b4817913ef9bae96234905ec1bce49fa6
[2]: https://github.com/soundcloud/lhm

### DIFF
--- a/db/migrate/20220106144058_remove_old_diagnostic_table.rb
+++ b/db/migrate/20220106144058_remove_old_diagnostic_table.rb
@@ -1,0 +1,9 @@
+class RemoveOldDiagnosticTable < ActiveRecord::Migration[6.1]
+  def up
+    drop_table :lhma_2019_04_29_13_15_20_064_event_logs, if_exists: true
+  end
+
+  def down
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
Nice one ⭐

PR #1775, opened 2022-01-17
Remove lhm dependency

barrucadu:
lhm[1] is a tool for doing large data migrations, and signon has used
it twice:

- In 2017 to add IP address and user agent to the event logs
- In 2019 to change how IP addresses are stored in the event logs

lhm hasn't been updated since 2018, so relying on it is a bit
questionable.  Furthermore, it requires either superuser access to
mysql, or the `log_bin_trust_function_creators` option[2] be enabled,
which the mysql docs make sound like less than ideal security
practice.  Admittedly, we implicitly trust lhm already.  By using it
we're giving it access to our production database.  But it's always
good to minimise the privilege something needs.

Since we've used this twice, I think we should remove lhm and find an
alternative approach to large data migrations instead.

See also https://github.com/alphagov/govuk-aws-data/pull/963

[1] https://github.com/soundcloud/lhm
[2] https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#sysvar_log_bin_trust_function_creators

---

[Trello card](https://trello.com/c/0pLBqqYJ/93-decide-if-we-need-the-logbintrustfunctioncreators-mysql-config)


richardTowers (approved):
I agree.

If it comes to it, we can always re-add lhm before the next big migration and remove it afterwards. It doesn't need to be there all the time, just in case there's a big migration.

barrucadu:
> If it comes to it, we can always re-add lhm before the next big migration and remove it afterwards. It doesn't need to be there all the time, just in case there's a big migration.

So, that's a little tricky, as applying the mysql option requires restarting the server.  So adding it back would require downtime.

richardTowers:
Also note (just in case we come back to this PR in the future wanting to use lhm):

> it requires either superuser access, _**or**_ the log_bin_trust_function_creators

So we could probably avoid the downtime of changing the  `log_bin_trust_function_creators` option by running the migration as the database superuser instead.

PR #1783, opened 2022-02-02
Remove non-robust DB migration

bilbof:
This db migration will fail if the Organisations do not exist in the Signon database.


fredericfran-gds (approved):
looks good, discussed on slack

PR #1810, opened 2022-03-11
Migrate from Jasmine gem to jasmine-browser-runner

ollietreend:
The Jasmine Ruby gems have been [deprecated][1] and are no longer supported. And whilst they still work with Rails 6 apps, they [aren't compatible with Rails 7][2] which is now becoming a blocker.

This commit implements the `jasmine-browser-runner` npm package, which is now the recommended approach for running Jasmine tests in the browser, and removes the deprecated Ruby gems.

This includes:

- Add Jasmine NPM packages
- Remove Jasmine Ruby gems
- Update dependabot configuration for the changed dependencies
- Re-implement Jasmine configuration for use by the NPM package

[1]: https://github.com/jasmine/jasmine-gem#discontinued
[2]: alphagov/govuk_publishing_components#2621 (comment)

Trello: https://trello.com/c/21VMx1NV/264-migrate-apps-to-use-jasmine-browser-runner


Gemfile
```diff
@@ -62,7 +60,6 @@ group :test do
   gem "factory_bot_rails"
   gem "minitest"
   gem "mocha", require: false
-  gem "poltergeist"
```
kevindew, line 65: Yay - good catch.

kevindew (approved):
⭐  this looks great

PR #1813, opened 2022-03-21
Add local jQuery

andysellick:
## What
Adds jQuery as a local dependency, instead of relying on the jQuery that comes from the components gem.

**This PR should be deployed immediately prior to the removal of jQuery from GOV.UK** in order to minimise the time when this application will effectively contain two copies of jQuery. This shouldn't cause any errors, but it will lead to an increased asset size.

Note that I've only been able to test the first sign in screen in this application to check it works - not sure how to progress further locally.

## Why
jQuery is going to be removed from the components gem soon.

## Visual changes
None.

Trello card: https://trello.com/c/mAfQNVRG/745-investigate-which-apps-we-could-remove-jquery-from?filter=member:andysellick2

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,3 +1,4 @@
+//= require jquery/dist/jquery
 //= require govuk_publishing_components/dependencies
 //= require govuk_publishing_components/all_components
 //= require password-strength-indicator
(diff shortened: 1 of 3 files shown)


kevindew (approved):
This looks good.

It looks like there is only the password indicator script that really needs jQuery for the non legacy layout so I'm going to have a quick look at if that is solvable - looks a bit hairy though so I don't expect I'll manage it.

PR #1818, opened 2022-03-24
Remove need for jQuery on design system layout

kevindew:
This PR will remove the need to delicately swap jQuery when govuk_publishing_components is released without it as per: https://github.com/alphagov/signon/pull/1813

Principally the work in this PR is to rewrite the password-strength-indicator script. Which was quite an old bit of jQuery we had with some interesting approaches to showing errors. This code is tested through a feature test: https://github.com/alphagov/signon/blob/3301f7028850afd389b9c42e4bef1636b8d326d0/test/integration/password_change_test.rb#L69-L104 which hasn't needed any changes. I looked into writing JS unit tests but that was complex because of the two different layouts JS files.

Here's a quick gif of how it looks for evidence:

Mar-24-2022 12-37-20

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,4 +1,4 @@
 //= require govuk_publishing_components/dependencies
 //= require govuk_publishing_components/all_components
 //= require password-strength-indicator
-//= require jquery_ujs
+//= require rails-ujs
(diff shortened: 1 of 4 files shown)


andysellick (approved):
Tested on integration, looks good 👍

PR #1832, opened 2022-04-05
Remove link to make a support request when logged out

brucebolt:
This link is rendered to users who are not logged in. However the support system is behind Signon, so requires a user to be logged in to make a request. This leads to them being sent back to the same login page.

Removing the link to remove the continuous loop.

When signed out:
Screenshot showing the footer links when signed out.  The "raise support request" link does not appear.

When signed in:
Screenshot showing the footer links when signed in.  The "raise support request" link appears.

[Trello card](https://trello.com/c/A55SZgO4/346-remove-support-link-from-signon-homepage)

### DIFF
--- a/app/views/layouts/admin_layout.html.erb
+++ b/app/views/layouts/admin_layout.html.erb
@@ -44,15 +44,17 @@
       {
         title: "Support and feedback",
         items: [
-          {
-            href: Plek.new.external_url_for("support"),
-            text: "Raise a support request"
-          },
+          (
+            {
+              href: Plek.new.external_url_for("support"),
+              text: "Raise a support request"
+            } if user_signed_in?
+          ),
           {
             href: "https://www.gov.uk/government/content-publishing",
-            text: "How to write, publish, and improve content"
+            text: "How to write, publish, and improve content",
           }
-        ]
+        ].compact
       }
     ]
   } %>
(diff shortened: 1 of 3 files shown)


app/views/layouts/admin_layout.html.erb
```diff
@@ -39,20 +39,25 @@
...
-          {
-            href: Plek.new.external_url_for("support"),
-            text: "Raise a support request"
-          },
-          {
-            href: "https://www.gov.uk/government/content-publishing",
-            text: "How to write, publish, and improve content"
-          }
-        ]
+        items: footer_links,
```
DilwoarH, line 60:
```suggestion
        items: [
          *([
            href: Plek.new.external_url_for("support"),
            text: "Raise a support request"
          ] if user_signed_in?),
          {
            href: "https://www.gov.uk/government/content-publishing",
            text: "How to write, publish, and improve content"
          }
        ],
```

Does this work?

kevindew: That doesn't quite work as the item we need to add is a hash. I've given it an update to bring it closer to your idea - it's a bit code golf like but I think it's ok. Mind giving this a quick re-review?

DilwoarH (approved):

PR #1855, opened 2022-05-10
Improve Dockerfile and allow running as non-root.

sengi:
- Fix Sidekiq config to log to stdout by default but still log to `log/sidekiq.log` in the EC2 environments.
- Use debian-slim as base instead of the full version.
- Multi-stage build to separate out build dependencies from runtime dependencies.
- Add .dockerignore.
- Don't run as root by default.
- Populate the bootsnap cache at build time so it doesn't try to write to the read-only image filesystem.

Image shrinks from ~800M to ~500M.

https://trello.com/c/Jef1t4xU

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+*/log
+.dockerignore
+.git
+.gitignore
+CONTRIBUTING.md
+Dockerfile
+Jenkinsfile
+Procfile
+README.md
+docs
+spec
+test
+tmp
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,55 @@
-ARG base_image=ruby:2.7.6
-FROM ${base_image}
+ARG base_image=ruby:2.7.6-slim-bullseye
 
-# Add yarn to apt sources
-RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
-RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
+FROM $base_image AS builder
+ENV RAILS_ENV=production \
+    NODE_ENV=production \
+    GOVUK_APP_DOMAIN=www.gov.uk \
+    GOVUK_WEBSITE_ROOT=https://www.gov.uk \
+    ASSETS_PREFIX=/assets/signon \
+    BOOTSNAP_CACHE_DIR=/var/cache/bootsnap
+# TODO: have a separate build image which already contains the build-only deps.
+RUN apt-get update -qy && \
+    apt-get upgrade -y && \
+    apt-get clean
+RUN apt-get install -y build-essential nodejs gnupg2 curl libmariadb-dev-compat
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" > /etc/apt/sources.list.d/yarn.list && \
+    apt-get install -y yarn
 
-RUN apt-get update -qq && apt-get upgrade -y
-RUN apt-get install -y build-essential nodejs yarn && apt-get clean
+RUN mkdir /app
+WORKDIR /app
+COPY Gemfile Gemfile.lock .ruby-version /app/
+RUN bundle config set without 'development test' && \
+    bundle install -j8 --retry=2
 
-# This image is only intended to be able to run this app in a production RAILS_ENV
-ENV RAILS_ENV production
+COPY . /app
 
-ENV APP_HOME /app
-ENV DATABASE_URL mysql2://root:root@mysql/signon
-ENV GOVUK_APP_NAME signon
-ENV PORT 3016
-ENV TEST_DATABASE_URL mysql2://root:root@mysql/signon_test
-ENV ASSETS_PREFIX /assets/signon
+RUN bundle exec bootsnap precompile --gemfile -v .
+RUN DEVISE_PEPPER=unused DEVISE_SECRET_KEY=unused bundle exec rails assets:precompile
 
-RUN mkdir $APP_HOME
 
-WORKDIR $APP_HOME
-ADD Gemfile* .ruby-version $APP_HOME/
-RUN bundle config set deployment 'true'
-RUN bundle config set without 'development test'
-RUN bundle install --jobs 4
+FROM $base_image
+ENV RAILS_ENV=production \
+    NODE_ENV=production \
+    GOVUK_APP_NAME=signon \
+    ASSETS_PREFIX=/assets/signon \
+    BOOTSNAP_CACHE_DIR=/var/cache/bootsnap
+WORKDIR /app
 
-ADD . $APP_HOME
+# TODO: have an up-to-date base image and stop running apt-get upgrade here.
+RUN apt-get update -qy && \
+    apt-get upgrade -y && \
+    apt-get clean
+RUN apt-get install -y libmariadb3
 
-RUN GOVUK_APP_DOMAIN=www.gov.uk \
-  DEVISE_PEPPER=`openssl rand -base64 40` \
-  DEVISE_SECRET_KEY=`openssl rand -base64 40` \
-  GOVUK_WEBSITE_ROOT=https://www.gov.uk \
-  bundle exec rails assets:clean assets:precompile
+RUN echo 'IRB.conf[:HISTORY_FILE] = "/tmp/irb_history"' > irb.rc
+COPY --from=builder /usr/bin/node* /usr/bin/
+COPY --from=builder /usr/share/nodejs/ /usr/share/nodejs/
+COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
+COPY --from=builder /var/cache/bootsnap/ /var/cache/
+COPY --from=builder /app ./
 
-HEALTHCHECK CMD curl --silent --fail localhost:$PORT || exit 1
-
+RUN groupadd -g 1001 app && \
+    useradd -u 1001 -g app app
+USER 1001
 CMD bundle exec puma
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,10 +1,11 @@
-# Sample configuration file for Sidekiq.
+# Configuration file for Sidekiq.
 # Options here can still be overridden by cmd line args.
-#   sidekiq -C config.yml
 ---
 :verbose: true
-:concurrency:  2
-:logfile: ./log/sidekiq.log
+:concurrency: 2
+<% if ENV.key?('SIDEKIQ_LOGFILE') %>
+:logfile: <%= ENV['SIDEKIQ_LOGFILE'] %>
+<% end %>
 :queues:
   - default
   - mailers


nsabri1 (approved):
Great stuff

.dockerignore
```diff
@@ -0,0 +1,13 @@
+*/log
```
kevindew, line 1: Should this be the whole log directory? It suprised me that the tmp directory is ignored but not log

Dockerfile
```diff
@@ -1,39 +1,55 @@
...
+RUN mkdir /app
+WORKDIR /app
+COPY Gemfile Gemfile.lock .ruby-version /app/
+RUN bundle config set without 'development test' && \
+    bundle install -j8 --retry=2
+
+COPY . /app
+
+RUN bundle exec bootsnap precompile --gemfile -v .
+RUN DEVISE_PEPPER=unused DEVISE_SECRET_KEY=unused bundle exec rails assets:precompile
```
kevindew, line 28: I don't think you benefit from these being run as part of an earlier stage build (as far as I can tell all the artefacts that are created will be copied over.

I imagine if you moved these you could reduce a bunch of the env vars involved in the first build step.

Dockerfile
```diff
@@ -1,39 +1,55 @@
...
-
-HEALTHCHECK CMD curl --silent --fail localhost:$PORT || exit 1
-
+ARG base_image=ruby:2.7.6-slim-bullseye
+
+FROM $base_image AS builder
+ENV RAILS_ENV=production \
+    NODE_ENV=production \
+    GOVUK_APP_DOMAIN=www.gov.uk \
+    GOVUK_WEBSITE_ROOT=https://www.gov.uk \
```
kevindew, line 7: Should we change these to be "unused" as per the DEVISE examples since we don't expect them to actually affect the app (since it'll probably cause integration/staging problems if they do). 

It's pretty annoying they're needed at boot time rather than at runtime given they're only for plek.

config/sidekiq.yml
```diff
@@ -1,10 +1,11 @@
-# Sample configuration file for Sidekiq.
+# Configuration file for Sidekiq.
```
kevindew, line 1: Nice catch

kevindew (commented):
> Gonna merge this now in order to unblock things, but more than happy to receive post-merge suggestions.

Thanks appreciate that comment, just been catching up now. Nice one putting this together, it looks like you tackled a bunch of problems and great result on shedding 300MB from the image.

I can see you got past the bootsnap cache problem by using an env var for that, however I remain concerned that this isn’t going to be a sufficient solution for all the uses of the Rails tmp directory. I’m expecting that you’re end up playing a long game of whack-a-mole trying to catch all the ways the tmp directory gets used across the suite of Rails app - as it’s a core expectation of Rails and the default place for any gem caching (I’m surprised puma actually starts, as it won’t if done through `bundle exec rails server`). I did have a look see if you could move it but it looks pretty ingrained into rails: https://github.com/rails/rails/blob/main/railties/lib/rails/tasks/tmp.rake

I’d suggest that to avoid a long run of paper cuts, that we find a way to embrace the presence of the app/tmp directory. I know that collides with a principle you’re working towards not having anything write to anything within /app but I think it’s going to be a long battle with various surprising errors.

I had a quick tinker and came up with below, though I imagine a volume mount would be an option too:

```
RUN groupadd -g 1001 app && \
    useradd -u 1001 -g app app && \
    mkdir -p tmp public/assets && \
    chown app:app tmp public/assets

USER 1001

RUN bundle exec bootsnap precompile --gemfile -v .
RUN DEVISE_PEPPER=unused DEVISE_SECRET_KEY=unused bundle exec rails assets:precompile

CMD bundle exec puma
```

Which would move the two tasks that write to tmp to be done by the user.

The other thing I noticed from trying out the image locally was that it seems a few of the Rails tasks expect a home directory (I’m guessing that’s what you were fighting with the .irb file) e.g:

<details>
<summary>Long stack trace</summary>

```
app@4a3214257344:/app$ GOVUK_APP_DOMAIN=www.gov.uk GOVUK_WEBSITE_ROOT=https://www.gov.uk DEVISE_PEPPER=unused DEVISE_SECRET_KEY=unused bundle exec puma
`/home/app` is not a directory.
Bundler will use `/tmp/bundler20220510-10-s6fdt210' as your home directory temporarily.
[10] Puma starting in cluster mode...
[10] * Puma version: 5.6.4 (ruby 2.7.6-p219) ("Birdie's Version")
[10] *  Min threads: 5
[10] *  Max threads: 5
[10] *  Environment: production
[10] *   Master PID: 10
[10] *      Workers: 2
[10] *     Restarts: (✔) hot (✖) phased
[10] * Preloading application
/app/vendor/bundle/ruby/2.7.0/gems/net-protocol-0.1.3/lib/net/protocol.rb:68: warning: already initialized constant Net::ProtocRetryError
/usr/local/lib/ruby/2.7.0/net/protocol.rb:66: warning: previous definition of ProtocRetryError was here
/app/vendor/bundle/ruby/2.7.0/gems/net-protocol-0.1.3/lib/net/protocol.rb:208: warning: already initialized constant Net::BufferedIO::BUFSIZE
/usr/local/lib/ruby/2.7.0/net/protocol.rb:206: warning: previous definition of BUFSIZE was here
/app/vendor/bundle/ruby/2.7.0/gems/net-protocol-0.1.3/lib/net/protocol.rb:504: warning: already initialized constant Net::NetPrivate::Socket
/usr/local/lib/ruby/2.7.0/net/protocol.rb:503: warning: previous definition of Socket was here
[10] * Listening on http://0.0.0.0:3000
[10] Use Ctrl-C to stop
[10] - Worker 0 (PID: 13) booted in 0.0s, phase: 0
[10] - Worker 1 (PID: 15) booted in 0.0s, phase: 0
^C[10] - Gracefully shutting down workers...
Traceback (most recent call last):
	36: from /usr/local/bin/bundle:23:in `<main>'
	35: from /usr/local/bin/bundle:23:in `load'
	34: from /usr/local/lib/ruby/gems/2.7.0/gems/bundler-2.1.4/libexec/bundle:34:in `<top (required)>'
	33: from /usr/local/lib/ruby/2.7.0/bundler/friendly_errors.rb:123:in `with_friendly_errors'
	32: from /usr/local/lib/ruby/gems/2.7.0/gems/bundler-2.1.4/libexec/bundle:46:in `block in <top (required)>'
	31: from /usr/local/lib/ruby/2.7.0/bundler/cli.rb:24:in `start'
	30: from /usr/local/lib/ruby/2.7.0/bundler/vendor/thor/lib/thor/base.rb:476:in `start'
	29: from /usr/local/lib/ruby/2.7.0/bundler/cli.rb:30:in `dispatch'
	28: from /usr/local/lib/ruby/2.7.0/bundler/vendor/thor/lib/thor.rb:399:in `dispatch'
	27: from /usr/local/lib/ruby/2.7.0/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
	26: from /usr/local/lib/ruby/2.7.0/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
	25: from /usr/local/lib/ruby/2.7.0/bundler/cli.rb:476:in `exec'
	24: from /usr/local/lib/ruby/2.7.0/bundler/cli/exec.rb:28:in `run'
	23: from /usr/local/lib/ruby/2.7.0/bundler/cli/exec.rb:63:in `kernel_load'
	22: from /usr/local/lib/ruby/2.7.0/bundler/cli/exec.rb:63:in `load'
	21: from /app/vendor/bundle/ruby/2.7.0/bin/puma:23:in `<top (required)>'
	20: from /app/vendor/bundle/ruby/2.7.0/bin/puma:23:in `load'
	19: from /app/vendor/bundle/ruby/2.7.0/gems/puma-5.6.4/bin/puma:10:in `<top (required)>'
	18: from /app/vendor/bundle/ruby/2.7.0/gems/puma-5.6.4/lib/puma/cli.rb:81:in `run'
	17: from /app/vendor/bundle/ruby/2.7.0/gems/puma-5.6.4/lib/puma/launcher.rb:182:in `run'
	16: from /app/vendor/bundle/ruby/2.7.0/gems/puma-5.6.4/lib/puma/cluster.rb:415:in `run'
	15: from /app/vendor/bundle/ruby/2.7.0/gems/puma-5.6.4/lib/puma/cluster.rb:72:in `spawn_workers'
	14: from /app/vendor/bundle/ruby/2.7.0/gems/puma-5.6.4/lib/puma/cluster.rb:72:in `times'
	13: from /app/vendor/bundle/ruby/2.7.0/gems/puma-5.6.4/lib/puma/cluster.rb:79:in `block in spawn_workers'
	12: from /app/vendor/bundle/ruby/2.7.0/gems/puma-5.6.4/lib/puma/cluster.rb:97:in `spawn_worker'
	11: from /app/vendor/bundle/ruby/2.7.0/gems/activesupport-7.0.2.3/lib/active_support/fork_tracker.rb:18:in `fork'
	10: from /app/vendor/bundle/ruby/2.7.0/gems/activesupport-7.0.2.3/lib/active_support/fork_tracker.rb:18:in `fork'
	 9: from /usr/local/lib/ruby/2.7.0/bundler.rb:666:in `block (2 levels) in tmp_home_path'
	 8: from /usr/local/lib/ruby/2.7.0/bundler.rb:305:in `rm_rf'
	 7: from /usr/local/lib/ruby/2.7.0/bundler/vendor/fileutils/lib/fileutils.rb:744:in `remove_entry_secure'
	 6: from /usr/local/lib/ruby/2.7.0/bundler/vendor/fileutils/lib/fileutils.rb:1508:in `postorder_traverse'
	 5: from /usr/local/lib/ruby/2.7.0/bundler/vendor/fileutils/lib/fileutils.rb:746:in `block in remove_entry_secure'
	 4: from /usr/local/lib/ruby/2.7.0/bundler/vendor/fileutils/lib/fileutils.rb:1453:in `remove'
	 3: from /usr/local/lib/ruby/2.7.0/bundler/vendor/fileutils/lib/fileutils.rb:1464:in `remove_file'
	 2: from /usr/local/lib/ruby/2.7.0/bundler/vendor/fileutils/lib/fileutils.rb:1470:in `platform_support'
	 1: from /usr/local/lib/ruby/2.7.0/bundler/vendor/fileutils/lib/fileutils.rb:1465:in `block in remove_file'
/usr/local/lib/ruby/2.7.0/bundler/vendor/fileutils/lib/fileutils.rb:1465:in `unlink': No such file or directory @ apply2files - /tmp/bundler20220510-10-s6fdt210 (Errno::ENOENT)
[10] === puma shutdown: 2022-05-10 23:46:44 +0000 ===
[10] - Goodbye!
```
</details>

I wonder if it’s worth creating a /home/app directory to resolve the expectation rather than having to fight it.

.dockerignore
sengi: Oh this means any file or directory called log that's exactly one level below the directory that contains `.dockerignore`. Some of the other ones probably need updating.

Dockerfile
```diff
@@ -1,39 +1,55 @@
...
+    apt-get clean
+RUN apt-get install -y build-essential nodejs gnupg2 curl libmariadb-dev-compat
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" > /etc/apt/sources.list.d/yarn.list && \
+    apt-get install -y yarn
+
+RUN mkdir /app
+WORKDIR /app
+COPY Gemfile Gemfile.lock .ruby-version /app/
+RUN bundle config set without 'development test' && \
```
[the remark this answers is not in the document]
sengi: I looked into this in quite some detail and concluded that omitting it was probably the lesser of evils, since it messes with where it installs the gems (because it was designed for a pre-container era). The deciding factor for me was that I really want to avoid the surprise/confusion of having gems installed in more than one place in the image filesystem, since the whole container is inherently specific to a single app. 

I can only see one potential upside to enabling this option: it might detect a scenario where someone has merged a change to `Gemfile` but not `Gemfile.lock` and the tests have nevertheless passed. Personally I think if we're concerned about that scenario then it'd be much kinder to the user if we detect it before merge. (I think the nicest way would be to add a pre-merge check that `bundle update` on a fresh checkout produces no diff.)

If you're still keen on enabling the option, I'd be open to that as long as it still installs the gems in the location that we've configured, i.e. doesn't introduce a second location that's different from the system-wide one. (I'd prefer to write a pre-merge check for Gemfile/Gemfile.lock consistency though!)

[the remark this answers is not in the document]
sengi: Oh just realised that might be a bit confusing cos I think I haven't raised the PR that deals with the GEM_HOME/BUNDLE_PATH stuff yet 😅 I'll keep you posted.

Dockerfile
sengi: There's a practical reason and a philosophical reason for running this in the build stage rather than the final output stage: 
* The practical reason is that in case it produces any temporary/intermediate files or anything else that we don't need/want in the final image, it'll be excluded - we just copy over the things we want and ignore any other gubbins.
* The philosophical reason is that the way the two-stage build is supposed to work is that all the build steps happen in the builder stage and the output stage just copies out the parts of the builder filesystem that are supposed to be in the final image. So even if it happens that some part of the build produces no temp/intermediate files, it's still easier to maintain if we stick with that structure.

Ideally these images wouldn't include the assets files at all and we'd have a separate build step (in GitHub Actions or whatever) to produce a tarball of them which then gets deployed alongside the app image - but that's substantially more faff than is warranted just yet. We're really just including the assets in the image as a short/medium-term bodge to get us up and running.

Dockerfile
sengi: Thanks!

PR #1911, opened 2022-08-16
Rewrite redirect and home URIs for OAuth Applications

theseanything:
This adds the ability to rewrite parts of the redirect and home URIs for OAuth Applications. This allows us to run Signon in our new infrastructure alongside the existing instance of Signon and redirect users to the correct domains for applications running in the new infrastructure. This will remove the need to create separate OAuth Applications for instances of applications in the new infrastructure.

You can use the following env vars to configure the URI string substitution:
SIGNON_APPS_URI_SUB_PATTERN: the pattern within URI to match and rewrite (e.g. "integration.publishing.service.gov.uk")
SIGNON_APPS_URI_SUB_REPLACEMENT: the string to replace the matching pattern (e.g. "integration.eks.govuk.digital")

### DIFF
--- a/app/models/doorkeeper/application.rb
+++ b/app/models/doorkeeper/application.rb
@@ -56,7 +56,26 @@ class ::Doorkeeper::Application < ActiveRecord::Base
     ].include?(name)
   end
 
+  def redirect_uri
+    substituted_uri(self[:redirect_uri])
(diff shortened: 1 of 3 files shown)


sengi (approved):
Nice!

PR #1946, opened 2022-09-13
Fix `devise_zxcvbn` gem to version >=1.1, <2

brucebolt:
We are using this gem to determine whether a password is considered weak, according to the rules implemented in the gem at the time it was first used in signon.

In later releases, these rules have changed (e.g. commonly used passwords are not permitted). We cannot upgrade the gem as the password rules will change but won't have been communicated to users.

Fixing to this specific version will mean we retain the existing rules.

This will allow us to close https://github.com/alphagov/signon/pull/1924.


callumknights (approved):
This seems like a fair compromise to me 👍

formigarafa:
you only need to check for password strength on password change, though.
You could still let the old users still use their old password on the login without any issues.

Now, if their previous password were considered weak it is probably because it is. 
A newer version could help new users creating more of those and old users updating to better ones.

You could also figure if an old password is weak by testing their password on login. At this point something could be done to help them to keep themselves safe.

Please, make sure changes on zxcvbn and zxcvbn-ruby were double checked before accepting those compromises.

PR #1953, opened 2022-09-20
Add rake tasks to sync OAuth Secrets as Kubernetes Secrets

theseanything:
This adds the `kubernetes:sync_token_secrets[<config_map_name>]` rake task, which will retrieve a config map from Kubernetes which lists the emails for ApiUsers who's authorisation tokens are to be synchronised with Kubernetes. The task will then create or update Kubernetes Secrets for every token associated for that ApiUser. 

Also adds the `kubernetes:sync_app_secrets[<config_map_name>]` rake task, which will retrieve a config map from Kubernetes which lists the names for OAuthApps who's client ID and Secret are to be synchronised with Kubernetes. 

We'll use a cron job/workflow to trigger this task on a regular interval to keep the secrets up to date. This intended to replace signon-resources.


lib/kubernetes/client.rb
```diff
@@ -0,0 +1,51 @@
+require "kubeclient"
+
+module Kubernetes
+  class Client
+    API_VERSION = "v1".freeze
+    API_SERVER = "https://kubernetes.default.svc".freeze
+    SERVICE_ACCOUNT = "/var/run/secrets/kubernetes.io/serviceaccount".freeze
+
+    BEARER_TOKEN_FILE = "#{SERVICE_ACCOUNT}/token".freeze
```
sihugh, line 9: Are these actual files on the filesystem? Are they there already?

lib/tasks/sync_kubernetes_secrets.rake
```diff
@@ -0,0 +1,37 @@
+namespace :kubernetes do
+  desc "Synchronise OAuth Token secrets in Kubernetes"
+  task :sync_token_secrets, %i[config_map_name environment_name] => :environment do |_, args|
```
theseanything: remove environment_name

test/factories/users.rb
```diff
@@ -102,3 +102,14 @@
     end
   end
 end
+
+def api_user_with_token(name, token_count: 2)
+  FactoryBot.create(:api_user, name: name) do |api_user|
+    app = FactoryBot.create(:application)
```
theseanything, line 108: Create separate app per token, so token name changes.

test/lib/tasks/sync_kubernetes_secrets_test.rb
```diff
@@ -0,0 +1,89 @@
...
+          config_map_name: "config_map_name",
+          environment_name: "doesnotexist",
+        })
+      end
+
+      assert_match(/doesnotexist/, err.message)
+    end
+  end
+
+  def expect_secret_tokens_created_for_users(client, users)
```
theseanything, line 93: Name change to make it obviously that expects token created ONLY for users specified.

test/lib/tasks/sync_kubernetes_secrets_test.rb
```diff
@@ -0,0 +1,89 @@
...
+        client.expects(:apply_secret).with(
+          "signon-token-#{user.name}-#{authorisation.application.name}".parameterize,
+          { bearer_token: authorisation.token },
+        ).once
+      end
+    end
+  end
+
+  def stub_user_config_map(client, emails)
+    email_list = emails.map { |e| "\"#{e}\"" }.join(",")
```
theseanything, line 114: Nit:
`email_list = emails.map { |e| %Q("#{e}") }.join(",")`

lib/kubernetes/client.rb
theseanything: Ah sorry I missed your comment earlier - yeah, they are mounted by Kubernetes into the container when you define that the pod uses a ServiceAccount.

lib/tasks/sync_kubernetes_secrets.rake
```diff
@@ -0,0 +1,57 @@
...
+  desc "Synchronise OAuth Token secrets in Kubernetes"
+  task :sync_token_secrets, %i[config_map_name] => :environment do |_, args|
+    client = Kubernetes::Client.new
+    config_map = client.get_config_map(args[:config_map_name])
+
+    emails = JSON.parse(config_map.data["api_user_emails"])
+    missing_users = []
+
+    emails.each do |email|
+      api_user = ApiUser.find_by(email: email)
```
robinjam, line 11: Suggest hoisting this query out of the loop (e.g. https://gist.github.com/robinjam/5279a5cfe3345823e46857510b961fed)

robinjam (changes requested):

theseanything:
@robinjam Thanks again for the review - implemented all the changes you've suggested!

robinjam (approved):
LGTM 👍

PR #1959, opened 2022-09-26
Increase test coverage

BeckaL:
Adding tests for a couple of classes/modules that weren't tested. Part of increasing test coverage ahead of switching this application to CD.

This gets us from ~94% coverage to 96.77% 🥳 

[Trello](https://trello.com/c/jBtHWHBh/14-enable-continuous-deployment-for-signon)

### DIFF
--- a/test/signin_permission_granter_test.rb
+++ b/test/signin_permission_granter_test.rb
@@ -1,0 +1,31 @@
+require "test_helper"
+
+class SigninPermissionGranterTest < ActiveSupport::TestCase
+  setup do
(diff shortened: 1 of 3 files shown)


test/signin_permission_granter_test.rb
```diff
@@ -0,0 +1 @@
+
```
JonathanHallam, line 2: assume this line was a mistake/generated thing

JonathanHallam (approved):
Nice one, thank you :)

PR #1975, opened 2022-10-07
Fix ArgumentError in Kubernetes::Client.new

sengi:
This broke with the Ruby 2 to 3 upgrade and we didn't notice because even though the broken constructor was covered by a test, the test still passes because the assertion is too permissive.

(This PR only fixes the implementation, not the test — I ran out of time trying to figure out a way to do that without rewriting the whole approach. The test still passes either way 😞)

https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

https://testing.googleblog.com/2013/05/testing-on-toilet-dont-overuse-mocks.html

### DIFF
--- a/lib/kubernetes/client.rb
+++ b/lib/kubernetes/client.rb
@@ -16,12 +16,8 @@ module Kubernetes
 
       @client = Kubeclient::Client.new(
         API_SERVER, API_VERSION,
-        {
-          auth_options: {
-            bearer_token_file: BEARER_TOKEN_FILE,
-          },
-          ssl_options:,
-        }
+        auth_options: { bearer_token_file: BEARER_TOKEN_FILE },
+        ssl_options:
       )
     end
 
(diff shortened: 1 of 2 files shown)


theseanything (approved):

sengi:
Do shout if this isn't the right fix, or if there's a simple way to make the test meaningful.

sengi:
I think we can actually ditch `Kubeclient` and use the [official one](https://github.com/kubernetes-client/ruby/) instead, which would eliminate most of what our `Kubernetes::Client` wrapper class is doing. But it was late and I'd already spent too long looking at this 😅

theseanything:
Yeah, wasn't super happy with these tests - originally tried asserting that the actually web request gets made to the Kubernetes API. But there's limitation with the testing framework with asserting requests in Rake tasks 😔. Ended up failing back to mocking the client, trying to not spend too much time on it vs getting something working.

PR #1976, opened 2022-10-08
Add build/test instructions to README + minor improvements.

sengi:
Improve instructions for first-time contributors and bring the document more into line with the [GDS Way] and [readme template].

- List the actual commands necessary to run the testsuite under GOV.UK Docker.
- Add copy/pasteable instructions for running the test suite without GOV.UK Docker. This lowers the barrier to entry for potential contributors (including me!), because setting up GOV.UK Docker requires a high-end machine, good Internet connectivity and significant time.
- Delete the _technical documentation_ heading, since the whole page is technical documentation and that heading wasn't adding anything.
- Update links to 3rd-party gems' websites.
- en-GB spelling for _licence_ (noun).
- Move the reference-style links underneath the sections where they're used. This makes it easier to read the document in plain text form.

[GDS Way]: https://gds-way.cloudapps.digital/manuals/readme-guidance.html
[readme template]: https://github.com/alphagov/readme-template/blob/main/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,40 +1,79 @@
 # Signon
 
-Signon is a centralised OAuth2 based single sign-on provider for GDS services that provides username/password and 2-Factor authentication.
+Signon is a centralised, OAuth2-based single sign-on (SSO) provider for GDS services that provides username/password and 2-factor authentication.
 
-## Technical documentation
+Signon uses [Devise] for username/password sign-in, and [Doorkeeper] as an OAuth 2 provider. Details of our interpretation of OAuth are provided in [an accompanying document](docs/oauth.md).
 
-[Devise] is used to provide username password sign-in, and [Doorkeeper] as an OAuth 2 provider. Details of our interpretation of OAuth are provided in [an accompanying document][auth].
+Signon is a Ruby on Rails app and should follow [our Rails app conventions][conventions].
 
-This is a Ruby on Rails app, and should follow [our Rails app conventions][conventions].
+[Devise]: https://github.com/heartcombo/devise
+[Doorkeeper]: https://github.com/doorkeeper-gem/doorkeeper
+[conventions]: https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html
 
-You can use the [GOV.UK Docker environment][govuk-docker] to run the application and its tests with all the necessary dependencies. Follow the [usage instructions][docker-usage] to get started.
 
-**Use GOV.UK Docker to run any commands that follow.**
+## Running the test suite using GOV.UK Docker
 
-### Running the test suite
+You can use the [GOV.UK Docker environment][govuk-docker] to run the application, its tests and all its dependencies. Follow the [govuk-docker installation instructions][govuk-docker-install] to get started.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,40 +1,79 @@
 # Signon
 
-Signon is a centralised OAuth2 based single sign-on provider for GDS services that provides username/password and 2-Factor authentication.
+Signon is a centralised, OAuth2-based single sign-on (SSO) provider for GDS services that provides username/password and 2-factor authentication.
 
-## Technical documentation
```
kevindew, line 5: This is part of the convention in: https://docs.publishing.service.gov.uk/manual/readmes.html your point is right that this is all technical documentation to a degree but I don't think this is worth breaking the convention for.

README.md
```diff
@@ -1,40 +1,79 @@
...
+[Doorkeeper]: https://github.com/doorkeeper-gem/doorkeeper
+[conventions]: https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html
+
+
+## Running the test suite using GOV.UK Docker
+
+You can use the [GOV.UK Docker environment][govuk-docker] to run the application, its tests and all its dependencies. Follow the [govuk-docker installation instructions][govuk-docker-install] to get started.
 
-You can use the [GOV.UK Docker environment][govuk-docker] to run the application and its tests with all the necessary dependencies. Follow the [usage instructions][docker-usage] to get started.
+The first build with GOV.UK Docker typically takes 5-20 minutes, depending on your computer and network connection.
```
kevindew, line 18: This information seems like it should be in the GOV.UK docker documentation rather than in here

README.md
```diff
@@ -1,40 +1,79 @@
...
+```sh
+bundle exec rake test TEST=test/lib/kubernetes/client_test.rb TESTOPTS='-n "/should call apply secret/"'
+```
+
+When finished, you may want to shut down the MySQL and Redis servers.
+
+```sh
+brew services stop redis
+brew services stop mysql
+```
```
kevindew, line 68: We shouldn't include this as it has a high risk of rotting - the wave of work on READMEs to bring them consistent involved lots and lots of removals of out of date information. 

It also implies that these steps are supported but the only dev environment that is supported is the govuk-docker one (and that's only barely) not aspects of peoples brew set-ups. There's potentially all sorts of little problems that can occur: MySQL/Redis versions that are different to actual usage causing surprises, Ruby dependencies like rbenv, the frequent [mysql build issue](https://github.com/brianmario/mysql2/issues/1175) that need extra flags.

So I think it's better to think of running at on your local machine as something that is fine but unsupported. Generally if people are comfortable with the challenges of running on their local machine they're also sufficiently knowledgable that they don't need the extra instructions.

kevindew (commented):
Thanks for your work on this @sengi - it does seem to drift a bit from the conventions though in: https://docs.publishing.service.gov.uk/manual/readmes.html - they of course can always be changed but it took [a lot of work to get them consistent last time](https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh) so would advise being delicate.

Sorry to be a bit of a pain.

sengi:
That's fine, no probs. Thanks for asking :)

PR #1982, opened 2022-10-13
Log the names of the k8s secrets in the app/token sync tasks.

sengi:
This makes it easier to figure out what's wrong when a name doesn't match.

This PR shouldn't affect anything outside of Kubernetes.

Tested: ran in staging with a locally-built image; produces expected log output.


theseanything (approved):

sengi:
The image build is failing on Jenkins for some reason yet I can't repro locally 🤷 #1983 fixes it so I'll just get that merged first.

PR #1983, opened 2022-10-14
Clean up Dockerfile and use govuk-ruby-base.

sengi:
Tested: builds and runs locally:

    docker build -t signon .
    docker run -it --rm -eDEVISE_PEPPER=x -eDEVISE_SECRET_KEY=x signon


sengi:
Fixed up this one cos it was easier than trying to figure out why the old Dockerfile wasn't building on Jenkins 😅

marcpomfret (approved):

PR #2074, opened 2023-01-24
Allow all levels of admin to reset 2fa 

BeckaL:
Currently, only super admins can reset 2sv or ask users to set it up. This grants the ability to reset / set up 2sv for other levels of admin (super_org_admin, organisation_admin, admin) for any user they have permissions to edit.

These are:
* admins -> anyone with equal or lower roles (so not super admins)
* super org admins -> anyone with an equal or lower role (so not admins or super admins) in the same organisation or child organisation
* org admins -> anyone with an equal or lower role (not admins, super admins, super org admins) in the same organisation (but not child oragnisation)

This is part of work to make 2fa mandatory for everyone, and will make it easier for individuals within organisations to manage 2sv for people within their organisations.

[Trello](https://trello.com/c/2V2skE0x/431-allow-users-with-role-organisation-admin-or-organisation-super-admin-to-reset-2fa-on-the-users-they-can-manage)


app/policies/user_policy.rb
```diff
@@ -44,6 +33,21 @@ def cancel_email_change?
     allow_self_only || edit?
   end
 
+  def has_user_management_permissions
```
kevindew, line 36: I think this should be a private method (nitpick too that as it's a question method it should have a question mark suffix)

I'd suggest renaming this as "permissions" is a domain concept.

I actually wonder whether we should leave this as the edit? method and then alias flag_2sv? and reset_2sv? to edit? ?

app/views/users/_form_fields.html.erb
```diff
@@ -25,7 +25,7 @@
   </p>
 <% end %>
 
-<% if policy(User).assign_role? %>
+<% if policy(current_user).assign_role? %>
```
kevindew, line 28: I'm not sure I understand why this and the previously version was changed? was it previously a buggy implementation?

test/integration/reset_two_step_verification_test.rb
```diff
@@ -0,0 +1,109 @@
+require "test_helper"
+
+class ResetTwoStepVerificationTest < ActionDispatch::IntegrationTest
```
kevindew, line 3: Should we also be testing that users can setup 2sv? I think from the change above it looks that it is also an affect of this change

kevindew (commented):
Looks great Becka. I've deployed it on Integration and it all seems to work as expected. I've just made some method organisation suggestions and a query about tests

app/views/users/_form_fields.html.erb
BeckaL: Yup good point, will switch it back since this is unrelated to the changes here

test/integration/reset_two_step_verification_test.rb
BeckaL: Good point @kevindew, are you thinking of e.g. [this test](https://github.com/alphagov/signon/blob/main/test/integration/super_admin_flagging_two_step_verification_test.rb)

kevindew: I'd wonder if it could be part of this test somehow. Perhaps it could go to each user ask them to set-up 2sv and then reset it to do both actions?

BeckaL: Cool, I'll have a look. Good suggestion too - while testing this it turns out that other admin users _cannot_ set 2sv on users, so good shout

BeckaL: Have updated - I didn't end up putting it in the same test because it would have become a more involved test requiring us to log in as the user to set up 2sv in the middle, before we could reset it. But yes, adding the tests uncovered that other kinds of admin users couldn't actually update the 2sv requirement, so have definitely caught a bug that would have at some point been raised on zendesk.

test/integration/flagging_two_step_verification_test.rb
```diff
@@ -0,0 +1,111 @@
...
+      uncheck "Ask user to set up 2-step verification"
+      click_button "Update User"
+
+      assert_not last_email
+    end
+
+    assert_not user.reload.require_2sv
+  end
+
+  context "updating a user" do
```
kevindew, line 40: I don't think we need this context, as it seems to be for all of the tests.

In general, at least [in RSpec parlance](https://rspec.rubystyle.guide/#context-cases), a context should always start with a "when or a with" (our RSpec validator enforces that).

test/integration/flagging_two_step_verification_test.rb
```diff
@@ -0,0 +1,111 @@
...
+    setup do
+      @organisation = create(:organisation)
+      @child_organisation = create(:organisation, parent: @organisation)
+      @user = create(:user, organisation: @organisation)
+      @user_in_child_organisation = create(:user, organisation: @child_organisation)
+    end
+    context "when logged in as a super admin" do
```
kevindew, line 47:
```suggestion
    end
    
    context "when logged in as a super admin" do
```

nit

test/integration/flagging_two_step_verification_test.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      should "be able to send a notification to a user to set up 2fa" do
+        assert_admin_can_send_2fa_email(@org_admin, @user)
+      end
+
+      should "be able to unset the requirement for 2fa" do
+        assert_admin_can_remove_2sv_requirement_without_notifying_user(@org_admin, @user)
+      end
+    end
+  end
```
kevindew, line 110: Do we want a case for when you're not an admin and can't do these actions?

kevindew (commented):
Looks great, glad you caught the bug. Just a couple of minor bits.

kevindew (approved):
Super 👍

PR #2075, opened 2023-01-24
Configure GovukContentSecurityPolicy for govuk_app_config changes

kevindew:
Trello: https://trello.com/c/lxxx5XLZ/178-govuk-has-a-half-implemented-content-security-policy-csp

This configures the content security policy in preparation for the
breaking changes coming from
https://github.com/alphagov/govuk_app_config/pull/279. As this app uses
govuk_admin_template and that uses jQuery 1.x and inline script tags,
then this app needs unsafe_inline for script and the nonce generator
disabled.

As an aside, It was a surprise that this application had configured the
GovukContentSecurityPolicy as this had been initially done just in
Frontend apps and it looks like this made it through in some outsourced
Rails updates [1]. I'm leaving this config in so there is an example of
an app outside of frontend using it to build on and as a case study in
configuring an app.

[1]: https://github.com/alphagov/signon/commit/45a5a51abe3fb07f885c17aec2b7b271e1874e5b

---


config/initializers/content_security_policy.rb
```diff
@@ -1 +1,12 @@
-GovukContentSecurityPolicy.configure
+GovukContentSecurityPolicy.configure do |policy|
+  # Ensures the ability to use inline JavaScript without protections. This is
+  # required for compatibility with govuk_admin_template which both uses script
+  # tags without nonces and uses jQuery 1.x which requires unsafe-inline in
+  # some browsers (Firefox is one)
+  script_policy_with_unsafe_inline = (policy.script_src + ["'unsafe-inline'"]).uniq
```
BeckaL, line 6: I'm assuming the double quotes here is intended and not a typo, but just checking?

kevindew: Yup, thanks. They're intentional - we need unsafe-inline in quotes.

kevindew:
Thanks for the review @BeckaL - yup I've had to learn this stuff quite quickly, it'll be a whole safe new future!

PR #2091, opened 2023-01-31
Remove test for doorkeeper version

brucebolt:
We currently have a test to ensure doorkeeper matches a hardcoded version. This value has to be updated each time the doorkeeper gem is upgraded.

We understand the reason being that a controller was based on (and inherits from) another controller in doorkeeper.

The expectation was developers would compare the controller in this app to the equivalent in doorkeeper before making any updates.

We know from anecdotal (and commit history) that this has not happened. It therefore feels this test is not achieving its original purpose.

We know the test coverage of this controller is good, so are removing the version check test and instead relying on the controller's tests to see if there have been any breaking changes in doorkeeper.


BeckaL (approved):
👍

PR #2083, opened 2023-01-26
Allow gds admins to exempt users from 2sv with reason

BeckaL:
This PR introduces the basics for exempting users from 2sv, adding a link on the 'edit user' page and a new page for adding and editing the exemption reason. 

The functionality is behind a feature flag, so users will not be able to exempt unless the environment has a "PERMIT_2SV_EXEMPTION". We plan to remove this feature flag once the other 2sv features are ready.

Screenshots of new functionality added:

| Scenario | Screenshot |
|----|-----|
| Link to add an exemption for a user |Screenshot 2023-02-03 at 08 09 04|
| Link to edit an exemption for a user|Screenshot 2023-02-03 at 08 07 47|
| Exemption reason displaying but no link to edit when logged in as a non-gds admin |Screenshot 2023-02-03 at 08 10 04|
|Create / edit an exemption reason interface|Screenshot 2023-02-03 at 09 47 32|

### Summary of permissions

| Logged in as user | Editing user | Can exempt? | Can mandate 2sv for an exempted user (remove an exemption)? |
|---|---|---|---|
| Gds superadmin or admin | Normal user in any organisation | Yes | Yes |
| Gds admin | Admin (any level) user in any organisation | No | n/a (can't be exempted) |
| Gds org admin | Normal gds user | No | Yes |
| Non-gds admin | Normal user in their organisation | No | Yes |

Note that any user who can edit a user with an exemption reason will be able to see the exemption reason (but not edit it).
As a reminder, the following users have the following permissions to edit users:

* Super admins - any users at any level
* Admins - any users at same or lower level below in the hierarchy (so admins, super-org admins, org admins, normal users)
* Super org admins - users in their org or a child org at same or lower level below in the hierarchy (so super-org admins, org admins, normal users)
* Org admins - users in their org or a child org at same or lower level below in the hierarchy (so org admins, normal users)
* Normal users - no edit permissions

### What's not included in this PR

Ensuring that exempted users do not get unnecessary prompts to set up 2sv - that to be covered in [this ticket](https://trello.com/c/v5jGPVxF/447-ensure-user-experience-for-2fa-exempted-users-is-consistent)

[Trello](https://trello.com/c/qijNj4uh/433-allow-admins-super-admins-gds-only-to-un-set-the-requires-2sv-flag-on-any-user-if-they-provide-a-reason-free-text-similar-to-the)

[...]

PR #2095, opened 2023-02-03
Prevent api users from having 2sv required or exemption reasons set

BeckaL:
Depends on https://github.com/alphagov/signon/pull/2083 [merged]

Signon has two types of users - human users, and 'api users'. An api user is created in the case where we want communication between two applications to be authenticated, so an api user is created for one of the applications, granting it access to another via a bearer token. Api users can be for applications internal to GDS and external.

Because we don't want to enforce 2sv on these users, this contains:

* some preventative work ensuring that api users cannot have the 'require_2sv' flag set to true. This should give us a layer of safety when setting this flag across big groups of users, when we turn 2sv on for all human users.
* Some work to prevent users from being able to give api users exemption reasons in the UI - since these users cannot have 2sv set on them, setting / editing an exemption doesn't make sense.

[Trello](https://trello.com/c/ezvggxS1/450-ensure-api-users-cannot-be-exempted-and-cannot-have-2sv-required)

### DIFF
--- a/app/models/api_user.rb
+++ b/app/models/api_user.rb
@@ -1,6 +1,18 @@
 class ApiUser < User
   default_scope { where(api_user: true).order(:name) }
+  validate :reason_for_2sv_exemption_blank
+  validate :require_2sv_is_false
 
   DEFAULT_TOKEN_LIFE = 2.years.to_i
   AUTOROTATABLE_TOKEN_LIFE = 60.days.to_i
+
+private
+
+  def require_2sv_is_false
+    errors.add(:require_2sv, "can't be true for api user") if require_2sv
+  end
+
+  def reason_for_2sv_exemption_blank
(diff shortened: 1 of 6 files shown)


app/models/api_user.rb
```diff
@@ -1,6 +1,22 @@
...
 
   DEFAULT_TOKEN_LIFE = 2.years.to_i
   AUTOROTATABLE_TOKEN_LIFE = 60.days.to_i
+
+private
+
+  def require_2sv_is_false
+    if require_2sv
+      errors.add(:require_2sv, "can't be true for api user")
+    end
```
brucebolt, line 14: I wonder if these would be better written on a single line, e.g. `errors.add(:require_2sv, "can't be true for api user") if require_2sv`?

brucebolt (approved):
All looks good to me 👍 .

Just a minor non-blocking comment.

PR #2098, opened 2023-02-06
Make exempted 2sv user experience consistent with their status

BeckaL:
In a [previous PR](https://github.com/alphagov/signon/pull/2095), we introduced the ability to exempt a user from 2sv, based on providing an exemption reason.

This PR ensures that users who are exempted (i.e. those with exemption reasons) have an experience that is consistent with this status. This means:

* They are excluded from any rake tasks to bulk set the `require_2sv' flag on groups of users
* They do not see large call to actions to set up 2sv, although they are still able to set this up in the UI if their circumstances change
* When they set up 2sv, their exemption reason is removed, meaning that they no longer appear in the list of exempted users.

Screenshot to illustrate the difference when logging in as an exempted user:

| Before this change | After this change |
| ---- | ---- |
|Screenshot 2023-02-06 at 11 19 02|Screenshot 2023-02-06 at 11 17 48|

[Trello](https://trello.com/c/v5jGPVxF/447-ensure-user-experience-for-2fa-exempted-users-is-consistent)

### DIFF
--- a/app/views/root/index.html.erb
+++ b/app/views/root/index.html.erb
@@ -3,7 +3,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
 
-    <% unless current_user.has_2sv? %>
+    <% unless current_user.has_2sv? || current_user.reason_for_2sv_exemption.present? %>
       <%= render "govuk_publishing_components/components/notice", {
         title: "Make your account more secure",
         description_text: (render "devise/two_step_verification/make_your_account_more_secure")
--- a/app/views/root/index.html.erb
+++ b/app/views/root/index.html.erb
@@ -3,7 +3,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
 
-    <% unless current_user.has_2sv? || current_user.reason_for_2sv_exemption.present? %>
+    <% unless current_user.has_2sv? || current_user.exempt_from_2sv? %>
       <%= render "govuk_publishing_components/components/notice", {
         title: "Make your account more secure",
         description_text: (render "devise/two_step_verification/make_your_account_more_secure")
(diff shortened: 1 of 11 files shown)


app/views/root/index.html.erb
```diff
@@ -3,7 +3,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
 
-    <% unless current_user.has_2sv? %>
+    <% unless current_user.has_2sv? || current_user.reason_for_2sv_exemption.present? %>
```
brucebolt, line 6: Not really related directly to this PR, but I wonder if we need an `exempt_from_2sv?` method on the user model, so we don't need to repeat `reason_for_2sv_exemption.present?` in so many places?  It might also be helpful to have this method if we implement exemption expiry times, as the expiry can be checked in the same method.

lib/tasks/users.rake
```diff
@@ -156,6 +156,7 @@ namespace :users do
     raise "Couldn't find organisation: '#{args.org}'" unless organisation
 
     users_to_update = User.where(organisation_id: organisation.id, require_2sv: false)
+                          .where("reason_for_2sv_exemption IS NULL")
```
brucebolt, line 159: This can also be written as `where(reason_for_2sv_exemption: nil)`.  Either work, but the `nil` is more Rails-y.

brucebolt (approved):
Looks good to me.  I've just made a couple of really small suggestions.

lib/tasks/users.rake
BeckaL: Oh good shout, I'd copied this is not when I was checking for not null, will amend

app/views/root/index.html.erb
BeckaL: 👍 added this in - might be of use in your PR if you rebase and end up needing to add this in to the logic

PR #2104, opened 2023-02-09
Add expiry date to 2sv exemptions

BeckaL:
This PR implements adding an expiry date to the 2sv exemptions form. Exemptions must now be created with expiry dates, reflecting the thought that some exemption reasons will be temporary and should be reviewed at some point.

This PR:
* Adds the exemption expiry date to the user interface and handles in the backend code
* Adds the expiry date to access logs when an exemption is created or updated
* Adds the expiry date to the information displayed about users on the users index page, and in the csv export from that page
* Updates some copy on the existing form as requested by Indy.

Screenshots:

| | Currently | Edit exemption form in this PR |
|----|----|----|
| Edit exemption form |Screenshot 2023-02-10 at 14 40 43|Screenshot 2023-02-10 at 14 29 54|
| Account access log when exemption is created |Screenshot 2023-02-10 at 14 41 37|Screenshot 2023-02-10 at 14 30 59|
| Users index page |Screenshot 2023-02-10 at 14 38 05| Screenshot 2023-02-10 at 14 31 12|

### What's not in this PR 

At least for the MVP, we're not planning on implementing any behaviour around expiry dates. For example, once someone's 2sv exemption expiry date has passed, they will still be able to log in without setting up 2sv, and nobody will be automatically alerted. Currently, the plan is to manage exemption expiries in an offline process, given the small number we expect.

[Trello](https://trello.com/c/Eelrmilj/456-implement-exemption-expiry-dates-for-2sv)

### DIFF
--- a/app/controllers/two_step_verification_exemptions_controller.rb
+++ b/app/controllers/two_step_verification_exemptions_controller.rb
@@ -4,31 +4,45 @@ class TwoStepVerificationExemptionsController < ApplicationController
   def edit; end
 
   def update
-    if params[:user][:reason_for_2sv_exemption].empty?
-      flash[:alert] = "Reason for exemption must be provided"
-
+    error_string_or_parsed_params = parsed_params_or_error_string(params[:user])
+    case error_string_or_parsed_params
+    when String
+      flash[:alert] = error_string_or_parsed_params
       redirect_to edit_two_step_verification_exemption_path(@user)
-    elsif params[:user]["expiry_date_for_2sv_exemption(1i)"]
-      expiry_date = parse_date_from_user_params(params[:user])
-      if Time.zone.today >= expiry_date
-        flash[:alert] = "Expiry date must be in the future"
-
-        redirect_to edit_two_step_verification_exemption_path(@user)
-      else
-        @user.exempt_from_2sv(params[:user][:reason_for_2sv_exemption], current_user, expiry_date)
-
-        flash[:notice] = "User exempted from 2SV"
-
-        redirect_to edit_user_path(@user)
-      end
+    else
+      reason_for_2sv_exemption, valid_expiry_date = error_string_or_parsed_params
+      @user.exempt_from_2sv(reason_for_2sv_exemption, current_user, valid_expiry_date)
+      flash[:notice] = "User exempted from 2SV"
+      redirect_to edit_user_path(@user)
     end
   end
 
 private
 
+  def parsed_params_or_error_string(user_params)
+    if user_params[:reason_for_2sv_exemption].empty?
+      "Reason for exemption must be provided"
+    else
+      begin
+        expiry_date = parse_date_from_user_params(params[:user])
+        if Time.zone.today >= expiry_date
+          "Expiry date must be in the future"
+        else
+          [user_params[:reason_for_2sv_exemption], expiry_date]
+        end
+      rescue Date::Error
+        "Expiry date is not a valid date"
+      end
+    end
+  end
+
   def parse_date_from_user_params(user_params)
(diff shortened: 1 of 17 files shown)


app/controllers/two_step_verification_exemptions_controller.rb
```diff
@@ -19,6 +22,13 @@ def update
 
 private
 
+  def parse_date_from_user_params(user_params)
+    year = user_params["expiry_date_for_2sv_exemption(1i)"].to_i
```
JonathanHallam, line 40: What format is the date being passed through in? I'd expect to be able to do something like DateTime.parse(date) assuming it's in a consistent string

test/integration/managing_two_step_verification_test.rb
```diff
@@ -168,24 +168,25 @@ class ManagingTwoStepVerificationTest < ActionDispatch::IntegrationTest
...
+        should "be able to exempt a user requiring 2sv from 2sv" do
           user_requiring_2sv = create(:two_step_mandated_user, organisation: @organisation)
 
-          user_can_be_exempted_from_2sv(@super_admin, user_requiring_2sv, @reason_for_exemption)
+          user_can_be_exempted_from_2sv(@super_admin, user_requiring_2sv, @reason_for_exemption, @expiry_date)
           assert_user_access_log_contains_messages(user_requiring_2sv, ["Exempted from 2-step verification by #{@super_admin.name} for reason: #{@reason_for_exemption}"])
         end
 
-        should "be able to see a link to exempt a user who does not yet require 2sv but is not exempt" do
+        should "be able to see exempt a user who does not yet require 2sv but is not exempt" do
```
JonathanHallam, line 185: errant see

test/integration/managing_two_step_verification_test.rb
```diff
@@ -197,16 +198,15 @@ def exemption_message(initiator_name, reason)
         end
 
         context "when a exemption reason already exists" do
-          should "be able to see a link to edit exemption reason" do
-            user_requiring_2sv = create(:user, organisation: @organisation, reason_for_2sv_exemption: "user is exempt")
+          should "be able to see a to edit exemption reason and date" do
```
JonathanHallam, line 212: `be able to edit exemption reason and date`

app/controllers/two_step_verification_exemptions_controller.rb
BeckaL: So because the date comes from three form fields (as is given by the date_select in rails), it comes through in the shape of a year string, month string and day string. I _think_ DateTime.parse wouldn't be useful in this case since we'd have to artificially make it into a date time string,  but let me know if you can see a better way (also might be worth noting we're not using date times, but just dates)

test/integration/managing_two_step_verification_test.rb
BeckaL: Good spot, have fixed

test/integration/managing_two_step_verification_test.rb
BeckaL: 👍  have fixed

app/controllers/two_step_verification_exemptions_controller.rb
BeckaL: Had another think and have now changed this more along your lines to:
```    
date_part_params = ["expiry_date_for_2sv_exemption(1i)", "expiry_date_for_2sv_exemption(2i)", "expiry_date_for_2sv_exemption(3i)"]
    date_string = date_part_params.map { |param_name| user_params[param_name] }.join("/")
    Date.parse(date_string)
```

This also has the benefit of meaning that if one of the parts is ever not an integer (which should never happen in practice), it'll be caught by our existing rescue from an invalid date. `.to_i` is a bit sketchy.

JonathanHallam (approved):
Nice one, thank you for this :)

PR #2105, opened 2023-02-09
Require 2SV for users from organisations where 2SV is mandatory

brucebolt:
This implements the following:

- requires 2SV for all existing users (who are not exempt) from an organisation when their organisation is set to require 2SV
- requires 2SV for all new users from an organisation when their organisation is set to require 2SV (e.g. for users created by rake tasks, or in bulk)
- moves the 2SV checkbox during registration to a separate page, so it is only presented to users who are not required to have 2SV (either by membership of an organisation or because they are an admin/superadmin role)

Validations are still carried out when the first form is submitted, hence why we now need to explicitly check the submission is valid without persisted the user in the database.

The following user flows are now present during registration:

1. New user who is being assigned to an organisation that has mandatory 2SV or has admin/superadmin status:

Screenshot showing registration form for a new user belonging to a 2SV mandated organisation

Screen showing flash message of An invitation email has been sent to test11@gov.uk.

2. New user who is being assigned to an organisation that does not have mandatory 2SV and is not admin/superadmin:

Screenshot showing registration form for a user being assigned to an organisation without mandatory 2SV

Screenshot showing new intermediate page with 2SV question and flash message of An invitation email has been sent to test2@gov.uk.

[Trello card](https://trello.com/c/a3Zu69Td)

### DIFF
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -131,6 +131,6 @@ private
   end
 
   def new_user_requires_2sv(params)
-    params[:organisation_id].present? && Organisation.find(params[:organisation_id]).require_2sv?
+    (params[:organisation_id].present? && Organisation.find(params[:organisation_id]).require_2sv?) || %w[superadmin admin].include?(params[:role])
   end
 end
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -18,14 +18,17 @@ class InvitationsController < Devise::InvitationsController
     if (self.resource = User.find_by(email: params[:user][:email]))
       authorize resource
       flash[:alert] = "User already invited. If you want to, you can click 'Resend signup email'."
-      respond_with resource, location: after_invite_path_for(resource)
+      respond_with resource, location: users_path
     else
       # workaround for invitatable not providing a build_invitation which could be authorised before saving
-      user = User.new(resource_params)
-      user.organisation_id = resource_params[:organisation_id]
+      all_params = resource_params
+      all_params[:require_2sv] = new_user_requires_2sv(all_params.symbolize_keys)
+
+      user = User.new(all_params)
+      user.organisation_id = all_params[:organisation_id]
       authorize user
 
-      self.resource = resource_class.invite!(resource_params, current_inviter)
+      self.resource = resource_class.invite!(all_params, current_inviter)
       if resource.errors.empty?
         grant_default_permissions(resource)
         set_flash_message :notice, :send_instructions, email: resource.email
@@ -50,7 +53,11 @@ class InvitationsController < Devise::InvitationsController
 private
 
   def after_invite_path_for(_resource)
-    users_path
+    if new_user_requires_2sv(resource)
+      users_path
+    else
+      require_2sv_user_path(resource)
+    end
   end
 
   # TODO: remove this method when we're on a version of devise_invitable which
@@ -121,5 +128,9 @@ private
     SupportedPermission.default.each do |default_permission|
       user.grant_permission(default_permission)
     end
+  end
+
+  def new_user_requires_2sv(params)
+    params[:organisation_id].present? && Organisation.find(params[:organisation_id]).require_2sv?
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -40,6 +40,7 @@ class User < ApplicationRecord
   validate :organisation_admin_belongs_to_organisation
   validate :email_is_ascii_only
   validate :exemption_from_2sv_data_is_complete
+  validate :organisation_has_mandatory_2sv, on: :create
 
   has_many :authorisations, class_name: "Doorkeeper::AccessToken", foreign_key: :resource_owner_id
   has_many :application_permissions, class_name: "UserApplicationPermission", inverse_of: :user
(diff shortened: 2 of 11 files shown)


app/models/user.rb
```diff
@@ -95,6 +95,12 @@ class User < ApplicationRecord
           end
         }
 
+  def require_2sv?
+    return require_2sv unless organisation
+
+    organisation.require_2sv? || require_2sv
```
BeckaL, line 102: I think we should still include whether someone is exempted here as part of the logic (so if (organisation requires AND not exempted) OR user requires 2sv)). 

I think there are potentially some bits of logic that would be unintentionally changed if we didn't include this condition, that won't be caught by tests at the moment since we don't widely use mandated 2sv orgs in test cases - for example, the behaviour of this method https://github.com/alphagov/signon/blob/main/app/models/user.rb#L323-L325

I think even if it didn't affect any existing behaviour, it'd be good protection against future use of this method, since I think I'd find it surprising it didn't take into account exemptions.

app/controllers/invitations_controller.rb
```diff
@@ -9,23 +9,33 @@ class InvitationsController < Devise::InvitationsController
 
   def new
     authorize User
+    session[:invitation_params] = {}
```
BeckaL, line 12: I'm interested in why we need a session cookie here. What would the implications be of saving the user and then updating with require 2sv?

I guess I feel like using a session cookie introduces some extra complexity, I'm not sure if it has some security and/or PII concerns, and might introduce an extra zendesk layer of debugging if a user doesn't allow cookies and runs into some weird behaviour. So I'd rather avoid unless there's a strong reason.

BeckaL (commented):
Nice one! Code in general looks good and neat, and big improvement to introduce this flow on the existing admin / superadmin creation experience.

Just a couple of questions about implementation.

app/controllers/invitations_controller.rb
brucebolt: The main reason for using a session cookie was so we didn't persist any data into the database after the first part of the form is submitted.  By saving to the database, the user would receive their invitation email after the first part of the form is completed.  If they are quick and do this before the second part of the form is submitted, they'll be able to set up their account without 2SV (although they'll be prompted next time they login).

I don't think it's possible to use Signon without having cookies enabled, so disabled cookies won't be an issue.

I'll have a look and see what would happen if I remove the cookie here and write directly to the database after the first step.

BeckaL: Ah got you. Gah, this is kind of why I hate callbacks. Interested to see if you find a way around it.

brucebolt: I've removed the need for cookies in [this commit](https://github.com/alphagov/signon/pull/2105/commits/0eb62f4e6ae890cb34c22da63602c425fb0df582).  I'd be keen to get your views on it.

Although I prefer the solution without cookies from a technical perspective, it means we end up sending users 2 emails (if they have 2SV mandated, but are from a non-2SV organisation): one to set up their account and the other telling them about 2SV.

app/models/user.rb
brucebolt: Sure, that seems a reasonable thing to add. I'll look at this after lunch.

app/controllers/invitations_controller.rb
brucebolt: This is done, I've now removed the use of session cookies here.

app/views/users/require_2sv.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <p class="checkbox">
+    <%= f.label :require_2sv do %>
+      <%= f.check_box :require_2sv %> Mandate 2-step verification for this user <%= "(this will remove their exemption)" if @user.exempt_from_2sv? %>
+    <% end %>
+    <br/>
+    User will be prompted to set up 2-step verification again the next time they sign in.
+  </p>
+
+  <%= f.submit "Update user", :class => 'btn btn-success' %>
+<% end %>
```
BeckaL, line 16: Nitpick: newline missing

test/models/user_test.rb
```diff
@@ -63,6 +63,24 @@ def setup
...
+      user = create(:user, organisation:)
+      assert_not user.require_2sv?
+
+      organisation.update!(require_2sv: true)
+      assert user.require_2sv?
+    end
+
+    should "not require 2SV for the user when it is switched on for their organisation and they are exempt" do
+      organisation = create(:organisation, require_2sv: false)
+      user = create(:user, organisation:, reason_for_2sv_exemption: "User is exempt")
```
BeckaL, line 78: Just a note - if my PR gets merged in before this, this will start failing as it introduces a validation that if a reason is present, an expiry date is also present. There's a factory for an exempted user in that PR. Only flagging to save you a bit of time debugging because I remember it silently failing on create before and that being confusing 🙃

BeckaL (commented):
This looks good to me! Minor thing about a newline. We discussed reviewing this with Indy tomorrow before merging. One thing I do wonder (can discuss in that review) is whether it'd be worth adding some explanatory text when choosing the organisation saying 'if this org does not yet have 2sv mandated, you will be asked to set this up on the next page'. Just from the perspective of people not realising why they're getting different behaviour for setting up different users.

Looks good from a dev perspective, I'll do the official tick post product review tomorrow in case anything significant comes out of that.

app/views/users/require_2sv.html.erb
brucebolt: Good spot.  I've just moved from Atom to VSCode and hadn't enabled the setting to automatically add new-lines at end of file yet.

test/models/user_test.rb
brucebolt: Cool, thanks for the warning.  I'll assume your PR will get merged first, so I'll rebase and update this test once that's done.

brucebolt:
Thanks @BeckaL.  I've rebased against main and I don't think there were any changes needed to this PR from the product review.

BeckaL (approved):
Nice one!

PR #2172, opened 2023-05-23
Migrate API users index page to GOV.UK Design System

floehopper:
Trello: https://trello.com/c/U0fhO3gx

This is part of the ongoing effort to migrate all of the application's user interface to using the GOV.UK Design System.

* We've moved name & email address into separate columns rather than distinguishing them using typography.
* The "Create API user" button is now on the left rather than over on the right. It's not obvious that having it on the right is something that the design system recommends.
* We've retained the use of the `del` element to visually distinguish suspended users since this seems semantically correct and their suspended state is also visible in a separate column anyway.
* We've retained the use of the `abbr` element for application permissions, even though we think it's being abused in order to provide a tooltip. Note that originally this element was also being given a tooltip by JavaScript from Bootstrap; this is no longer the case. Perhaps a `details` element would be more appropriate here?
* We've had to make a few assertions less specific, because the GOV.UK design system table doesn't make it easy to add CSS classes to individual columns.

Note that we've redacted email addresses in the following screenshots:

| Description | Before | After |
|-------------|--------|-------|
| Filter by "Collections" | filter-by-collections before | filter-by-collections after |
| Filter by "Calc" | filter-by-calc before | filter-by-calc after |

We have screenshots of the full page before and after the changes, but we didn't want to include them here because they included some PII. Let us know in the comments if it would be useful to see them.

### DIFF
--- a/app/controllers/api_users_controller.rb
+++ b/app/controllers/api_users_controller.rb
@@ -1,6 +1,8 @@
 class ApiUsersController < ApplicationController
   include UserPermissionsControllerMethods
 
+  layout "admin_layout", only: %w[index]
+
   before_action :authenticate_user!
   before_action :load_and_authorize_api_user, only: %i[edit update]
   helper_method :applications_and_permissions, :visible_applications
--- a/app/helpers/api_users_helper.rb
+++ b/app/helpers/api_users_helper.rb
@@ -2,4 +2,23 @@ module ApiUsersHelper
   def truncate_access_token(token)
     raw "#{token[0..7]}#{'&bull;' * 24}#{token[-8..]}"
   end
+
+  def api_user_name(user)
+    anchor_tag = link_to(user.name, edit_api_user_path(user), class: "govuk-link")
+    user.suspended? ? content_tag(:del, anchor_tag) : anchor_tag
+  end
+
+  def permissions_by_application(user)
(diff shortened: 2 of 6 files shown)


pdagnall:
Feedback from design:
- The table is wider than the page/column
- Can we make the ‘sign-ins’ title in the table one line
- App column in table - I’ve never seen hint text indicated with dots under the wording in design system - could be worth us checking with an interaction designer to see what the design system pattern is and in research to see if users know what the dots mean

Notes for future investigation/research:
- ‘Almost’ 2 years - is this something we use in the design system?
- Could we filter it to only active accounts as default
- Could we make it easier for users to filter by active/suspended

chrislo:
I made the "sign-ins" header unbreakable so it stays on a single line even when the page width is small in [a00917e](https://github.com/alphagov/signon/pull/2172/commits/a00917e8a4ddcbb85f2027a87472b229dd2b69b8).

chrislo:
I spoke to @pdagnall about the other two pieces of feedback and we've agreed to defer any changes to the use of `<abbr>` and using something other than a table to avoid it breaking out of the container for a future iteration.

chrislo (approved):

chrislo:
Rebasing on main before merging.

PR #2174, opened 2023-05-24
Migrate 2SV setup page to GOVUK Design System

chrislo:
Trello: https://trello.com/c/Ht3aHQTT

This PR adopts the GDS design system for the show and update actions on the TwoStepVerificationController. A user sees this page when they add or change the two-step verification code for their account.

The main change to the page is the switch from collapsable sections to having everything visible by default. We considered using the step-by-step, accordion or tabs components/patterns but it's not obvious that these are suitable for forms. Also since there's not a lot of content in each "step", it's not obvious that there's a lot of benefit to the user of hiding some of the "steps". So we've just made the "steps" into numbered headings with non-collapsing content. We don't think this looks too bad. If there's too much content on the page, we could consider splitting the page into multiple pages.

We've made use of the details component to hide the explanation for those users that can't use the barcode. The Design System recommends this component for content that's not relevant to all users which seems appropriate here.

As the design system [recommends][1], we now show an error summary at the top of the page with a link that takes the user to the heading for step 3 as well as highlighting the form field where the error has occurred.

We've retained the QR code as it was previously rendered (using a data uri in an img tag). We noticed that the [qrcode](https://github.com/whomwah/rqrcode) gem supports SVG rendering so we could adopt that as an enhancement (it would make the QR code more responsive to screen width). This would also fix a content security policy warning we're seeing in production [2].

The new version of the page doesn't include any of the custom analytics tracking that was in the original version. We've recently discovered that no events have been recorded in GA for this app since June 2017, so we assume something has inadvertently broken the code. This means it doesn't seem worthwhile trying to reimplement the custom analytics tracking in the new version of the page. Note that we will still see page views for this page as we do for all pages.

The following screenshots capture the main interactions, we can demo the additional states (such as a cancel link being available when the user first sets up 2sv) if it helps to review this. 

[1]: https://design-system.service.gov.uk/components/error-summary/
[2] We see `[Report Only] Refused to load the image...` in production, but the QR code image *does* seem to render (in Chrome at least, we haven't tested it elsewhere). The image doesn't render in development, so changing it would improve the development story.  

# Before

## Step 1 expanded (default)
before_step_1

## Step 2 expanded
before_step_2

## Step 3 expanded
before_step_3

## When incorrect code entered
before_error

# After

## On page load
on_page_load

## User opens additional details
details_expanded

## When incorrect code entered
error

### DIFF
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -1,7 +1,7 @@
 class Devise::TwoStepVerificationController < DeviseController
   before_action -> { authenticate_user!(force: true) }, only: :prompt
   before_action :prepare_and_validate, except: :prompt
-  layout "admin_layout", only: %w[prompt]
+  layout "admin_layout", only: %w[show update prompt]
 
   attr_reader :otp_secret_key
 
--- a/app/views/devise/two_step_verification/show.html.erb
+++ b/app/views/devise/two_step_verification/show.html.erb
@@ -1,97 +1,67 @@
-<% title = current_user.has_2sv? ? "Use a new phone" : "Set up 2-step verification" %>
-<% content_for :title, title %>
-<% invalid_code_entered = flash[:invalid_code] %>
+<% content_for :title, current_user.has_2sv? ? "Use a new phone" : "Set up 2-step verification" %>
 
-<h1><%= title %></h1>
-<div class="row">
-  <div class="col-md-8">
-    <% if current_user.has_2sv? %>
-      <div class="alert alert-warning">
-        <p>Setting up a new phone will replace your existing one. You will only be able to sign in with your new phone.</p>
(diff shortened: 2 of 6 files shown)


pdagnall:
Design feedback:
Looks great! Few suggestions -
- Can we switch the lead text and the warning - so 'make your signon account more secure' comes first, followed by warning
- Can we use the step-by-step component for these pages, it's more like-for-like - https://components.publishing.service.gov.uk/component-guide/step_by_step_nav
- Would be great to set up analytics for the show all/hide all toggle to see how users use this page
- Warning needs to come above the title 'use a new phone' - https://design-system.service.gov.uk/components/error-summary/

floehopper (approved):
This all looks good to me! 👍

chrislo:
Rebasing on main and force-pushing before merging.

PR #2178, opened 2023-05-24
Move two TwoStepVerificationController methods into TwoStepVerificationHelper

floehopper:
### Move `otp_secret_key_uri` method from controller to helper

It's not idiomatic Rails to have a public method on a controller which is not an action. I considered just making the method private, but since it was being called from a bunch of tests I decided to move it to a helper.

While the current_user method is probably accessible from `TwoStepVerificationHelper` methods, I decided it was helpful to make the user an explicit method argument to highlight that the URI is different for different users.

### Move `qr_code_data_uri` method from controller to helper

This helper method is only used in a template (not in the controller), so it makes more sense to move it to the relevant helper.

### DIFF
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -24,16 +24,6 @@ class Devise::TwoStepVerificationController < DeviseController
     end
   end
 
-  def otp_secret_key_uri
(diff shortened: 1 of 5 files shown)


chrislo (approved):
Feels like a good improvement to me, it makes that controller more idiomatic for sure.

PR #2185, opened 2023-06-01
Fix 2SV flash alert displayed on success after failure

floehopper:
Trello: https://trello.com/c/9C8iNaKn

In `Devise::TwoStepVerificationSessionController#create`, if the 2SV code was invalid, the flash alert was set and the "new" template was *rendered*.  Although the flash alert was rendered at this point, it was not removed from the flash. So if the user then submitted the correct code, they were redirected to the "prior flow" (the root path by default) and the flash alert was (incorrectly) displayed again.

This commit uses `ActionDispatch::Flash::FlashHash#now` to set the flash alert - it is now removed as soon as it is rendered for the first time in the "new" template and not the second time in the subsequent "success" page.

Fixes #1417.


jkempster34 (approved):
Makes sense to me! ✅

PR #2188, opened 2023-06-05
Fix flakey sign-in and sign-out integration tests

floehopper:
* Fix `AnalyticsHelper#refute_dimension_is_set` assertion which hasn't been doing anything useful since [this commit](https://github.com/alphagov/signon/commit/7fe1e64b593e01e3cc1d38049c78778262a47a3d).
* Use Capybara matcher in `AnalyticsHelper` assertions so they will use the built-in waiting mechanism in the hope this fixes the intermittent integration test failures we've seen in `test/integration/sign_in_test.rb` &
`test/integration/sign_out_test.rb`.

See commit notes for more details.

### DIFF
--- a/test/support/analytics_helpers.rb
+++ b/test/support/analytics_helpers.rb
@@ -10,12 +10,12 @@ module AnalyticsHelpers
 
   def refute_dimension_is_set(dimension)
     js_code = dimension_set_js_code(dimension)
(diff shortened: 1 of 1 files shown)


chrislo (approved):
This looks good to me. Fingers crossed it fixes the flakey test - I guess we'll know if that has happened if we stop seeing the flakey test failures?

PR #2192, opened 2023-06-06
Remove broken analytics event tracking code

floehopper:
While working on #2174 we discovered that no events have been sent to Google Analytics from the signon app since June 2017. This strongly suggests that nobody has been looking at them and so it seems sensible to simplify the app by removing the relevant code.

I haven't spent any time investigating why the code hasn't been working, but my guess is that something changed in govuk_admin_template and the signon app wasn't updated accordingly, or perhaps a bit like in #2190 it got broken inadvertently when some of the pages were moved to follow the Design System.

### DIFF
--- a/app/helpers/analytics_helper.rb
+++ b/app/helpers/analytics_helper.rb
@@ -1,9 +1,0 @@
-module AnalyticsHelper
-  def track_analytics_data(type, message)
-    {
-      "module" => "auto-track-event",
-      "track-action" => "alert-#{type}",
-      "track-label" => strip_tags(message),
-    }
-  end
-end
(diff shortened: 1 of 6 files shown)


chrislo (approved):
This looks good to me! I had a bit of a poke around to see if there was any other code that could be removed. I noticed there's some GA testing happening in, e.g. `inviting_users_test.rb` but that all seems to be based on page views, not events so it needs to stay I think.

floehopper:
@chrislo 

> I noticed there's some GA testing happening in, e.g. inviting_users_test.rb but that all seems to be based on page views, not events so it needs to stay I think.

Yes, I added those tests in #2190 and they are related to page views, 👍

PR #2193, opened 2023-06-07
Replace QR Code PNG with SVG

chrislo:
Trello: https://trello.com/c/tnjBnZ7j/77-2sv-qr-code-is-not-responsive-and-generates-a-content-security-policy-report

This replaces the PNG QR code on the 2SV setup page with an SVG. This has the following benefits:

1. It removes the use of a `data:` URL which was throwing a console warning due to the content security policy provided by the `govuk_app_config` gem[1]
2. It allows the QR code to be displayed in the development environment (the PNG image `data` URL was previously blocked due to 1.)
3. It makes the QR code responsive to the screen width

I've enabled the `use_path` option in the call to `as_svg` to reduce the size of the generated SVG code. The `viewbox` option makes the SVG responsive to its container width.

[1] https://github.com/alphagov/govuk_app_config/blob/main/lib/govuk_app_config/govuk_content_security_policy.rb#L46

# Before 
before

# After
after


floehopper (approved):
This looks good to me - good work!

I've tried this out in Chrome on MacOS by actually setting up 2SV for a new user to check the QR code is actually "scannable" and it all worked OK (using 1Password to do the scanning). 🙌 

I can also confirm that the QR code now displays OK even in development with no errors/warnings in the console! 🎉 

I've also viewed the page using a bunch of different emulated devices and the responsive design seems to work well.

I have a few minor suggestions:

1. I think the CSP issue actually appears as an "error" (as opposed to a "warning") in the console in production. However, the reason the error doesn't actually cause a problem for users is that the CSP production environments is set to "report-only" mode, i.e. the error is sent to a reporting endpoint, it doesn't actually break the image. I think it might be worth clarifying that in the PR description. It might be worth linking to the [GOV.UK CSP History](https://docs.publishing.service.gov.uk/manual/content-security-policy.html#gov-uk-csp-history) which explains how we're currently using "report-only" mode, but we are trying to move to "enforcement" mode. The change in this PR will be a small step in that direction.
2. It looks as if almost all modern browsers have [basic support for SVGs](https://caniuse.com/svg), but it might be worth cross-referencing the "Can I use" matrix with any list we have of browsers we're supposed to support, e.g. note 3 on that page mentions a problem with SVG scaling in IE9-11 & Edge that might be worth investigating further.
3. I wonder if it might be better to have a bit more whitespace between the QR code and the details element immediately below it - it looks a bit "tight" at the moment. One possibly related thing I wondered was should other bits of the page be wrapped in divs with the "govuk-grid-row" CSS class or is it legit to use it as a one-off here?
4. Similarly I wonder if it might be better to have a bit more whitespace to the left and right of the QR code - it currently looks a bit "tight" on narrow screened devices.
5. I don't think it's essential (because the code is exercised by integration tests), but it might be nice to add a unit test to at least call the new `TwoStepVerificationHelper#qr_code_svg` method, even if it just checks that it returns a string.

I don't think any of these should necessarily block deployment, so I'm going to approve it!

chrislo:
> It looks as if almost all modern browsers have [basic support for SVGs](https://caniuse.com/svg), but it might be worth cross-referencing the "Can I use" matrix with any list we have of browsers we're supposed to support, e.g. note 3 on that page mentions a problem with SVG scaling in IE9-11 & Edge that might be worth investigating further.

The [developer docs](https://docs.publishing.service.gov.uk/manual/browser-support.html) suggest that browser support for publishing apps should be based on analysis of the browsers used by users as per Google Analytics. 

Checking GA for Signon I see Edge (41.17%), Chrome (39.47%), Android Browser (10.70%), Safari (5.17%), Firefox (1.57%), and Internet Explorer 11 (1.03%) account for over 99% of browers used.

The caniuse matrix suggests that there may be issues scaling SVGs on Edge and IE11 so I checked those using a Windows 10 environment on Browserstack. Edge scales the SVG correctly, while IE11 doesn't scale the SVG

Screenshot 2023-06-07 at 15 48 20

However, I think the SVG is still usable and not really any worse than the PNG presented currently and as it accounts for only 1% of users, I think it should be ok as-is.

chrislo:
> I wonder if it might be better to have a bit more whitespace between the QR code and the details element immediately below it - it looks a bit "tight" at the moment. One possibly related thing I wondered was should other bits of the page be wrapped in divs with the "govuk-grid-row" CSS class or is it legit to use it as a one-off here?

> Similarly I wonder if it might be better to have a bit more whitespace to the left and right of the QR code - it currently looks a bit "tight" on narrow screened devices.

I've added 5px of offset/padding to the generated SVG in 97591799e5 to address this.

chrislo:
> I think the CSP issue actually appears as an "error" (as opposed to a "warning") in the console in production. However, the reason the error doesn't actually cause a problem for users is that the CSP production environments is set to "report-only" mode, i.e. the error is sent to a reporting endpoint, it doesn't actually break the image. I think it might be worth clarifying that in the PR description. It might be worth linking to the [GOV.UK CSP History](https://docs.publishing.service.gov.uk/manual/content-security-policy.html#gov-uk-csp-history) which explains how we're currently using "report-only" mode, but we are trying to move to "enforcement" mode. The change in this PR will be a small step in that direction.

I'll improve the commit message to clarify when I squash, rebase and merge. Thanks!

chrislo:
I'm going to squash the commits I introduced to address @floehopper's feedback, rebase on main and merge.

PR #2200, opened 2023-06-12
Improvements to the configuration of the SSO syncing functionality

floehopper:
Trello: https://trello.com/c/8kRu7Zwi

Recently we saw [a bunch of `SSOPushCredential::UserNotFound` exceptions](https://govuk.sentry.io/issues/3832203443/events/?project=202251). This turned out to be due to the `SSO_PUSH_USER_EMAIL` environment variable not being set and that has since been resolved. However, @sengi, @BeckaL & @kevindew all suggested the `SSOPushCredential` code could be improved.

This PR is my attempt at addressing the concerns raised. As well as a number of refactorings/simplifications, the significant changes in this PR are as follows:

1. The email address that was provided by `SSO_PUSH_USER_EMAIL` is now hard-coded since it doesn't really need to be configurable. I plan to remove this environment variable from the `govuk-helm-charts` repo once this PR is merged and deployed.
2. The `ApiUser` used for the SSO syncing functionality is now automatically created if it doesn't already exist. I was originally planning to fail fast at app start-up time if the user did not exist, but I think there are a couple of advantages of automatically creating the user. Please let me know if I've missed something here!
3. The removal of the comment in `config/initializers/sso_push.rb`. I feel as if the original version of the comment was misleading, because it implied that if the `SSO_PUSH_USER_EMAIL` environment variable was not set, then the app would work fine except that the SSO syncing functionality would be disabled. However, in fact (as we saw) the functionality was not disabled in an "orderly manner"; the functionality *failed* due to exceptions being raised which was far from ideal!

See the individual commit notes for more details.


chrislo (approved):
This all looks great to me @floehopper. I think it would be worth one or more of the folks who suggested these changes also taking a look because I can't guarantee I know about any external impacts of making this change - but as stand-alone changes they look like a great simplification.  

(PS: I appreciated the tip to use `git diff --ignore-all-space`!)

sengi (approved):
Premise of the change LGTM. (I haven't reviewed line-by-line.)

FWIW, I'm pretty sure that even if someone outside GDS wanted to deploy their own Signon again in future, the hardcoded `signon+permissions@alphagov.co.uk` would still be fine. It looks like an email address, but it's really more like a protocol extension ID so it actually makes sense for it to be a fixed string even in the general case. That address hasn't been able to receive mail for years (if ever?) and Signon doesn't send mail to it.

sengi:
Thanks so much for cleaning this up btw!

BeckaL (approved):
This all looks sensible to me, thanks for nice easy to follow commits. This looks like a massive improvement on a bit of the codebase that was overly complicated / a bit confusing ⭐

floehopper:
Thanks, all. Much appreciated!

PR #2202, opened 2023-06-14
Move SSO Push check from job creation to job execution time

floehopper:
Trello: https://trello.com/c/chkpC1am

We've been seeing `SSOPushError` exceptions for an application even having disabled its support for push updates. We believe this is happening, because some jobs were created before support for push updates was disabled, the job then failed, and now it's retrying. And since the check against `Doorkeeper::Application#supports_push_updates?` was only happening at job creation time, the job will keep failing and triggering these exceptions.

By moving the check to job execution time, the failed jobs will succeed when they next retry without triggering an exception. This change will also protect us against something similar happening again in the future.

Note that I had to remove `with_supported_permissions: %w[user_update_permission]` from the creation of the application in `PermissionUpdaterTest#setup`, because that permission now gets automatically added in `Doorkeeper::Application#create_user_update_supported_permission` which is triggered by an `after_save` callback not that `Doorkeeper::Application#supports_push_updates` is `true`. Leaving it in place wasn't an option, because it triggered an exception due to a database uniqueness constraint. In any case, the new setup seems more realistic.

### DIFF
--- a/app/jobs/permission_updater.rb
+++ b/app/jobs/permission_updater.rb
@@ -6,6 +6,7 @@ class PermissionUpdater < PushUserUpdatesJob
     application = ::Doorkeeper::Application.find_by(id: application_id)
     # It's possible they've been deleted between when the job was scheduled and run.
     return if user.nil? || application.nil?
+    return unless application.supports_push_updates?
 
     api = SSOPushClient.new(application)
     presenter = UserOAuthPresenter.new(user, application)
--- a/app/jobs/push_user_updates_job.rb
+++ b/app/jobs/push_user_updates_job.rb
@@ -7,7 +7,7 @@ class PushUserUpdatesJob < ApplicationJob
 
   class << self
     def perform_on(user)
-      user.applications_used.select(&:supports_push_updates?)
+      user.applications_used
         .each { |application| perform_later(user.uid, application.id) }
     end
   end
--- a/app/jobs/reauth_enforcer.rb
+++ b/app/jobs/reauth_enforcer.rb
@@ -5,6 +5,7 @@ class ReauthEnforcer < PushUserUpdatesJob
     application = ::Doorkeeper::Application.find_by(id: application_id)
     # It's possible the application has been deleted between when the job was scheduled and run.
     return if application.nil?
+    return unless application.supports_push_updates?
 
     api = SSOPushClient.new(application)
     api.reauth_user(uid)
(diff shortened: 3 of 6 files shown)


chrislo (approved):
Looks good to me. As we discussed earlier this new job code should be run when the failing jobs are next picked up to be worked on. If we wanted to know when that was likely to happen, perhaps we could look at the current retry count against those jobs if we can get access to the queues?

floehopper:
@chrislo 

> Looks good to me. As we discussed earlier this new job code should be run when the failing jobs are next picked up to be worked on. If we wanted to know when that was likely to happen, perhaps we could look at the current retry count against those jobs if we can get access to the queues?

I ran the following code on integration:

```ruby
require 'sidekiq/monitor'
Sidekiq::RetrySet.new.each do |job|
  p [job.display_class, job.display_args]
end
```

And got the following output:

```
["ReauthEnforcer", ["ea7ac5c0-e1be-013b-cc9c-redacted", 44576]]
["ReauthEnforcer", ["ea7ac5c0-e1be-013b-cc9c-redacted", 44576]]
["ReauthEnforcer", ["ea7ac5c0-e1be-013b-cc9c-redacted", 44576]]
["ReauthEnforcer", ["3756bd30-29ab-0135-c3ba-redacted", 44576]]
["ReauthEnforcer", ["3756bd30-29ab-0135-c3ba-redacted", 44576]]
["ReauthEnforcer", ["3756bd30-29ab-0135-c3ba-redacted", 44576]]
["ReauthEnforcer", ["3756bd30-29ab-0135-c3ba-redacted", 44576]]
["ReauthEnforcer", ["3756bd30-29ab-0135-c3ba-redacted", 44576]]
["ReauthEnforcer", ["ea7ac5c0-e1be-013b-cc9c-redacted", 44576]]
["ReauthEnforcer", ["ea7ac5c0-e1be-013b-cc9c-redacted", 44576]]
["ReauthEnforcer", ["ea7ac5c0-e1be-013b-cc9c-redacted", 44576]]
["ReauthEnforcer", ["ea7ac5c0-e1be-013b-cc9c-redacted", 44576]]
["ReauthEnforcer", ["ea7ac5c0-e1be-013b-cc9c-redacted", 44576]]
["ReauthEnforcer", ["ea7ac5c0-e1be-013b-cc9c-redacted", 44576]]
```

PR #2205, opened 2023-06-15
Simplify the User#authorised_applications method

floehopper:


chrislo (approved):
Good simplifications!

PR #2209, opened 2023-06-15
Avoid ActionView::TemplateError on signin-required page

floehopper:
Trello: https://trello.com/c/RPnXXRR2

We've been seeing `ActionView::TemplateError` exceptions on this page. It seems as if it's possible for the `application` to be `nil` which seems to be reinforced by the use of `Object#try` when setting `session[:signin_missing_for_application]` in `SigninRequiredAuthorizationsController` actions.

I did try to work out under what circumstances the `application` might not be set, but it's really hard to follow all the code involving the Devise, Doorkeeper and/or Pundit gems. So for the moment, I've just added a guard condition to `RootHelper#gds_only_application_and_non_gds_user?` to make sure the user doesn't end up seeing an error page.

### DIFF
--- a/app/helpers/root_helper.rb
+++ b/app/helpers/root_helper.rb
@@ -1,6 +1,6 @@
 module RootHelper
   def gds_only_application_and_non_gds_user?(application)
-    application.gds_only? && !current_user.belongs_to_gds?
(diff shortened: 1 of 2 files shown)


chrislo (approved):
Feels like a pragmatic way to squash this exception 👍

PR #2210, opened 2023-06-15
Use design system for batch invitations show

chrislo:
Trello: https://trello.com/c/jPK7WHy8/135-migrate-batchinvitations-to-use-design-system

This PR migrates the batch invitations show page (e.g. http://signon.dev.gov.uk/batch_invitations/1) to use the design system. 

I've made the following significant choices: 

1. Use the `notice` component to indicate when a batch is still being processed (the previous `alert-info` div)
2. Use a `success_alert` component to indicate that all records were successfully processed (the previous `alert-success` div)
3. Use an `error_alert` component to indicate if one or more records in the batch couldn't be processed (the previous `alert-danger` div). The previous alert didn't provide a mechanism for the user to fix the error so for now I've retained that behaviour by using an `error_alert` rather than a `form_error_summary` for example.
4. I've removed the button to link to "All users". The design system docs suggest that buttons should only be used to "move through a transaction" and that isn't really what it was doing here. I've instead added a breadcrumb trail which includes "all users" to reinstate the behaviour provided by this button.

# Before
before

# After
success
in_progress
error

### DIFF
--- a/app/controllers/batch_invitations_controller.rb
+++ b/app/controllers/batch_invitations_controller.rb
@@ -7,6 +7,8 @@ class BatchInvitationsController < ApplicationController
   helper_method :applications_and_permissions
   helper_method :recent_batch_invitations
 
+  layout "admin_layout", only: %w[show]
+
   def new
     @batch_invitation = BatchInvitation.new(organisation_id: current_user.organisation_id)
     authorize @batch_invitation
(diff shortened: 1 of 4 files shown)


chrisroos (approved):

chrislo:
Thanks @chrisroos @floehopper. I'll rebase this on main and force push before merging.

PR #2213, opened 2023-06-19
Improve logging of failed batch invitations

chrislo:
Trello: https://trello.com/c/8oHupLq2/138-investigate-errors-with-uploading-batch-of-users-to-production

We've recently received notice that a user attempted to upload a batch of users and some or all of those users failed. We don't currently provide much guidance to the user on why their batches have failed, and in fact, we don't receive any information as the product team because those exceptions are currently being rescued silently.

This commit logs (via `GovukError`) some information about why the process failed, and I'm hoping we'll be able to use that information to improve the experience for the users, or at least be able to deal with their issues manually via the second-line process.

### DIFF
--- a/app/models/batch_invitation_user.rb
+++ b/app/models/batch_invitation_user.rb
@@ -19,8 +19,9 @@ class BatchInvitationUser < ApplicationRecord
     )
 
     invite_user_with_attributes(sanitised_attributes, inviting_user)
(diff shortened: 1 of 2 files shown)


chrisroos (approved):

chrislo:
Thanks @chrisroos! Rebasing on main before merging.

PR #2214, opened 2023-06-20
Migrate api_users new page to design system

chrislo:
Trello: https://trello.com/c/XwCQCdMu/140-migrate-apiusers-new-to-use-design-system

This commit moves the api_users new page to use the design system. The changes here should be fairly uncontroversial, but have involved rewriting most of the template.

I've specified that the `create` action should use the `admin_layout` in the ApiUsersController so that the correct layout is used when the page re-renders on validation error.

The `_form_fields` partial referenced in the old template is still used by the `edit` view, which will be migrated as a separate piece of work.

I've had to slightly change the integration test to target the error summary div provided by the design system, and to use a regex to match the text rather than an exact string (since the design system `error_summary` partial suggests adding a title).

# Before
old

# After
new
errors

### DIFF
--- a/app/controllers/api_users_controller.rb
+++ b/app/controllers/api_users_controller.rb
@@ -1,7 +1,7 @@
 class ApiUsersController < ApplicationController
   include UserPermissionsControllerMethods
 
-  layout "admin_layout", only: %w[index]
+  layout "admin_layout", only: %w[index new create]
 
   before_action :authenticate_user!
   before_action :load_and_authorize_api_user, only: %i[edit update]
--- a/app/views/api_users/new.html.erb
+++ b/app/views/api_users/new.html.erb
@@ -1,9 +1,45 @@
 <% content_for :title, "Create new API user" %>
 
-<h1>Create new API user</h1>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for @api_user do |f| %>
+      <% if @api_user.errors.present? %>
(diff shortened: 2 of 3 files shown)


chrislo:
@pdagnall - I've noticed in most of the pattern library examples form field labels are often phrased as a question. Do you think we should do something similar here, e.g. "What is the user's email address?" rather than "Email"?

chrisroos (approved):

pdagnall:
> @pdagnall - I've noticed in most of the pattern library examples form field labels are often phrased as a question. Do you think we should do something similar here, e.g. "What is the user's email address?" rather than "Email"?

[Names](https://design-system.service.gov.uk/patterns/names/) & [emails](https://design-system.service.gov.uk/patterns/email-addresses/) have their own patterns which typically are just 'name' or 'email' - but I think we could be more helpful here and give some context

I think we could bring clarity by explaining that we need the new users name, and that they will be invited by email.
For example:
Full name
This will be the name affiliated to the new API user

Email address
We will use this to invite the new API user to log in to publisher

chrislo:
@pdagnall thank you! I'm happy to keep "name" and "email" given they're established patterns, that's good to know. I think some additional hint text might be very useful here, but on closer investigation I'm not sure what we use the API user's email for (we don't seem to send them invitations), and in real data the API user's name is often, but not always, an application or team. I'm going to capture adding additional hint text as a future enhancement to this page, and merge as-is.

chrislo:
Rebasing on main before merging.

PR #2217, opened 2023-06-20
Delete migrations from before 2023

chrislo:
I don't think there's much utility in keeping very old migrations hanging around in the codebase. They make it slightly harder to grep and apply style/linting changes and are very unlikely to be re-run.


chrisroos (approved):
Seems reasonable to me :+1:

PR #2218, opened 2023-06-20
Migrate authorisations new to design system

chrislo:
Use the design system to render the authorisations new page (e.g. `/api_users/1/authorisations/new`).

# Before 

before

# After

after


chrisroos (approved):

PR #2219, opened 2023-06-21
Fix typo in call to GovukError.notify

chrislo:
I introduced this typo in 37b9ce89.


chrisroos (approved):
I can see that [`GovukError.notify` is expecting an `extra` arg](https://github.com/alphagov/govuk_app_config/blob/e615e9b14966d992dd2b5a7a558a3e110d4fee5d/lib/govuk_app_config/govuk_error.rb#L17) instead of `extras` so I'm assuming this is all OK :+1:

PR #2221, opened 2023-06-22
Move bulk_grant_permission_sets/new page to design system

chrislo:
Trello: https://trello.com/c/Kf19VBou/148-migrate-bulkgrantpermissionsets-new-to-use-design-system

We've decided to replace this page with a simple form that allows an admin to grant the signin permission to all users for a specific application, since that is all the current system seems to be used for (there are three BulkGrantPermissionSet records in the production database, and they all grant "signin" permission to one or two applications).

This means we don't have to worry about finding a design-system alternative for the permissions multi-select box.

We think this can be extended to include specific permissions at a later date (perhaps as a second "step" in the flow) if we want/need to.

We still have a page that shows the results of previous runs of this bulk update (e.g. `http://signon.dev.gov.uk/bulk_grant_permission_sets/1`) so I haven't been able to simplify the code in BulkGrantPermissionSet any further at this stage. Perhaps when we come to port that page to the design system we can decide if we want to continue to support viewing details of batches that were processed with multiple applications and/or permissions and simplify the code further at that point.

# Before

Screenshot_2023-06-26_14-35-41

# After

Screenshot_2023-06-26_14-35-10

### DIFF
--- a/app/views/bulk_grant_permission_sets/new.html.erb
+++ b/app/views/bulk_grant_permission_sets/new.html.erb
@@ -1,25 +1,14 @@
-<% content_for :title, "Grant permissions to all users" %>
+<% content_for :title, "Grant access to all users" %>
 
-<div class="page-title">
-  <h1>Grant permissions to all users</h1>
-</div>
+<%= form_for @bulk_grant_permission_set do |f| %>
+  <%= render "govuk_publishing_components/components/select", {
+    id: "application_id",
+    label: "Application",
+    name: "application_id",
+    options: Doorkeeper::Application.all.map { |application| { text: application.name, value: application.id } }
+    } %>
 
-<div class="well">
-  <%= form_for @bulk_grant_permission_set do |f| %>
-    <h2>Permissions to grant</h2>
-
-    <% if @bulk_grant_permission_set.errors.count > 0 %>
-      <div class="alert alert-danger">
-        <ul>
-          <% @bulk_grant_permission_set.errors.full_messages.each do |message| %>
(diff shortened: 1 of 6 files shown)


app/views/bulk_grant_permission_sets/new.html.erb
```diff
@@ -1,25 +1,14 @@
...
-  <% end %>
-</div>
+<% content_for :title, "Grant access to all users" %>
+
+<%= form_for @bulk_grant_permission_set do |f| %>
+  <%= render "govuk_publishing_components/components/select", {
+    id: "application_id",
+    label: "Application",
+    name: "application_id",
+    options: Doorkeeper::Application.all.map { |application| { text: application.name, value: application.id } }
```
chrisroos, line 8: [In the shared/user_permissions partial the list of applications appears to be based on those that the user has access to](https://github.com/alphagov/signon/blob/41505143e821d5e5a64e7daa9ad9776ea8b3654d/app/views/shared/_user_permissions.html.erb#L15). Do we need to do something similar here? Or does the partial end up returning all applications because we're passing in `User.new`?

I see that the partial also differentiates between retired and non-retired applications. Was it a conscious decision to remove that distinction?

Should the applications be sorted alphabetically? Or is there some default sort order that means they're alphabetical?

chrisroos: Thanks @chrislo - I've been through the extra commits and it all looks OK to me. I'll get this approved now.

chrisroos (approved):
All looks good to me :+1:

PR #2224, opened 2023-06-26
Add client side validation to catch common 2sv code errors

chrislo:
Trello: https://trello.com/c/eZussIrd/158-accept-2sv-codes-with-spaces

We've had reports from our users that when they type in the code from their 2sv application, if they include a space (which the apps often include in their display between each group of 3 digits) we don't accept the code and the "Sorry that code didn’t work. Please try again." error doesn't help them resolve the issue.

This commit improves things slightly by changing the type of the input field to "number" and including a regex pattern to match 6 consecutive numbers. I've added some JavaScript to further customise the error message (by default it is "Please match the format requested") to be more helpful.

Note that we need to clear the validation message with an `oninput` handler to clear any errors previously set by the `oninvalid` handler if the user has already made an invalid attempt.

We think a longer-term solution to this problem is to improve the input field (for example by having 6 separate boxes, one for each number). I think it would also be a good idea to change the controller code to use a model with validations so that we can do more comprehensive validation of the code on the server side. But those are bigger changes and this feels like a useful improvement to the current behaviour.

Screenshot 2023-06-26 at 16 56 22

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,4 +1,5 @@
 //= require govuk_publishing_components/dependencies
 //= require govuk_publishing_components/all_components
 //= require password-strength-indicator
+//= require two-factor-validation-error
 //= require rails-ujs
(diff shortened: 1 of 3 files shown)


chrislo:
We discussed this PR in standup this morning and @pdagnall pointed me at [this pattern from the design system](https://design-system.service.gov.uk/patterns/confirm-a-phone-number/) which says we should ""Let the user enter the code in whatever format is familiar to them. Allow additional spaces, hyphens and dashes." 

I'm going to close this PR in favour of implementing some more liberal server-side validation of the code.

PR #2227, opened 2023-06-27
Fix batch invitation for organisations that require 2sv

chrislo:
Trello: https://trello.com/c/8oHupLq2

We've been seeing a "User not persisted" exception in Sentry[1] when a user batch invites users whose organisation requires 2sv. This was because we were not providing the `require_2sv` parameter in the call to `User.invite!`. We've been validating that created users must have `require_2sv` set to `true` when their organisation requires it since c5a4fedc57 but hadn't updated the batch invitation tool to reflect that.

This commit sets the appropriate `require_2sv` value on the invited user. I've made sure that the value is true even when no organisation is provided to the batch or the individual user - I'm not entirely sure if it's possible for the batch upload code to reach that state but given that we're now enforcing 2sv for all users unless they seek an exemption it seems like a sensible fallback value.

[1] e.g. https://govuk.sentry.io/issues/4270811143/?project=202251

### DIFF
--- a/app/models/batch_invitation_user.rb
+++ b/app/models/batch_invitation_user.rb
@@ -14,6 +14,7 @@ class BatchInvitationUser < ApplicationRecord
         email:,
         organisation_id:,
         supported_permission_ids:,
+        require_2sv:,
       },
       inviting_user,
     )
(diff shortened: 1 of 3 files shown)


chrisroos (approved):
Looks good to me.

chrisroos:
Rebased on main and force pushed in preparation for merging.

PR #2230, opened 2023-06-29
Accept 2sv codes with spaces

chrislo:
Trello: https://trello.com/c/eZussIrd

We've had reports from our users that when they type in the code from their 2sv application, if they include a space (which the apps often include in their display between each group of 3 digits) we don't
accept the code and the "Sorry that code didn’t work. Please try again." error doesn't help them resolve the issue.

This PR allows the user to separate the digits of their code with either space or `-` characters, and also include leading or trailing spaces. So for a valid code `123456` we now accept, for example, ` 123456 `, `123 456`, `123-456`, `12 34 56`.

The first two commits in this PR aren't strictly required to achieve this change, but I found it a little confusing that we had two slightly different ways of asking the user for the 6 digit code so I took the opportunity to extract the generation of that input field into a helper method and make them consistent.

### DIFF
--- a/app/controllers/devise/two_step_verification_controller.rb
+++ b/app/controllers/devise/two_step_verification_controller.rb
@@ -50,8 +50,9 @@ private
 
   def verify_code_and_update
     @otp_secret_key = params[:otp_secret_key]
-    totp = ROTP::TOTP.new(@otp_secret_key)
-    if totp.verify(params[:code], drift_behind: User::MAX_2SV_DRIFT_SECONDS)
+    result = CodeVerifier.new(params[:code], @otp_secret_key).verify
+
+    if result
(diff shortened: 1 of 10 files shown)


floehopper (commented):
I've had an initial look through this and it all looks good to me. I'm in a bit of a rush and so I haven't yet had a chance to actually test it in the browser, so I'm not actually going to approve it now.

chrisroos:
I've captured screenshots to show the change in the user interface from this PR.

## Before this PR

Screen Shot 2023-07-10 at 11 33 28

## After this PR

Screen Shot 2023-07-10 at 11 27 29

chrisroos (approved):
This all looks good to me, @chrislo :+1:

PR #2231, opened 2023-06-29
Remove references to sign-on-o-tron

floehopper:
This app was renamed from Sign-on-o-tron to Signon some years ago, but a bunch of the code and docs still reference the old name which is confusing.

A couple of the changes are not without risk of breaking things:

1. Changing the name of the Basic HTTP Authentication realm
2. Changing the name of the SSO Push user

However, I think the risk is worth it for reducing developer confusion! What do you think?

See the individual commit notes for more details.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -69,7 +69,7 @@ Devise.setup do |config|
   # config.http_authenticatable_on_xhr = true
 
   # The realm used in Http Basic Authentication. "Application" by default.
-  config.http_authentication_realm = "Sign-On-O-Tron"
+  config.http_authentication_realm = "Signon"
 
   # It will change confirmation, password recovery and other workflows
   # to behave the same regardless if the e-mail provided was right or wrong.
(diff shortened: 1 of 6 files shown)


config/initializers/devise.rb
```diff
@@ -69,7 +69,7 @@
   # config.http_authenticatable_on_xhr = true
 
   # The realm used in Http Basic Authentication. "Application" by default.
-  config.http_authentication_realm = "Sign-On-O-Tron"
+  config.http_authentication_realm = "Signon"
```
chrislo, line 72: I had a [search through the alphagov organisation](https://github.com/search?q=org%3Aalphagov%20Sign-On-O-Tron&type=code) and I couldn't see anywhere where this realm was being used explicitly in other codebases. That doesn't mean it's not being set in, for example, an ENV variable or something but I agree it's unlikely!

chrislo (approved):
I agree this makes things less confusing and as far as I can tell it's safe to make the change 👍

config/initializers/devise.rb
floehopper: Thanks, Chris!

PR #2234, opened 2023-06-29
Remove Doorkeeper's /oauth/applications routes

chrisroos:
The Doorkeeper Gem makes several /oauth/applications routes available to allow users to CRUD Applications. We don't want to allow access to these routes but a misconfiguration in our doorkeeper.rb initializer meant that we were only accidentally preventing access.

The first commit in this PR fixes the Doorkeeper configuration to properly prevent access to /oauth/applications.

The second commit in this PR then removes the route entirely.

### DIFF
--- a/test/integration/doorkeeper_integration_test.rb
+++ b/test/integration/doorkeeper_integration_test.rb
@@ -3,7 +3,7 @@ require "test_helper"
 class DoorkeeperIntegrationTest < ActionDispatch::IntegrationTest
   test "prevents access to Doorkeeper's /oauth/applications page" do
     visit "/oauth/applications"
-
-    assert_equal 403, page.status_code
+  rescue ActionController::RoutingError => e
+    assert_equal 'No route matches [GET] "/oauth/applications"', e.message
   end
 end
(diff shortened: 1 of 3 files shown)


test/integration/doorkeeper_integration_test.rb
```diff
@@ -0,0 +1,9 @@
+require "test_helper"
+
+class DoorkeeperIntegrationTest < ActionDispatch::IntegrationTest
+  test "foo" do
```
floehopper, line 4: I'm guessing you meant to rename this placeholder test name!

test/integration/doorkeeper_integration_test.rb
```diff
@@ -0,0 +1,9 @@
...
+    visit "/oauth/applications"
+
+    assert_equal 403, page.status_code
+  end
+end
```
floehopper: I think it's usual to have a newline at the end of a file.

chrisroos: My bad! Now fixed.

chrisroos:
Thanks, @chrislo :+1:

PR #2249, opened 2023-07-10
Migrate bulk grant permission sets show to design system

chrislo:
Trello: https://trello.com/c/gGY6x4mJ/173-migrate-bulkgrantpermissionsets-to-use-design-system

This PR migrates the BulkGrantPermissionSets#show action ([example, integration](https://signon.integration.publishing.service.gov.uk/bulk_grant_permission_sets/1)) to use the design system.

Note, since #2221 we now only support the bulk granting of the `signin` permission to applications. However there are older "permission sets" that granted other permissions, so I've decided that the show page should still display the additional permissions that were granted by those sets. We may want to revisit that decision later, and perhaps simplify this view further but for now, and in the interest of moving the page to the design system in a straightforward way, I've kept the current behaviour. 

# Before

before

# After

after

### DIFF
--- a/app/controllers/bulk_grant_permission_sets_controller.rb
+++ b/app/controllers/bulk_grant_permission_sets_controller.rb
@@ -1,5 +1,5 @@
 class BulkGrantPermissionSetsController < ApplicationController
-  layout "admin_layout", only: %w[new create]
+  layout "admin_layout"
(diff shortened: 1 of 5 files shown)


chrisroos (approved):

PR #2252, opened 2023-07-11
Migrate OrganisationsController#edit page to design system

chrislo:
Trello: https://trello.com/c/uPouwNMf

I decided to remove the "cancel" button as we have the breadcrumb trail to return to the previous page, and the component guide suggests we should aim to have a single button per page[1].

I've downcased the button text to be consistent with other uses of buttons.

[1] https://components.publishing.service.gov.uk/component-guide/button

# Before
before

# After
after


floehopper (approved):
LGTM 👍

PR #2257, opened 2023-07-11
Allow (super) org admins to re-try editing users

chrislo:
Trello: https://trello.com/c/1c6vLRnN

We found a bug[1] which prevented organisation and super organisation admins to retry validation errors when editing a user. This commit fixes that bug and adds a couple of tests to catch any regressions.

[1] https://govuk.sentry.io/issues/4307435188/?project=202251


floehopper (approved):
LGTM 👍

PR #2259, opened 2023-07-13
Remove Doorkeeper's /oauth/token/info route

chrisroos:
Trello: https://trello.com/c/JvVlLMB4/182-remove-oauth-token-info-doorkeeper-route

Doorkeeper's [/oauth/token/info route is used as the redirect URI for native apps in the Implicit grant flow](https://github.com/doorkeeper-gem/doorkeeper/blob/v5.6.6/lib/doorkeeper/oauth/authorization/token.rb#L77-L83). Our Doorkeeper configuration prevents the Implict grant flow so we can remove this route.

The first 9 commits improve our `SigninRequiredAuthorizationsController` and associated tests and the final commit removes the /oauth/token/info route.

### DIFF
--- a/app/controllers/signin_required_authorizations_controller.rb
+++ b/app/controllers/signin_required_authorizations_controller.rb
@@ -25,6 +25,10 @@ class SigninRequiredAuthorizationsController < Doorkeeper::AuthorizationsControl
     render plain: "Not found", status: :not_found
   end
 
+  def destroy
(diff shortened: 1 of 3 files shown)


floehopper (approved):
@chrisroos walked @chrislo & myself through these commits earlier today and I've just had another look through them in more detail. The changes all look good to me 👍

PR #2260, opened 2023-07-14
Migrate user suspension edit page to design system

chrislo:
Trello: https://trello.com/c/il1tKvaJ/181-migrate-suspensions-edit-to-use-design-system

I've removed the "cancel" button in favour of the breadcrumb trail, as the component guide suggests[1] that we should try to have a single button on each page.

I've also improved the error handling and made it follow the design system pattern by adding a Suspension model which uses ActiveRecord validations. Details in the second commit.

[1] https://components.publishing.service.gov.uk/component-guide/button

# Before

before

## With errors

before_errors

# After

after

## With errors

after_errors

### DIFF
--- a/app/controllers/suspensions_controller.rb
+++ b/app/controllers/suspensions_controller.rb
@@ -4,27 +4,24 @@ class SuspensionsController < ApplicationController
 
   layout "admin_layout"
 
-  rescue_from ActiveRecord::RecordInvalid, with: :render_edit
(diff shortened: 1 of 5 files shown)


chrisroos (approved):
Looks good to me, @chrislo! Nice work :+1:

chrislo:
Thanks @chrisroos - rebasing before merging.

PR #2269, opened 2023-07-19
Migrate UsersController#require_2sv page to design system

chrislo:
Trello: https://trello.com/c/pEwkk3Dt

I've used the standard hint, checkbox and button components.

This page is shown as part of the user creation journey for users where their organisation doesn't mandate 2sv but we want to mandate it for the specific user. I've decided not to include a breadcrumb trail as there's not really any consideration made for navigating forward and backwards in this journey (the record is saved and the user emailed before we reach this page, for example).

# Before
before

# After
after

### DIFF
--- a/app/views/users/require_2sv.html.erb
+++ b/app/views/users/require_2sv.html.erb
@@ -1,16 +1,23 @@
-<% content_for :title, "Create new user" %>
+<% content_for :title, "2-step verification settings for new user" %>
 
-<h1>2-step verification settings for new user</h1>
-
-<%= form_for @user, :html => {:class => 'well'} do |f| %>
+<%= form_for @user do |f| %>
   <%= f.hidden_field :skip_update_user_permissions, value: "true" %>
-  <p class="checkbox">
-    <%= f.label :require_2sv do %>
-      <%= f.check_box :require_2sv %> Mandate 2-step verification for this user <%= "(this will remove their exemption)" if @user.exempt_from_2sv? %>
-    <% end %>
-    <br/>
-    User will be prompted to set up 2-step verification again the next time they sign in.
-  </p>
+  <%= render "govuk_publishing_components/components/hint", {
+    text: "User will be prompted to set up 2-step verification again the next time they sign in."
+  } %>
 
-  <%= f.submit "Update user", :class => 'btn btn-success' %>
+  <%= render "govuk_publishing_components/components/checkboxes", {
+    name: "user[require_2sv]",
+    items: [
+      {
+        label: "Mandate 2-step verification for this user #{'(this will remove their exemption)' if @user.exempt_from_2sv?}",
+        value: 1,
+        checked: @user.require_2sv?,
+      }
+    ]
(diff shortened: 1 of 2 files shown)


app/views/users/require_2sv.html.erb
```diff
@@ -1,16 +1,23 @@
...
   <%= f.hidden_field :skip_update_user_permissions, value: "true" %>
-  <p class="checkbox">
-    <%= f.label :require_2sv do %>
-      <%= f.check_box :require_2sv %> Mandate 2-step verification for this user <%= "(this will remove their exemption)" if @user.exempt_from_2sv? %>
-    <% end %>
-    <br/>
-    User will be prompted to set up 2-step verification again the next time they sign in.
-  </p>
+  <%= render "govuk_publishing_components/components/hint", {
+    text: "User will be prompted to set up 2-step verification again the next time they sign in."
```
floehopper, line 6: Minor: I know you haven't changed this, but I'm a bit confused by the word "again" in this sentence:

> User will be prompted to set up 2-step verification again the next time they sign in.

Surely since this is in the user creation workflow, the user won't have ever set up their 2-step verification, so how can they be prompted to do it "again"...? Maybe I'm missing something!

app/views/users/require_2sv.html.erb
```diff
@@ -1,16 +1,23 @@
...
+<%= form_for @user do |f| %>
   <%= f.hidden_field :skip_update_user_permissions, value: "true" %>
-  <p class="checkbox">
-    <%= f.label :require_2sv do %>
-      <%= f.check_box :require_2sv %> Mandate 2-step verification for this user <%= "(this will remove their exemption)" if @user.exempt_from_2sv? %>
-    <% end %>
-    <br/>
-    User will be prompted to set up 2-step verification again the next time they sign in.
-  </p>
+  <%= render "govuk_publishing_components/components/hint", {
```
floehopper, line 5: Minor: Did you consider using the `hint_text` option on the `checkboxes` component? It's not very clear, but [this code](https://github.com/alphagov/govuk_publishing_components/blob/cb400fd0526e4f796a978ed3a932f1e1c9e05a15/lib/govuk_publishing_components/presenters/checkboxes_helper.rb#L40C10-L40C19) suggests it's an available option...

floehopper (approved):
I've added a couple of very minor suggestions/questions, but otherwise this looks good to me.

As an aside, it looks as if this form is submitted to the generic `UsersController#update` action. I wonder whether there is some potential for abuse if other user attributes were sent...? Anyway, this is an existing problem and not something you've made any worse!

app/views/users/require_2sv.html.erb
chrislo: Yes, that is odd - I found this flow a little hard to understand in general and to be honest, adding this page to the design system while having the first "step" in the create user journey in the old design doesn't really help. I think it'd be worth seeing if we can improve this whole flow and the copy at some point.

PR #2271, opened 2023-07-19
WIP: Migrate doorkeeper applications users with access to design system

chrislo:
TODO: 
- [ ] migrate page
- [ ] set format on cells correctly


app/helpers/users_with_access_helper.rb
```diff
@@ -0,0 +1,29 @@
...
+             elsif user.suspended?
+               " (suspended)"
+             elsif user.access_locked?
+               " (access locked)"
+             else
+               ""
+             end
+
+    "#{link_to(user.name, edit_user_path(user))}#{status}".html_safe
+  end
```
floehopper, line 14: Minor: It might be nice to add a unit test for this helper method and possibly the other helper methods you've added since they aren't completely trivial logic-wise.

floehopper (approved):
I've added one minor comment, but otherwise this all looks good to me! 👍

app/helpers/users_with_access_helper.rb
chrislo: Thanks James, that's a good suggestion. I'll rewrite the history to add the tests at the appropriate points.

chrislo:
Force-pushing to address @floehopper's suggestion.

chrislo:
I'm not around for the whole day today, so I'm a little reluctant to merge this now. Would you mind merging it on Monday @floehopper, if you're happy with how I've addressed your suggestion?

floehopper:
This all looks good to me - good work, @chrislo! 👍

PR #2272, opened 2023-07-20
Move 2SV exemptions page to use GOV.UK Design System

floehopper:
Trello: https://trello.com/c/oqt4y7dH

I've introduced a new `TwoStepVerificationExemption` `ActiveModel` model to implement the field validation for this form; previously the validation was all done in the controller.

My main motivation for doing this was to be able to store validation errors against a particular field so that I could implement the Design System error summary correctly with links to the invalid fields and so I could implement something close to [the date field][1] validation recommended by the Design System.

However, I think it also bring some other benefits: the code is now a bit more idiomatic Rails, it's easier to test the validation, and when validation fails the values the user entered are retained (i.e. we re-render the form rather than redirecting as we did previously). The latter is more relevant now that we're effectively using 3 text inputs for the date field rather than 3 select elements.

I've intentionally implemented the `expiry_date` validation in a "heirarchical" manner, i.e. the user doesn't immediately see a bunch of validation errors as they would if I'd used standard Rails validation. Instead they see the most "important" validation error(s) first and then if they fix those, they might see less "important" validation errors. I haven't actually seen any guidance on this, but it felt overwhelming for the user to see many different validation errors for a single date field when e.g. the problem is that none of the values have been entered at all. Also this was how the validation was working previously.

The Design System documentation re [validation errors on a date input][2] recommends highlighting individual inputs within the date input (e.g. just the `day` field) if it's possible to determine that's where the problem lies. Although the new `TwoStepVerificationExemption` model makes it feasible to do this, the `date_input` component [3] in the `govuk_publishing_components` gem doesn't currently support setting an error on the individual `day`, `month` & `year` fields. I plan to suggest changing the `date_input` component to support this in the future.

I've haven't included a "Cancel" button, because we now have a decent breadcrumb trail. This is what we've done elsewhere when converting pages to use the GOV.UK Design System.

### Before

signon dev gov uk_two_step_verification_exemptions_15805_edit

signon dev gov uk_two_step_verification_exemptions_15805_edit (1)

### After

signon dev gov uk_two_step_verification_exemptions_15805_edit (2)

signon dev gov uk_two_step_verification_exemptions_15805

[1]: https://design-system.service.gov.uk/components/date-input/
[2]: https://design-system.service.gov.uk/components/date-input/#error-messages
[3]: https://github.com/alphagov/govuk_publishing_components/blob/45ea966343920fb73990a4c3b3ed517b76219d56/app/views/govuk_publishing_components/components/_date_input.html.erb

### DIFF
--- a/app/controllers/two_step_verification_exemptions_controller.rb
+++ b/app/controllers/two_step_verification_exemptions_controller.rb
@@ -1,49 +1,27 @@
 class TwoStepVerificationExemptionsController < ApplicationController
+  layout "admin_layout"
+
   before_action :authenticate_user!, :load_and_authorize_user
 
-  def edit; end
+  def edit
+    @exemption = TwoStepVerificationExemption.from_user(@user)
+  end
 
   def update
-    error_string_or_parsed_params = parsed_params_or_error_string(params[:user])
-    case error_string_or_parsed_params
-    when String
-      flash[:alert] = error_string_or_parsed_params
-      redirect_to edit_two_step_verification_exemption_path(@user)
-    else
-      reason_for_2sv_exemption, valid_expiry_date = error_string_or_parsed_params
-      @user.exempt_from_2sv(reason_for_2sv_exemption, current_user, valid_expiry_date)
+    @exemption = TwoStepVerificationExemption.from_params(exemption_params)
+    if @exemption.valid?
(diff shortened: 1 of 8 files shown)


chrislo (approved):
This looks great 👍

PR #2274, opened 2023-07-20
Move application permissions index page to use Design System

floehopper:
Trello: https://trello.com/c/S2Gq9Nrn

I've moved the database query to obtain the list of supported permissions into a controller action, because I think that's a bit more idiomatic.

I haven't linkified the name of each permission as it was before, because the link is the same as for the "Edit" action on the same row which seemed like overkill and made the table harder to scan.

I've had to tweak the assertions in the controller test, because it's not straightforward to add the same CSS classes to the table cells. Since these classes were only added for the assertion selectors, I don't think this is a bad thing.


app/views/supported_permissions/index.html.erb
```diff
@@ -1,58 +1,72 @@
...
-        </td>
-        <td>
-          <%= link_to 'Edit', edit_doorkeeper_application_supported_permission_path(@application, supported_permission) %>
-        </td>
-      </tr>
-    <% end %>
-  </tbody>
-</table>
+<%= render "govuk_publishing_components/components/table", {
+  caption: "API users",
```
chrislo, line 33: I don't think this caption and label are correct / necessary.

chrislo (approved):
Looks great, just noticed a small issue with the table caption.

app/views/supported_permissions/index.html.erb
floehopper: Addressed in 02bb452f32d3afb07850d59297146d4b4e3a73b5.

PR #2279, opened 2023-07-26
Allow setting secrets via environment variables in development

floehopper:
Previously the secrets were hard-coded in development. I've found myself needing to change them, but if I just change the hard-coded values, there's a danger I might push sensitive secrets into the public repo. So it seems safer to allow the secrets to be overridden using the same env vars that are used in production, but fallback to the hard-coded values if those env vars are not set.


kevindew (approved):
Seems good to me 👍 I don't think we expected these would need to change when added.

PR #2281, opened 2023-07-26
Incremental improvements to the users index page

floehopper:
Trello: https://trello.com/c/QuwuPtZW

We're planning to move this page (like every other page in the Signon app) to use the GOV.UK Design System. However, we haven't yet worked out how to implement the filtering using the Design System. So in the meantime, we've decided to make some incremental improvements which will also help when we come to move the page to the Design System.

Please see the commit note for details and motivations for each change, but in summary we've:

* removed some of the less useful columns
* split the name & email into separate columns
* replaced the "Users by initial" pagination with standard pagination
* removed the pagination controls from the top of the table and reduced the number of users on each page to make the pagination controls at the bottom of the table more discoverable
* made the filter controls more prominent and more closely associated them with the table
* moved the name & email filter to the left of the other filter controls and changed the button text from "Search" to "Filter"
* moved the "Export as CSV" button away from the filter controls and particularly away from the clear filters button
* only display "Create user" button if you have actually have permission to create a user

Note: Names & emails were blurred to anonymise the users just for the screenshots - they're not really blurred in the implementation!

#### No filters applied
Screenshot 2023-07-27 at 15 42 23

#### 2 dropdown filters applied
Screenshot 2023-07-27 at 15 44 10

#### Name & email filter applied
Screenshot 2023-07-27 at 15 46 17

#### No filters applied
Screenshot 2023-07-27 at 15 51 06

#### 2 dropdown filters applied
Screenshot 2023-07-27 at 15 50 52

#### Name & email filter applied
Screenshot 2023-07-27 at 15 52 20

### DIFF
--- a/app/assets/stylesheets/_filters.scss
+++ b/app/assets/stylesheets/_filters.scss
@@ -48,6 +48,10 @@ a.filter-option:visited,
 
 .filter-by-name-field {
   width: 290px;
+
+  &::placeholder {
+    color: #757575;
+  }
 }
 
 // Nav compact
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -152,11 +152,7 @@ private
   end
 
   def paginate_users
-    @users = if @users.is_a?(Array)
-               Kaminari.paginate_array(@users).page(params[:page]).per(100)
-             else
-               @users.page(params[:page]).per(100)
-             end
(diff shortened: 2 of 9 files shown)


mike29736 (approved):
Some nice improvements :+1: Good to see it moving forward even without the design system

PR #2290, opened 2023-08-02
Clarify purpose of /oauth/access_token custom route

floehopper:
Trello: https://trello.com/c/aS6YuhX6

It wasn't obvious that the `Doorkeeper::TokensController` was needed, because the standard user sign-in workflow involved a `POST` request to `/oauth/access_token`; not to `/oauth/token`. However, we had missed the custom route defined for `/oauth/access_token` (at the bottom of `config/routes.rb`!) which is handled by `Doorkeeper::TokensController#create`.

Delving through the history of Signon, [its predecessor][1] and the [oauth2-provider gem][2] that the latter used, it became apparent that this was probably the standard route in OAuth v1 and even possibly early/draft versions of OAuth v2.

Most importantly, the `gds-sso` gem currently uses the `/oauth/access_token` path [3] and we suspect that even apps which are not using `gds-sso` may well have copied its behaviour. Although we haven't searched exhaustively, we've only seen traffic to `/oauth/access_token` in the logs and none to `/oauth/token`.

In this commit I have:

* Moved the custom route closer to where the default `Doorkeeper` routes are defined. I have confirmed there are no other routes with the `/oauth` path prefix, so this move should not change the behaviour of the app.
* Changed the comment to reference OAuth v1 versus Signon v1. If that comment is unclear, hopefully its association with this commit note will clarify matters to other developers.
* Added tests to check that the app handles `POST` requests to both `/oauth/access_token` & `/oauth/token` in `Doorkeeper::TokensController#create`.

Ideally I would have changed the `Doorkeeper` configuration to support the alternative path rather than using a custom Rails route, but that doesn't seem to be possible.

[1]: https://github.com/alphagov/sign-on-o-tron
[2]: https://github.com/freerange/oauth2-provider
[3]: https://github.com/alphagov/gds-sso/blob/af8df2c7e60f33ed6ffaf7b01e453520a45a7edc/lib/gds-sso.rb#L39


mike29736 (approved):
:ok_hand: 

I had to re-read that a few times.

Some tests exercising (and documenting) that integration sounds good to me.

floehopper:
@mike29736 

> I had to re-read that a few times.

I'm not surprised. @chrisroos had more context, so I nearly suggested he review this, but then I thought it would be good to get a fresh set of eyes on it - so thank you! 😄

PR #2291, opened 2023-08-02
Remove unwanted spaces from users names

mike29736:
https://trello.com/c/WVFDaA9C/203-remove-spaces-at-the-beginning-of-some-users-names-which-confuses-ordering

We have some cases of `User` and `BatchInvitationUser` records in our production data that have leading and/or trailing whitespace in their name columns and we'd like to remove that whitespace because, in the case of `User`, it breaks alphabetical sorting by name.

Here we introduce some code to stop that happening again and a migration to fix the existing data.

### DIFF
--- a/test/models/batch_invitation_user_test.rb
+++ b/test/models/batch_invitation_user_test.rb
@@ -1,6 +1,14 @@
 require "test_helper"
 
 class BatchInvitationUserTest < ActiveSupport::TestCase
+  context ".create!" do
+    should "strip unwanted whitespace from name before persisting" do
+      user = create(:batch_invitation_user, name: "  Ailean Millard ")
+
+      assert_equal "Ailean Millard", user.name
+    end
+  end
+
   context "invite" do
     setup do
       @inviting_user = create(:admin_user)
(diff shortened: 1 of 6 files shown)


test/models/batch_invitation_user_test.rb
```diff
@@ -1,6 +1,24 @@
...
+    end
+
+    should "strip unwanted whitespace from name before persisting" do
+      user = BatchInvitationUser.create!(
+        name: "  Ailean Millard ",
+        email: "ailean@example.com",
+        batch_invitation: @batch_invitation,
+      )
+
+      assert user.persisted?
```
floehopper, line 17: Minor: I think this assertion is redundant, because you've (sensibly) used the bang variation of `BatchInvitationUser.create!`, so an exception would be raised if the user wasn't persisted successfully.

Actually, thinking about it - perhaps it would be worth using FactoryBot's create method here so you don't have to specify the email attribute? I _think_ FactoryBot create uses an ActiveRecord bang method under the hood, so even then you might not need this assertion...

floehopper (approved):
Overall this looks good to me! I've checked the branch out locally, run the migration, and confirmed that none of the users are left with leading/trailing spaces in their names.

I've added one minor suggestion inline.

And it might be nice to change the subject line of the 2nd commit to be include the word "user", e.g. "Strip leading and trailing space from batch user names".

However, neither of those are critical, so I'm happy to mark this as approved and leave any changes to your discretion.

PR #2292, opened 2023-08-02
Change trailing_message column from varchar(255) to text

chrisroos:
https://trello.com/c/mZuBjUXu

When a user has permissions added/removed we [record an event that includes the name of each permission][1]. If the length of the string containing all of these permission names exceeded 255 characters then the following exception was raised:

ActiveRecord::ValueTooLong
Mysql2::Error: Data too long for column 'trailing_message' at row 1

I considered truncating the message if it exceeded 255 characters but can't be confident that this data isn't being used elsewhere.

[1]: https://github.com/alphagov/signon/blob/0a2293f76e0ceb5d3eaf4c5e5167037cc82b80d1/app/services/user_update.rb#L52


floehopper (approved):
LGTM 👍

chrisroos:
Thanks, @floehopper

PR #2296, opened 2023-08-03
Remove Doorkeeper client credentials grant flow

floehopper:
Trello: https://trello.com/c/Jx1718RH

By default Doorkeeper is configured to support both `authorization_code` & `client_credentials` grant flows.

The `authorization_code` grant flow is being used e.g. when a user signs in from a publishing application. However, it was unclear whether the `client_credentials` flow was being used at all.

When a `Doorkeeper::AccessToken` is created as part of the client credentials grant flow, the `resource_owner` is always set to `nil`. However, there is a `NOT NULL` database constraint on `oauth_access_tokens.resource_owner_id` which triggers an `ActiveRecord::NotNullViolation` exception.

That `NOT NULL` constraint has been in place since the Doorkeeper-related tables were first added, presumably generated by [a very early version of the Doorkeeper generator][1]. However, a few minor versions of Doorkeeper later, that same constraint was [removed][2], but it was not removed from Signon.

I'm pretty confident this means that the client credentials grant flow has never worked in Signon. Rather than fixing it, this indicates that we can't need it and so I'm removing it in this commit to reduce the "surface area" for any attacks.

For more details of my investigation into this, see the Trello card comments.

[1]: https://github.com/doorkeeper-gem/doorkeeper/commit/fb76738381f27fe4cb62d126c1f41d45bba77dce
[2]: https://github.com/doorkeeper-gem/doorkeeper/commit/66ba66d9fecae64b9e3e8918830f682bf0df718e


mike29736 (approved):
Impressive deep dive! :diving_mask: 

You didn't mention it directly in the Trello card, but maybe you already saw that the Doorkeeper commit (removing the database column constraint) was part of the same release that added support for the Client Credentials flow (https://github.com/doorkeeper-gem/doorkeeper/blob/main/CHANGELOG.md#040---2012-05-26)?

floehopper:
@mike29736 

> Impressive deep dive! 🤿

😂 

> You didn't mention it directly in the Trello card, but maybe you already saw that the Doorkeeper commit (removing the database column constraint) was part of the same release that added support for the Client Credentials flow (https://github.com/doorkeeper-gem/doorkeeper/blob/main/CHANGELOG.md#040---2012-05-26)?

Thanks for taking the time to read through my notes in the Trello card so thoroughly. I had meant to check when the client credentials flow was added to Doorkeeper - thanks for finding it! I will update the PR description & commit note to include that.

floehopper:
> > You didn't mention it directly in the Trello card, but maybe you already saw that the Doorkeeper commit (removing the database column constraint) was part of the same release that added support for the Client Credentials flow (https://github.com/doorkeeper-gem/doorkeeper/blob/main/CHANGELOG.md#040---2012-05-26)?
> 
> Thanks for taking the time to read through my notes in the Trello card so thoroughly. I had meant to check when the client credentials flow was added to Doorkeeper - thanks for finding it! I will update the PR description & commit note to include that.

Addressed in re-written/force-pushed commit: 27f43f766aca547fc6fa356aa76c83193782f975

PR #2297, opened 2023-08-07
Give batch uploading users more information when an email address in their batch is invalid

mike29736:
https://trello.com/c/HOwVtQLy/184-give-batch-uploading-users-more-information-when-an-email-address-in-their-batch-is-invalid

Until now, one invalid email address in a batch upload (CSV) would raise
an ambiguous error and give the user no information about what went
wrong.

Now, if the batch encounters a validation problem, it returns to the
original form as before but also displays a more helpful error message.

Though, as I've implemented it, the message could be more helpful. But see for yourself:

Screenshot_2023-08-07_09-07-32

### DIFF
--- a/app/controllers/batch_invitations_controller.rb
+++ b/app/controllers/batch_invitations_controller.rb
@@ -43,8 +43,6 @@ class BatchInvitationsController < ApplicationController
       return
     end
 
-    @batch_invitation.save!
-
     csv.each do |row|
       batch_user_args = {
         batch_invitation: @batch_invitation,
@@ -54,8 +52,13 @@ class BatchInvitationsController < ApplicationController
       if policy(@batch_invitation).assign_organisation_from_csv?
         batch_user_args[:organisation_slug] = row["Organisation"]
       end
-      BatchInvitationUser.create!(batch_user_args)
+      batch_user = BatchInvitationUser.new(batch_user_args)
+
+      @batch_invitation.batch_invitation_users << batch_user
     end
+
+    @batch_invitation.save!
+
     @batch_invitation.enqueue
     flash[:notice] = "Scheduled invitation of #{@batch_invitation.batch_invitation_users.count} users"
     redirect_to batch_invitation_path(@batch_invitation)
--- a/app/controllers/batch_invitations_controller.rb
+++ b/app/controllers/batch_invitations_controller.rb
@@ -54,6 +54,11 @@ class BatchInvitationsController < ApplicationController
       end
       batch_user = BatchInvitationUser.new(batch_user_args)
 
+      unless batch_user.valid?
+        flash[:alert] = batch_users_error_message(batch_user)
+        return render :new
+      end
+
       @batch_invitation.batch_invitation_users << batch_user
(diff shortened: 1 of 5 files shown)


app/controllers/batch_invitations_controller.rb
```diff
@@ -91,4 +96,17 @@ def grant_default_permissions(batch_invitation)
...
+
+    case e.attribute
+    when :email
+      "One or more emails were invalid"
+    when :name
+      "One or more names were invalid"
+    else
+      e.full_message
+    end
+  end
```
floehopper, line 111: In the commit note you say:

> This is a small improvement which avoids some of the complexities of what might be even more helpful error reports (e.g. listing one or all of the invalid email addresses).

I may well be missing something, but wouldn't it be _relatively_ easy to include the _first_ invalid email address by generating an error message along the following lines:

```ruby
"One or more emails were invalid, e.g. #{batch_user.email}"
```

I think this might make the error message more "actionable".

app/controllers/batch_invitations_controller.rb
```diff
@@ -91,4 +96,17 @@ def grant_default_permissions(batch_invitation)
...
   end
+
+  def batch_users_error_message(batch_user)
+    e = batch_user.errors.first
+
+    case e.attribute
+    when :email
+      "One or more emails were invalid"
+    when :name
+      "One or more names were invalid"
```
floehopper, line 107: I'm almost certainly missing something obvious here, but it doesn't look as if there's any validation on the `BatchInvitationUser#name` attribute. So I'm unclear how this case can happen.

floehopper (changes requested):
I think this is definitely an improvement, but I've added a comment inline suggesting we might be able to make the error message more actionable by including the first invalid email address. What do you think...?

I'm happy to chat about it if that would be easier.

app/controllers/batch_invitations_controller.rb
mike29736: Over the weekend I somehow convinced myself that including one example was more confusing for users than including none. So I removed something very close to this this morning :grimacing: 

I assume you're right

floehopper: > I assume you're right

Hmm - I wouldn't be too sure about that - it might be worth getting a 2nd opinion from Joseph or someone...?

app/controllers/batch_invitations_controller.rb
mike29736: Sorry, I'm not sure why I typed this. I'll push the fix together with the new error message when that's ready

app/controllers/batch_invitations_controller.rb
mike29736: Joseph voted for keeping it without the example. After arguing in favour of some more complicated options!

I'll write up some ideas we had and add them to that other general improvements card (https://trello.com/c/kXMWrDu1/44-improve-batchinvitations)

floehopper: 👍

PR #2298, opened 2023-08-08
Missing validation on new/edit application permission page

mike29736:
https://trello.com/c/kgyjym0j/212-missing-validation-on-new-edit-application-permission-page

This pull request adds a validation rule that results in a more helpful error message if a user attempts to give 2 permissions the same name for a single application.

Screenshot_2023-08-08_10-29-37


floehopper (approved):
LGTM 👍

PR #2300, opened 2023-08-08
Set autocomplete attribute on input elements appropriately

floehopper:
Trello: https://trello.com/c/WHsgcFDz

This explicitly sets the autocomplete attribute to "off" on all input elements unless there is [a specificly suitable value][1], e.g. "name", "email", "current-password", "new-password". Note that these values must apply for the current user and not for another user, e.g. when a user is editing the details of another user, autocomplete should be set to "off" for the "name" & "email" fields.

The original motivation behind this was that some password managers were filling in the "name" and "email" fields on the edit user page. This was because autocomplete was not set on the fields, the password manager heuristics decided that they were legitimate "name" & "email" fields, and (since there was only one matching item for the domain) the password manager filled the field with the current user's name and email address. This should no longer happen, because we are now setting autocomplete to "off" for those two fields.

I have manually tested all the affected pages in Chrome on MacOS.

[1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#values

### DIFF
--- a/app/views/api_users/_form_fields.html.erb
+++ b/app/views/api_users/_form_fields.html.erb
@@ -10,10 +10,10 @@
 
 <p class="form-group">
   <%= f.label :name %>
-  <%= f.text_field :name, autofocus: true, class: 'form-control input-md-6', required: true %>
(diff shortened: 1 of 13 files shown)


mike29736 (approved):
:+1:

PR #2302, opened 2023-08-09
Prevent new users being created with non-governmental email addresses

mike29736:
https://trello.com/c/gME44pGU/210-prevent-new-users-with-being-created-with-non-governmental-email-addresses

Right now, a user can be created in Signon with any email address at
all. For example, you can create (and evidently people have created) new
users with `gmail.com` addresses.

This openness presents a number of potential security risks.

Here we use a custom validator to reject email addresses that match any of
a short list of well known email providers' domains.

Including these validation rules in `BatchInvitationUser` as well as
`User` means that the email addresses included in CSV uploads will be
verified, too. (Note that batch uploads encountering this new validation
error will still only see our generic "One or more emails were invalid"
error message for now. But the more specific error will be available to
us when we get to work on our upcoming improvements to that page.)

:pray: Please help me out with the error message content! As I've got it,
it says, "Email is not a workplace-provided address".

### DIFF
--- a/app/models/reject_non_governmental_email_addresses_validator.rb
+++ b/app/models/reject_non_governmental_email_addresses_validator.rb
@@ -1,0 +1,25 @@
+class RejectNonGovernmentalEmailAddressesValidator < ActiveModel::EachValidator
+  NON_GOVERNMENTAL_EMAIL_DOMAIN_KEYWORDS = %w[
+    aol btinternet gmail hotmail outlook yahoo
+  ].freeze
+
+  MESSAGE = "not accepted. Please enter a workplace email to continue.".freeze
+
+  def validate_each(record, attribute, value)
+    return if value.blank?
+
+    domain_part = value.split("@").last
+
+    return if domain_part.blank?
+
+    if keyword_matchers.any? { |keyword| keyword.match?(domain_part) }
+      record.errors.add(attribute, (options[:message] || MESSAGE))
+    end
+  end
+
+private
+
+  def keyword_matchers
+    NON_GOVERNMENTAL_EMAIL_DOMAIN_KEYWORDS.map { |keyword| /\b#{keyword}\b/ }
+  end
+end
--- a/app/models/batch_invitation_user.rb
+++ b/app/models/batch_invitation_user.rb
@@ -2,6 +2,7 @@ class BatchInvitationUser < ApplicationRecord
   belongs_to :batch_invitation
 
   validates :email, presence: true, format: { with: Devise.email_regexp }
+  validates :email, reject_non_governmental_email_addresses: true, on: :create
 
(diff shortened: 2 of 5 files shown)


mike29736:
FYI I'm planning to force-push a change to that last commit today, restricting the new validation to `on: :create` only

app/models/reject_non_governmental_email_addresses_validator.rb
```diff
@@ -0,0 +1,25 @@
...
+  NON_GOVERNMENTAL_EMAIL_DOMAIN_KEYWORDS = %w[
+    aol btinternet gmail hotmail outlook yahoo
+  ].freeze
+
+  MESSAGE = "is not a workplace-provided address".freeze
+
+  def validate_each(record, attribute, value)
+    return if value.blank?
+
+    domain_part = value.split("@").last
```
floehopper, line 11: Minor: I have a feeling valid email addresses can technically have an "@" character in the username part if that part is quoted. It might be worth checking!

mike29736: Thanks, yeah, that sounds familiar... I always forget how weird email addresses can be.

But! Luckily Signon is already fairly strict with email address format, so another validation would fail if it didn't have exactly 1 `@`. And I _think_ this method is careful enough not to choke on an input that's only going to be caught by another validation rule.

(That format is configured here https://github.com/alphagov/signon/blob/main/config/initializers/devise.rb#L164 as `/\A[^@\s]+@([a-z0-9-]+\.)+[a-z0-9]+\z/`. And then that's used as you'd expect, in a Devise-provided validation, and also recently duplicated https://github.com/alphagov/signon/blob/main/app/models/batch_invitation_user.rb#L4 :smile:)

chrisroos (approved):
With the caveat that you're going to change the last commit so that it only works on create, this all looks good to me, @mike29736 :+1: 

I used this validation to check the users in integration and discovered 232 invalid:

```
users = User.find_each.select { |u| !u.valid? && u.errors[:email].find { |m| m =~ /workplace/ } }
users.length
=> 232
users.map { |u| u.email.split('@').last }.uniq.sort
=> ["aol.co.uk", "aol.com", "btinternet.com", "gmail.com", "hotmail.co.uk", "hotmail.com", "yahoo.co.uk", "yahoo.com"]
```

I do half wonder whether we should explicitly deny these domains rather than regexp matching on `\b{keyword}\b`. It seems unlikely we'd encounter this but at the moment something like `foo@aol.gov.uk` would be invalid. I won't let this idle wondering hold up you merging this PR though :-)

mike29736:
Just added:

* Nick's new copy for the error message: "Email not accepted. Please enter a workplace email to continue."
* Config restricting the validation to `on: :create`

PR #2301, opened 2023-08-09
Add user research recruitment banner

floehopper:
Trello: https://trello.com/c/yGMU3sR2

Supersedes #2295.

This is a reworked version of #2295 where I've curated the commits as per @mike29736's [suggestion](https://github.com/alphagov/signon/pull/2295#discussion_r1287358396) and changed the "Hide this" button to actually look like a button (to address @chrisroos' question in #2295 and @mike29736's [concern](https://github.com/alphagov/signon/pull/2295#discussion_r1287352083).

This adds a "sticky" banner on the Signon home page to recruit user research participants for GOV.UK Publishing. The "Find out more" button opens a Google Form (managed by the User Research folk) in a new browser tab. If the user clicks on this button, we record that in the Signon database and we never show them the banner again. If the user clicks on the "Hide this" button then we hide the banner for the duration of the current browser session, but it will be displayed to the user in their next browser session.

As per [these recommendations](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in-from-june-2022), I tested the version in #2295 against the following devices/browsers (using BrowserStack). Since making the change to the "Hide this" button implementation, I've re-tested in the same browsers...

* Windows 11
  * [x] Edge 115
  * [x] Firefox 116
  * [x] Chrome 115
* MacOS 13.4.1
  * [x] Firefox 115
  * [x] Chrome 115
  * [x] Safari 16.5.2
* iPhone SE 2022 v15.4
  * [x] Safari
  * [x] Chrome
* iPhone 14 iOS 16
  * [x] Safari
* Galaxy S23 Android 13
  * [x] Chrome
  * [x] Samsung Internet

### Desktop
Screen Shot 2023-08-09 at 15 31 21

### iPad
Screen Shot 2023-08-09 at 15 31 53

### iPhone
Screen Shot 2023-08-09 at 15 32 07

PR #2303, opened 2023-08-10
Add user research recruitment banner v3

floehopper:
Trello: https://trello.com/c/yGMU3sR2

Supersedes #2301.

This adds a "sticky" banner on the Signon home page to recruit user research participants for GOV.UK Publishing. The "Find out more" button opens a Google Form (managed by the User Research folk) in a new browser tab. If the user clicks on this button, we record that in the Signon database and we never show them the banner again. If the user clicks on the "Hide this" button then we hide the banner for the duration of the current browser session, but it will be displayed to the user in their next browser session.

This is a reworked version of #2301 where I've addressed the issue that Calum spotted where the buttons in the banner were not full width on mobile screen widths. I've solved this by using a single form wrapped around the button group with each button having a different value and combining the two controller actions into one switching on the button value. Previously we were using the Rails `button_to` helper method which was generating two forms within the button group and this was upsetting the button layout on mobiles. For now the buttons are defined using plain HTML which is what the GOV.UK Components button would generate if it supported the `govuk-button--inverse` CSS class. Our intention is to open a PR against the GOV.UK Components repo to add support for this class at which point we can address the duplication in this button HTML.

One thing I'm not 100% sure about is that I've had to use the `formtarget` attribute on the "Find out more" button in order to open the Google Form in another tab - however, it seems to be well supported in modern browsers.

As per [these recommendations](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in-from-june-2022), I've tested these changes against the following devices/browsers (using BrowserStack):

* Windows 11
  * [x] Edge 115
  * [x] Firefox 116
  * [x] Chrome 115
* MacOS 13.4.1
  * [x] Firefox 115
  * [x] Chrome 115
  * [x] Safari 16.5.2
* iPhone SE 2022 v15.4
  * [x] Safari
  * [x] Chrome
* iPhone 14 iOS 16
  * [x] Safari
* Galaxy S23 Android 13
  * [x] Chrome
  * [x] Samsung Internet

### Desktop
Screen Shot 2023-08-10 at 11 00 51

### iPad
Screen Shot 2023-08-10 at 11 00 37

### iPhone
Screen Shot 2023-08-10 at 11 00 26

### DIFF
--- a/app/assets/stylesheets/_user_research_recruitment_banner.scss
+++ b/app/assets/stylesheets/_user_research_recruitment_banner.scss
@@ -1,0 +1,22 @@
+.user-research-recruitment-banner {
+  background-color: $govuk-brand-colour;
+  @include govuk-responsive-padding(8, "top");
+}
+
+.user-research-recruitment-banner__divider {
+  border-bottom: 1px solid govuk-colour("white");
+}
+
+.user-research-recruitment-banner__title {
+  color: govuk-colour("white");
+  @include govuk-responsive-margin(5, "bottom");
+}
+
+.user-research-recruitment-banner__intro {
+  color: govuk-colour("white");
+}
+
+.user-research-recruitment-banner__buttons {
+  @include govuk-responsive-padding(6, "bottom");
+  align-items: center;
+}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,4 +1,5 @@
 @import "govuk_publishing_components/all_components";
+@import "user_research_recruitment_banner";
 
 // TODO: move into component
 .gem-c-success-alert,
--- a/app/controllers/root_controller.rb
+++ b/app/controllers/root_controller.rb
@@ -14,4 +14,11 @@ class RootController < ApplicationController
   def signin_required
     @application = ::Doorkeeper::Application.find_by(id: session.delete(:signin_missing_for_application))
   end
+
+private
+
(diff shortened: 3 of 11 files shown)


chrisroos (approved):
This all looks good to me, @floehopper :+1: Thanks for taking it on in my absence.

As discussed on the call just now:

- The "Hide this" button being smaller than the "Find out more" button looks odd to me (but I am very much not a designer).
- I found it slightly confusing that after clicking "Find out more" and then returning to Signon; the banner wasn't hidden and the "Find out more" button was yellow.

I'll leave it to you to decide what, if anything, to do with those points we discussed.

floehopper:
> As discussed on the call just now:
> 
>     * The "Hide this" button being smaller than the "Find out more" button looks odd to me (but I am very much not a designer).
> 
>     * I found it slightly confusing that after clicking "Find out more" and then returning to Signon; the banner wasn't hidden and the "Find out more" button was yellow.
> 
> 
> I'll leave it to you to decide what, if anything, to do with those points we discussed.

Having discussed these points within the team, we've agreed that this PR can be deployed and we'll create other Trello card(s) to capture these possible improvements.

floehopper:
> Having discussed these points within the team, we've agreed that this PR can be deployed and we'll create other Trello card(s) to capture these possible improvements.

Captured in https://trello.com/c/j3b9aTkU.

PR #2308, opened 2023-08-14
Revoke all access tokens when suspending a user

mike29736:
https://trello.com/c/UTcTxQKG/142-suspending-api-users-on-signon-doesnt-disable-their-tokens

Until now a suspended ApiUser is flagged as suspended in the database,
which results in Signon telling applications that the user has no
permissions. Here we add the extra step of revoking the user's access
tokens, too.

If tokens aren't revoked, a suspended user technically retains its
authorised access (potentially for years, because TTL is quite long).
It's unclear what could be achieved maliciously using that access
without any permissions, but the consensus is that it's safer and
cleaner to revoke the tokens of suspended accounts.

The intended target of this change is `ApiUser`s (i.e. applications and not
human users). The implementation is simpler if we just perform this step
for suspended `User`s more generally, and it seems safe to do that...
but I could be missing something here! Any idea how I could check that?


chrisroos (changes requested):
Hi @mike29736

Apologies it's taken so long for me to finish reviewing this PR. I had quite a lot to learn about how things work at the moment to understand the effect of this change!

> If tokens aren't revoked, a suspended user technically retains its
authorised access (potentially for years, because TTL is quite long).
It's unclear what could be achieved maliciously using that access
without any permissions, but the consensus is that it's safer and
cleaner to revoke the tokens of suspended accounts.

I've added a comment to the Trello card that demonstrates how an API Token for a suspended API User can be used to upload files to Asset Manager. So I think it depends on whether the API is checking that the API User has specific permissions, or whether (as I believe to be the case with Asset Manager) the "signin" permission allows that API User to interact fully with the API.

> The implementation is simpler if we just perform this step
for suspended Users more generally, and it seems safe to do that...
but I could be missing something here! Any idea how I could check that?

I agree that it seems safe to do this as Doorkeeper already handles the case of revoked Access Tokens. When a normal user signs in to one of the publishing apps (e.g. Whitehall) then they'll end up with an Access Token in the oauth_access_tokens table of Signon. If they sign out and then back in while that Access Token is active (i.e. not expired and not revoked) then they'll continue to use that same Access Token. If they sign out and that Access Token is revoked then the next time they sign in a new Access Token will be created.

Can you add a migration to this PR to revoke the Access Tokens for suspended users currently in the system. It might worth notifying 2ndline once we're ready to deploy this on the off chance that something is relying on using an API token of a suspended user.

While reviewing this I noticed that the `InactiveUsersSuspender#suspend` method duplicates some of the logic in `Suspendable#suspend`. At the moment this means that any users (presumably including API users) that are suspended due to inactivity won't have their Access Tokens revoked. For consistency I think we should revoke the Access Tokens irrespective of what triggered the suspension. Can you update the PR to do that, please?

I've asked for two changes based on my current understanding of the world. Let me know if you disagree and/or want to chat about it in person :-)

mike29736:
Thanks for the review, @chrisroos!

I've updated that first commit message to remove some of the doubt and reflect your comments. And I've added the same step to the `InactiveUsersSuspender`. And the migration.

It's not obvious to me how to de-duplicate those two competing user-suspending methods (`Suspension#save` and `InactiveUsersSuspender#suspend`/`#unsuspend`) without making a mess. Suggestions would be more than welcome, though.

I'll respond to your Asset Manager discovery in the Trello card

app/models/user.rb
```diff
@@ -159,6 +159,10 @@ def authorised_applications
   end
   alias_method :applications_used, :authorised_applications
 
+  def revoke_all_authorisations
+    authorisations.each(&:revoke)
```
chrisroos, line 163: Hi @mike29736 - I've just realised that this will update the revoked_at timestamp for _all_ tokens, including those that are already revoked. I think it should probably only update those that are not currently revoked so that we preserve the revoked_at timestamp of any tokens that have already been revoked. What do you think?

db/migrate/20230816164936_revoke_access_tokens_for_suspended_users.rb
```diff
@@ -0,0 +1,8 @@
+class RevokeAccessTokensForSuspendedUsers < ActiveRecord::Migration[7.0]
+  def up
+    User.with_status(User::USER_STATUS_SUSPENDED)
+      .find_each(&:revoke_all_authorisations)
```
chrisroos, line 4: @floehopper has [suggested that we shouldn't revoke tokens that are currently, but incorrectly, in use](https://trello.com/c/sIwETh2I/112-prevent-api-tokens-being-created-for-suspended-api-users#comment-64de4eb5f8d80719364c1520). I've had a look but can't see that we have any way to detect when an API token was last used. Do you know if that's possible?

app/models/user.rb
mike29736: Thanks, good catch

db/migrate/20230816164936_revoke_access_tokens_for_suspended_users.rb
chrisroos: @floehopper - how strongly do you feel that we should check whether any of these tokens are being used before revoking them?

floehopper: > @floehopper - how strongly do you feel that we should check whether any of these tokens are being used before revoking them?

I think at the very least we should be letting 2nd Line know we're doing it. Can we chat about it in the morning?

chrisroos: In integration there are 10 non-revoked/non-expired access tokens belonging to 6 suspended API users. I'm much less concerned about the impact of revoking these so we haven't investigated them as thoroughly:

```
users = User.with_status(User::USER_STATUS_SUSPENDED).select { |u| u.authorisations.where(revoked_at: nil).any? }
users.count
=> 6

users.select { |u| u.api_user? }.count
=> 6

users.flat_map { |u| u.authorisations.where(revoked_at: nil) }.count
=> 10

users.flat_map { |u| u.authorisations.where(revoked_at: nil) }.map(&:expired?).tally
=> {false=>10}
```

mike29736: Right, I'll update that method on the User model (to only revoke not-yet-revoked tokens) now.

And I think that's the only thing left to do, then?

chrisroos: > And I think that's the only thing left to do, then?

Yep! Thanks, Mike :+1:

chrisroos (approved):
Looks good to me! Thanks, @mike29736 :+1:

PR #425, opened 2015-10-23
2SV email notifications

benlovell:
This adds various user email notifications around the 2SV actions.

PR #2311, opened 2023-08-15
Fix use of INSTANCE_NAME to determine environment-specific behaviour

floehopper:
Trello: https://trello.com/c/8HOXkVAT

## Summary

Since the replatforming to k8s (a few months ago?), some application logic has not been correctly identifying the production environment. This has affected logic in the following functionality (see below for more details about each one):

* Mandating 2SV for users promoted to any admin role
* 2SV QR code issuer
* Automated email subject line, content & from address

## Problem

Previously the implementation relied on the `INSTANCE_NAME` environment variable not being set in the production environment. However, since the replatforming to k8s, that environment variable has been [incorrectly set][1] to "production".

That means that since the replatforming a bunch of logic in the app has been working _incorrectly_. This was highlighted by some non-sensical text in an automated email sent to a user whose account had been locked:

> Production account locks do not lock production accounts.

However, the problem has actually been a lot more widespread, because we've never been recognising we're in a production environment.

## Solution

I _could_ have fixed the problem by simply removing the value of `INSTANCE_NAME` in the production environment. However, I can see why the mistake was made and it looks as if it has already had to be rectified [once before][2] in the replatforming to AWS! So instead I've chosen to leave the value of `INSTANCE_NAME` as "production" and change the
implementation to check for this value.

I've performed a bunch of preparatory refactoring to make the change easier and to clarify what functionality will be affected by the change. The bug is only actually *fixed* in the final commit.

This PR will fix a bunch of issues:

### Mandating 2SV for users promoted to any admin role

Users promoted to any admin role in production will now be required to setup 2SV. Since we now mandate 2SV for most users in production, this is unlikely to be causing a security lapse, but it would be worth checking that no users have slipped through the cracks while this bug existed.

### 2SV QR code issuer

The issuer in the 2SV QR code will now be set to "GOV.UK Signon" rather than "Production GOV.UK Signon" in production. The issuer is used by Authenticator Apps to identify which website the secret belongs to. There _might_ be some confusion for people who originally setup 2SV while this bug existed who now come back to reset 2SV, but I plan to investigate that before I merge the changes.

### Email subject line

In production, the subject line of the automated email for 2SV enabled will now be "2-step verification set up" instead of "[Production] 2-step verification set up". It's not clear to me why this particular email is singled out in this way and unfortunately the [commit note][3] & [PR][4] where it was originally introduced doesn't shed any light on the matter. I've written up [this Trello card](https://trello.com/c/0uCDE6Kd/108-should-the-subject-line-of-all-automated-emails-have-an-env-specific-prefix) to consider using the same prefix for all automated emails.

In production, the subject line of the automated email for batches of
user invitations will now be:

> [SIGNON] Serena Williams created a batch of 37 users

instead of:

> [SIGNON] Serena Williams created a batch of 37 users in production

### Email content

In production, the content for the following automated email types will no longer include content which was originally only intended to be displayed in non-production environments. This will resolve the specific issue which drew our attention to the bug.
    * Suspension reminder
    * Suspension notification
    * Notify reset password disallowed due to suspension
    * 2SV enabled
    * Unlock instructions

In production, automated emails will now refer to "your GOV.UK Signon account" rather than "your production GOV.UK account".

### Email from address

The from address in all automated emails sent from the production environment will now be:

> GOV.UK Signon <noreply-signon@digital.cabinet-office.gov.uk>

versus:

> GOV.UK Signon production <noreply-signon-production@digital.cabinet-office.gov.uk>

Note that the reply-to address will not change, so there should NOT be any need to check that we can actually receive emails at the new from address.

## Suggested further work

* Consider using `GOVUK_ENVIRONMENT` or `GOVUK_ENVIRONMENT_NAME` which [both seem](https://github.com/alphagov/govuk-helm-charts/blob/4b330aa6e254f60d771b1452f85237ff8f8a1c02/charts/app-config/templates/env-configmap.yaml#L21-L22) to make the relevant environment-specific strings available to *ALL* GOV.UK apps. We could then remove `INSTANCE_NAME` altogether.
    * Note that we're already using `GOVUK_ENVIRONMENT_NAME` for [something similar](https://github.com/alphagov/signon/blob/db485c293839257e737b05b3c9b8fe5d3f45a5ef/config/initializers/govuk_admin_template.rb#L9-L10) and for [something else similar](https://github.com/alphagov/signon/blob/db485c293839257e737b05b3c9b8fe5d3f45a5ef/lib/healthcheck/api_tokens.rb#L22).
* Document all environment variables - I don't know if there is a recommended way to do this?
* Contact the team that's responsible for the Short URL Manager app which seems to be the only other app that uses an `INSTANCE_NAME` env var in case there is a similar bug in that app.

[1]: https://github.com/alphagov/govuk-helm-charts/blob/d9a542872a009363f9251d8276e98f1c94c6cd16/charts/app-config/values-production.yaml#L2289-L2290
[2]: https://github.com/alphagov/govuk-puppet/pull/7432
[3]: https://github.com/alphagov/signon/commit/89d2801d54afc778663bc4703c30cf47d3b1e3b5
[4]: https://github.com/alphagov/signon/pull/425

### DIFF
--- a/app/helpers/two_step_verification_helper.rb
+++ b/app/helpers/two_step_verification_helper.rb
@@ -28,8 +28,8 @@ private
 
   def otp_secret_key_uri(user:, otp_secret_key:)
     issuer = I18n.t("devise.issuer")
-    if Rails.application.config.instance_name
-      issuer = "#{Rails.application.config.instance_name.titleize} #{issuer}"
+    if GovukEnvironment.name
+      issuer = "#{GovukEnvironment.name.titleize} #{issuer}"
     end
 
     issuer = ERB::Util.url_encode(issuer)
--- a/app/helpers/two_step_verification_helper.rb
+++ b/app/helpers/two_step_verification_helper.rb
@@ -28,7 +28,7 @@ private
 
   def otp_secret_key_uri(user:, otp_secret_key:)
     issuer = I18n.t("devise.issuer")
-    if GovukEnvironment.name
+    unless GovukEnvironment.production?
       issuer = "#{GovukEnvironment.name.titleize} #{issuer}"
     end
 
--- a/app/mailers/mailer_helper.rb
+++ b/app/mailers/mailer_helper.rb
@@ -8,10 +8,10 @@ module MailerHelper
   end
 
   def email_from_address
-    if GovukEnvironment.name.present?
-      I18n.t("mailer.email_from_address.instance", instance_name: GovukEnvironment.name.parameterize)
-    else
+    if GovukEnvironment.production?
       I18n.t("mailer.email_from_address.no-instance")
+    else
+      I18n.t("mailer.email_from_address.instance", instance_name: GovukEnvironment.name.parameterize)
     end
   end
 end
--- a/app/mailers/mailer_helper.rb
+++ b/app/mailers/mailer_helper.rb
@@ -16,6 +16,6 @@ module MailerHelper
   end
 
   def instance_name
-    Rails.application.config.instance_name
+    GovukEnvironment.name
   end
 end
--- a/app/mailers/mailer_helper.rb
+++ b/app/mailers/mailer_helper.rb
@@ -4,7 +4,11 @@ module MailerHelper
   end
 
   def app_name
-    I18n.t("mailer.app_name", instance_name: GovukEnvironment.name)
+    if GovukEnvironment.production?
+      I18n.t("mailer.app_name.no_instance")
+    else
+      I18n.t("mailer.app_name.instance", instance_name: GovukEnvironment.name)
+    end
   end
 
   def email_from_address
--- a/app/mailers/mailer_helper.rb
+++ b/app/mailers/mailer_helper.rb
@@ -4,18 +4,14 @@ module MailerHelper
   end
 
   def app_name
-    I18n.t("mailer.app_name", instance_name:)
+    I18n.t("mailer.app_name", instance_name: GovukEnvironment.name)
   end
 
   def email_from_address
-    if instance_name.present?
-      I18n.t("mailer.email_from_address.instance", instance_name: instance_name.parameterize)
+    if GovukEnvironment.name.present?
+      I18n.t("mailer.email_from_address.instance", instance_name: GovukEnvironment.name.parameterize)
     else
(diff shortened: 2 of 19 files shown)


mike29736 (approved):
Nicely done.

So is there something stopping us from switching to `GOVUK_ENVIRONMENT` or `GOVUK_ENVIRONMENT_NAME` right now?

floehopper:
> Nicely done.
> 
> So is there something stopping us from switching to `GOVUK_ENVIRONMENT` or `GOVUK_ENVIRONMENT_NAME` right now?

No, I don't think so - I just thought I'd do it as a separate PR.

floehopper:
> The issuer in the 2SV QR code will now be set to "GOV.UK Signon" rather than "Production GOV.UK Signon" in production. The issuer is used by Authenticator Apps to identify which website the secret belongs to. There might be some confusion for people who originally setup 2SV while this bug existed who now come back to reset 2SV, but I plan to investigate that before I merge the changes.

I've just tested this using [this Authentication browser extension](https://authenticator.cc/) and I could see that there was no way to edit/update an existing entry - you have to delete the old one and add the new one - I suspect that's true for all authenticator apps. While I think there is _some_ possibility for confusion for users who have setup 2SV before the bug is fixed and then come to reset 2SV after the bug is fixed, I'm not sure it's super confusing. So I think it's better to fix the bug and make the app consistent.

floehopper:
> Users promoted to any admin role in production will now be required to setup 2SV. Since we now mandate 2SV for most users in production, this is unlikely to be causing a security lapse, but it would be worth checking that no users have slipped through the cracks while this bug existed.

There are only 3 users in production that have one of the 4 admin roles and have `User#require_2sv` set to `false`:

```ruby
users = User.where(role: %i(admin superadmin organisation_admin super_organisation_admin), require_2sv: false)
users.count # => 3
```

And all three of them have 2SV mandated via their organisation and all three of them already have it set up:

```ruby
users.map(&:require_2sv) # => [false, false, false]
users.map { |u| u.organisation.require_2sv? } # => [true, true, true]
users.map(&:exempt_from_2sv?) # => [false, false, false]
users.map(&:require_2sv?) # => [true, true, true]
users.map(&:has_2sv?) # => [true, true, true]
```

So I think we're all good! 😅

PR #2312, opened 2023-08-16
Enable OAuth2 PKCE extension in Doorkeeper

chrisroos:
https://trello.com/c/59EBweBx

Signon utilises the [OAuth 2.0 Authorization Code Grant type][1] to enable users to authorize client applications to act on their behalf. That page includes:

> It is recommended that all clients use the [PKCE](https://oauth.net/2/pkce/) extension with this flow as well to provide better security.

This PR enables the PKCE extension in Doorkeeper. We're not yet mandating Signon clients to use PKCE but it will be used if the client supplies a `code_challenge` and `code_challenge_method` in the request for the Authorization Code.

We hope to mandate the use of PKCE once we're confident all Signon clients are using it.

[1]: https://oauth.net/2/grant-types/authorization-code/

### DIFF
--- a/db/migrate/20230814160809_enable_pkce.rb
+++ b/db/migrate/20230814160809_enable_pkce.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class EnablePkce < ActiveRecord::Migration[7.0]
+  def change
+    change_table :oauth_access_grants, bulk: true do |t|
+      t.column :code_challenge, :string, null: true
+      t.column :code_challenge_method, :string, null: true
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -81,6 +81,8 @@ ActiveRecord::Schema[7.0].define(version: 2023_08_16_164936) do
     t.datetime "created_at", precision: nil, null: false
     t.datetime "revoked_at", precision: nil
     t.string "scopes", default: "", null: false
+    t.string "code_challenge"
+    t.string "code_challenge_method"
     t.index ["token"], name: "index_oauth_access_grants_on_token", unique: true
   end
 
--- a/test/integration/doorkeeper_integration_test.rb
+++ b/test/integration/doorkeeper_integration_test.rb
@@ -6,4 +6,58 @@ class DoorkeeperIntegrationTest < ActionDispatch::IntegrationTest
   rescue ActionController::RoutingError => e
     assert_equal 'No route matches [GET] "/oauth/applications"', e.message
   end
+
+  test "oauth2 authorization code grant type" do
+    without_csrf_protection do
+      app = create(:application)
+      password = SecureRandom.urlsafe_base64
+      user = create(:user, with_signin_permissions_for: [app], password:)
+
+      sign_in user.email, password
+
+      auth_code = request_authorization_code(app)
+
+      reset!
+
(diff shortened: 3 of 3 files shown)


floehopper (approved):
This looks good to me 👍  I have a few thoughts...

* I don't feel strongly about keeping CSRF enabled for the tests. Since `Capybara::DSL` is mixed into `ActionDispatch::IntegrationTest` would it be feasible/sensible to use those for the user bit of the interaction (i.e. signing in) thus coping with CSRF...? Or is there a problem that Capybara will follow redirects you don't want to...?
* I suppose ideally you'd use different sessions, although I have to admit I'm only 99% sure that's actually how things work! Having said that, I'm not sure it's worth putting a lot of effort into it. Would using a mixture of Capybara & Rails request helpers (as described in my previous point) achieve separate sessions? If not, could you not just call [`ActionDispatch::Integration::Session#reset!`](https://api.rubyonrails.org/classes/ActionDispatch/Integration/Session.html#method-i-reset-21) in the test after signing in and before requesting the authorization code...?
* These [Doorkeeper docs](https://doorkeeper.gitbook.io/guides/ruby-on-rails/pkce-flow#enable-pkce-in-doorkeeper) mention checking a couple of things when enabling PKCE - it would be nice to mention somewhere that you *have* checked them even if they don't apply.

I'm happy to leave all the above to your discretion - essentially it'd be fine to merge as is!

Lastly, don't forget to remove the TODO points from the commit note before you merge!

chrisroos:
Thanks, @floehopper :+1: 

> This looks good to me 👍 I have a few thoughts...
> 
> - I don't feel strongly about keeping CSRF enabled for the tests. Since Capybara::DSL is mixed into ActionDispatch::IntegrationTest would it be feasible/sensible to use those for the user bit of the interaction (i.e. signing in) thus coping with CSRF...? Or is there a problem that Capybara will follow redirects you don't want to...?

I haven't tried but I worry that it might be confusing to mix the two styles in a single test. Especially as I'd prefer to more explicitly split the integration-style and the system-style tests that already make up our "integration tests".

> - I suppose ideally you'd use different sessions, although I have to admit I'm only 99% sure that's actually how things work! Having said that, I'm not sure it's worth putting a lot of effort into it. Would using a mixture of Capybara & Rails request helpers (as described in my previous point) achieve separate sessions? If not, could you not just call [ActionDispatch::Integration::Session#reset!](https://api.rubyonrails.org/classes/ActionDispatch/Integration/Session.html#method-i-reset-21) in the test after signing in and before requesting the authorization code...?

I've updated the commit to use `reset!` as you've suggested. I'm not entirely sure how much it helps but hopefully it gives some clue that the exchange of an Authorization Code for an Access Token, and the use of the Access Token to request a protected resource, are triggered from the client and don't require the user to be signed in to Signon.

> - These [Doorkeeper docs](https://doorkeeper.gitbook.io/guides/ruby-on-rails/pkce-flow#enable-pkce-in-doorkeeper) mention checking a couple of things when enabling PKCE - it would be nice to mention somewhere that you have checked them even if they don't apply.

I've updated the commit note to explain why these don't concern us.

> I'm happy to leave all the above to your discretion - essentially it'd be fine to merge as is!

Thank you!

> Lastly, don't forget to remove the TODO points from the commit note before you merge!

Done.

PR #2321, opened 2023-08-23
Publish /privacy-notice page

mike29736:
https://trello.com/c/AgYOfi7y/239-publish-the-signon-privacy-notice-and-link-to-it-in-signon-page

To meet one of the requirements of our Data Protection Impact Assessment, we're publishing this privacy notice and cookie policy for users of Signon.

I've had one eye on the styling of https://www.gov.uk/help/privacy-notice while working on this. Though, since that's a public facing page, I don't really know how similar this needs to be :confused: 

There's a suggestion on the Trello card about explaining what the page is for when linking to it. I haven't found an appropriate place to do that, since I've followed the convention of just including the link in the site's footer.

Things I'm still wondering:
- should the page heading have fewer capital letters in it?
- should the URL and footer link text include "cookie policy" in them?

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,6 +1,9 @@
 @import "govuk_publishing_components/all_components";
 @import "user_research_recruitment_banner";
 
+// Components from this application
+@import "components/*";
+
 // TODO: move into component
 .gem-c-success-alert,
 .gem-c-error-alert {
--- a/app/assets/stylesheets/components/_contact-details.scss
+++ b/app/assets/stylesheets/components/_contact-details.scss
@@ -1,0 +1,13 @@
+.app-c-contact-details {
+  @include govuk-font($size: 19);
+  @include govuk-text-colour;
+  padding-left: govuk-spacing(3);
+  // Margin top intended to collapse
+  // This adds an additional 10px to the paragraph above
+  @include govuk-responsive-margin(6, "top");
+  @include govuk-responsive-margin(6, "bottom");
+
+  clear: both;
+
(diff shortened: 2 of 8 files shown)


chrisroos (commented):
Hi @mike29736.

Should the Privacy notice be publicly accessible? It feels slightly odd to me that it's linked to from the footer of the sign in page but that users have to sign in to be able to view it.

I'm not seeing the different `page_heading` and `title` when I view this page locally. I see "Privacy notice and cookie policy" in the browser tab title and on the page itself, which doesn't seem to be what you'd intended.

image

Do we need to include information about other cookies (e.g. the `dismiss_user_research_recruitment_banner`)?

mike29736:
> Hi @mike29736.
> 
> Should the Privacy notice be publicly accessible? It feels slightly odd to me that it's linked to from the footer of the sign in page but that users have to sign in to be able to view it.
> 
> I'm not seeing the different `page_heading` and `title` when I view this page locally. I see "Privacy notice and cookie policy" in the browser tab title and on the page itself, which doesn't seem to be what you'd intended.
> 
> image
> 
> Do we need to include information about other cookies (e.g. the `dismiss_user_research_recruitment_banner`)?

Thanks @chrisroos,

1. You're right, I'll make it public.

2. My only idea about the `page_heading`/`title` issue is _something something_ asset compilation. But the layout isn't an asset, so for now I'm stumped.

3. I'll add the banner cookie to the table and run it past Nick. I completely forgot to write this down when it came up yesterday :grimacing:

chrisroos (approved):
Thanks @mike29736 :-) This all looks good to me :+1:

PR #2323, opened 2023-08-24
Restructure asset directories

floehopper:
This is mainly about separating legacy JS & CSS from non-legacy JS & CSS, but also moving to a similar structure as used in other apps that makes it easier to e.g. add new non-legacy JS modules.

I think it _might_ also make sense to move `app/assets/stylesheets/_user_research_recruitment_banner.scss` into a new `app/assets/stylesheets/components` or `app/assets/stylesheets/views` directory, but I haven't worked out what the difference is between those two yet, so I've left it for now.


chrisroos (approved):
Looks good to me, @floehopper.

I didn't initially understand what "legacy" meant in the context of this PR but James explained that it's anything related to the GovukAdminTemplate that we're in the process of removing.

I ran `RAILS_ENV=production rails assets:precompile` locally and it completed successfully so I'm assuming this all works as expected.

PR #2324, opened 2023-08-24
Add sequence diagram to OAuth docs

floehopper:
Trello: https://trello.com/c/FmtE3aLc

This shows the sequence of HTTP requests that are made when a non-signed-in user visits Whitehall Publisher and is redirected to Signon to sign in before finally being returned to Whitehall Publisher.


chrisroos (approved):
Looks good to me, @floehopper!

PR #2330, opened 2023-08-29
Require user to sign in after accepting invitation

mike29736:
https://trello.com/c/SWVUwQHc/101-consider-requiring-users-to-sign-in-after-accepting-invitation

To mitigate the security risk posed by leaked invitation tokens, we're
no longer automatically signing a user in when they accept an
invitation.

This is a feature of Devise's Invitable module, so not a lot of work on
our part (to enable or disable).

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -116,6 +116,11 @@ Devise.setup do |config|
   # Default: false
   config.validate_on_invite = true
 
+  # Auto-login after the user accepts the invite. If this is false,
+  # the user will need to manually log in after accepting the invite.
+  # Default: true
+  config.allow_insecure_sign_in_after_accept = false
+
   # ==> Configuration for :confirmable
   # A period that the user is allowed to access the website even without
   # confirming their account. For instance, if set to 2.days, the user will be
--- a/test/integration/inviting_users_test.rb
+++ b/test/integration/inviting_users_test.rb
@@ -3,7 +3,7 @@ require "test_helper"
 class InvitingUsersTest < ActionDispatch::IntegrationTest
   include ActiveJob::TestHelper
 
-  should "send the user an invitation token" do
+  should "ask the invited user to set a password" do
     user = User.invite!(name: "Jim", email: "jim@web.com")
     visit accept_user_invitation_path(invitation_token: user.raw_invitation_token)
 
--- a/test/integration/inviting_users_test.rb
+++ b/test/integration/inviting_users_test.rb
@@ -11,7 +11,24 @@ class InvitingUsersTest < ActionDispatch::IntegrationTest
     fill_in "Confirm new password", with: "this 1s 4 v3333ry s3cur3 p4ssw0rd.!Z"
     click_button "Save password"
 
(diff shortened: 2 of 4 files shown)


test/integration/inviting_users_test.rb
```diff
@@ -3,7 +3,7 @@
 class InvitingUsersTest < ActionDispatch::IntegrationTest
   include ActiveJob::TestHelper
 
-  should "send the user an invitation token" do
+  should "ask the invited user to set a password" do
```
floehopper, line 6: Good catch! 😬

test/integration/inviting_users_test.rb
```diff
@@ -11,7 +11,24 @@ class InvitingUsersTest < ActionDispatch::IntegrationTest
...
+
+  should "require the invited user to sign in after setting their password" do
+    user = User.invite!(name: "Neptuno Keighley", email: "neptuno.keighley@office.gov.uk")
+
+    accept_invitation(
+      invitation_token: user.raw_invitation_token,
+      password: "pretext annoying headpiece waviness header slinky",
+    )
+
+    assert_response_contains("Sign in to GOV.UK")
```
floehopper, line 25: Does this page make it clear that you now need to sign in? Perhaps it's obvious. I mainly want to make sure we're not going to introduce a whole new class of support tickets.

config/initializers/devise.rb
```diff
@@ -116,6 +116,11 @@
   # Default: false
   config.validate_on_invite = true
 
+  # Auto-login after the user accepts the invite. If this is false,
+  # the user will need to manually log in after accepting the invite.
+  # Default: true
+  config.allow_insecure_sign_in_after_accept = false
+
```
floehopper, line 123: Nice!

This makes me wonder 🤔  if we're missing any other configuration options from newer versions of Devise, but that's definitely out-of-scope for this PR!

floehopper (approved):
This looks good to me - well done on finding the Devise configuration option!

My only slight reservation is about making best efforts to ensure the new UX doesn't cause a rise in support tickets. I haven't actually tried the workflow out myself, so I might well be making a mountain out of a mole hill!

Anyway, I'm happy to leave the above to your discretion, so I'm going to approve the PR.

config/initializers/devise.rb
mike29736: I think it'd be worth regenerating that file just for the improved explanations of some of the config options in the latest version. Anyhow: https://trello.com/c/ARpxcKEA/131-update-signons-devise-and-extensions-config

test/integration/inviting_users_test.rb
mike29736: I'll ask the team for a second opinion before I merge it.

But here's the page you land on from the invite link:

Screenshot from 2023-08-30 10-09-27

And here's the next page:

Screenshot from 2023-08-30 10-10-03

floehopper: Ah, nice! I reckon it's fine, but probably worth a 2nd opinion.

PR #2331, opened 2023-08-29
Avoid queueing job if Splunk-related functionality is disabled

floehopper:
Trello: https://trello.com/c/q1B1HKBy

Currently the Splunk-related functionality is effectively disabled, because neither of the relevant env vars are defined in any environment. See [this Trello card][1] for more details.

Prior to this commit, we were always queueing up a Sidekiq job whenever `EventLog.record_event` was called even though, as soon as that `SplunkLogStreamingJob` was executed, the first thing we did was give up if both the relevant env vars were not set.

Now we only queue up the job in the first place if the relevant env vars are defined. I've left the original check in `EventLog#send_to_splunk` in place, because there's a small chance the state of the env vars could change between the job being queued and the job being executed.

[1]: https://trello.com/c/Zp23Kd6c

### DIFF
--- a/test/factories/event_log.rb
+++ b/test/factories/event_log.rb
@@ -1,3 +1,6 @@
 FactoryBot.define do
-  factory :event_log
+  factory :event_log do
+    event_id { EventLog::NO_SUCH_ACCOUNT_LOGIN.id }
+    uid { create(:user).uid }
+  end
 end
--- a/app/models/event_log.rb
+++ b/app/models/event_log.rb
@@ -119,7 +119,9 @@ class EventLog < ApplicationRecord
 
     event_log_entry = EventLog.create!(attributes)
 
-    SplunkLogStreamingJob.perform_later(event_log_entry.id)
+    if splunk_endpoint_enabled?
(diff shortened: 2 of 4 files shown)


test/factories/event_log.rb
```diff
@@ -1,3 +1,6 @@
 FactoryBot.define do
-  factory :event_log
+  factory :event_log do
+    event_id { EventLog::NO_SUCH_ACCOUNT_LOGIN.id }
+    uid { create(:user).id }
```
mike29736, line 4: It'd be more accurate (as documentation) to use the user's `uid` instead of the integer `id` here

mike29736 (approved):
:+1:

test/factories/event_log.rb
floehopper: Ah - good idea! 👍

floehopper: Addressed in https://github.com/alphagov/signon/pull/2331/commits/4405fad9436d5eeb62412e25197e027d1701f32a

PR #2334, opened 2023-08-30
Add a page listing the current user's applications

chrisroos:
Trello: https://trello.com/c/7Hfe7eu0

This is the first step toward the new UI for allowing users to manage their own application permissions. It adds a page at `/account/applications` that lists the apps that a user does and doesn't have access to (i.e. those with and without the signin permission). We've chosen not to include retired apps as we don't think it makes sense for a user to be changing their permissions for retired apps.

The new applications page is currently only accessible to GOV.UK Admins (superadmins and admins). The intention is to make it available to Publishing Managers too but I'd prefer to get it working for the simpler case before tackling the complexity of Publishing Managers permissions.

Although not strictly necessary I've also chosen to add the `/account` path which currently redirects to either the user's edit page or the user's change email/password page depending on whether the user has permission to access the edit page. I'm imagining that at some point this resource will become a page containing information about the user and allowing them to edit various aspects of their account.

---

image

### DIFF
--- a/app/controllers/account/applications_controller.rb
+++ b/app/controllers/account/applications_controller.rb
@@ -6,6 +6,7 @@ class Account::ApplicationsController < ApplicationController
   def index
     authorize :account_applications
 
-    @applications = Doorkeeper::Application.can_signin(current_user)
+    @applications_with_signin = Doorkeeper::Application.can_signin(current_user)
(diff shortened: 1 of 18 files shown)


chrislo (approved):
Looks good to me @chrisroos 👍

PR #2344, opened 2023-09-06
Remove <abbr> tag from API Users table

mike29736:
https://trello.com/c/dggyhTVY/62-api-users-page-fix-and-iterate-how-this-page-is-displayed

This tooltip was on our list of things that don't seem like they'd
comply with the design system.

So far I haven't found anybody who uses this feature.

Used to look like this:
Screenshot from 2023-09-06 16-19-31

Now it doesn't:
Screenshot from 2023-09-06 16-20-51


chrisroos (approved):
Looks good to me, @mike29736 :+1:

PR #2345, opened 2023-09-06
Move users index page to design system

floehopper:
Trello: https://trello.com/c/2gvEJKw9 & https://trello.com/c/bXsVUK6c

## Summary

The main motivation for this work is to convert the users index page to use the GOV.UK Design System instead of [govuk_admin_template](https://github.com/alphagov/govuk_admin_template) and improve the accessibility & responsiveness of the page.

## Description

The new version of the page adds checkbox filters based on the [option-select](https://govuk-finder-frontend.herokuapp.com/component-guide/option_select) component from finder-frontend (e.g. [Research and statistics](https://www.gov.uk/search/research-and-statistics)) in a sidebar.

The new filters allow users to select *multiple* values per filter whereas previously it was only possible to select a single value per filter. In order to implement this new behaviour I've refactored/reimplemented a bunch of scopes & methods on the `User` model. In doing so I hope they're now more idiomatic and could possible already be used elsewhere to make the code more intention-revealing. URLs with query strings generated from the old filters will be redirected to query strings compatible with the new filters - this means that any bookmarked URLs should still work.

A bit like the old version of the page, the "Organisations" filter is not displayed if the current user can only "manage" a single organisation (i.e. Organisation Admins); instead the table caption is amended to make it clear the users in the table are all associated with a given organisation.

We're using JS to auto-submit the filter form when checkboxes are checked/unchecked (triggering a full page reload) rather than implementing a fully-Ajaxified solution like that in the finder-frontend page(s). In order to reduce any potential confusion caused by the full page reload we're doing the following:
* "open" any of the filters which have checked checkboxes
* list any checked checkboxes at the top of each filter to make them easier to find to uncheck
* "open" the top ("Status") filter if no filters have any checkboxes checked

We chose to add the JS behaviour described above, because the "Update results" button can easily disappear off the bottom of the browser viewport and, in any case, it's very similar to how the old version of the page worked. The page should also work fine without JS enabled - it's just that the filters are all always "open" and the user will have to click the "Update results" button to apply the filters. This feels like a siginificant improvement on the old version of the page which effectively had two fairly distinct implementations (one when JS was enabled; and one when it was not).

The introduction of the filters sidebar has significantly reduced the page width available for the table which was already overflowing the main container on screens narrower than "desktop" size. In order to avoid this overflow problem, we've had to make some compromises on what data is displayed in the table: we've removed the "Actions" column which used to display "Unlock account" & "Resend signup email" buttons - these actions are still available from the user edit page; and we've removed the "Organisation" and "Last sign in" columns.

To further address this overflow problem and to improve the "responsiveness" of the page, we've also made the following changes:
* Increased the `max-width` of `govuk-width-container` for all pages in the app following [a precendent set in whitehall](https://github.com/alphagov/whitehall/blob/b355ef2caf4b414487d3687495d133b66f7f7846/app/assets/stylesheets/application.scss#L1)
* Used @mike29736's new `vertical_on_small_screen` variation of the table component to "pivot" the table on narrow screens
* Increased the breakpoint used by that table component so that the table changes into "vertical" mode before the table content overflows the right-hand side of the container and before a scrollbar appears at the bottom of the page

The "Export users as CSV" button has moved to the bottom of the filters sidebar and its now styled as a link. The "Upload a batch of users" & "Grant access to all users" buttons have become secondary buttons.

## To do

* [x] Add before & after screenshots to this PR description and/or the Trello card(s). I've been using temporary CSS styles like `filter: blur(4px)` to blur the PII columns ("Name" & "Email") in screenshots that appear in public forums like PRs.
* [x] Write some Jasmine and/or JS-enabled integration tests to check the auto-submit JS behaviour, e.g. some along the lines of [these tests](https://github.com/alphagov/signon/blob/35621cd731495122e65fdf7452756c3185a0513d/test/integration/admin_user_index_test.rb#L48-L172). I ran into a problem with a JS-enabled integration test where Capybara/Selenium-Webdriver seemed to consider the filter checkboxes as not "visible" even though a screenshot seemed to suggest they should be visible. ~This is a bit of a concern because it might be an indication that some screen-readers might also have problems.~ I'll add some notes about where I got to with this below. **[Update: I got to the bottom of this - see [this comment](https://github.com/alphagov/signon/pull/2345#issuecomment-1709683484)]**
* [ ] Flag the changes to the Support team in case they see a flurry of tickets.
* [ ] Consider landing some of the commits separately as a precursor to this PR, e.g.
    * "Extract User#role_class method"
    * "Encapsulate more of UserPolicy#can_manage? in User#can_manage?"
    * "Inline Role::Base#can_manage? into User#can_manage?"
    * "Introduce User#manageable_organisations method"
* [x] Check the order of the checkboxes in the "Roles" filter is consistent for current users with different roles. Unfortunately `Superadmin.manageable_roles` is implemented differently than in the other role classes and it returns roles in the opposite order. It would be nice to make all the implementations of `.manageable_roles` return roles in the same order, perhaps making use of the `level` attribute...?
* [ ] Possibly make the values in the "Email" column of the table bold as per the [original Figma design](https://www.figma.com/file/ArJZK2wjwRwIHJItpRTGjq/GOV.UK-Publishing-SignOn---June-2023?type=design&node-id=247-13459&mode=design&t=6eVzWKSt0f2NH8a2-0). Although I'm not sure this will play well with the strikethrough the replacement of which I've extracted into [this Trello card](https://trello.com/c/a5XHI9NS).

## Notes on the problem with JS-enabled integration tests

* I was able to enable JS by calling `use_javascript_driver` in the test. Note that calling this resets the session, so you may need to call it in the test setup if that's where the user signs in.
* With the JS driver enabled, I was able to capture screenshots using `save_screenshot(Rails.root.join("screenshot.png"))`. By default `save_screenshot()` saves the screenshot in the project `tmp` directory which is not accessible from outside the `govuk-docker` container by default. Saving the screenshot to the project root directory means that it is accessible.
* I found it easier to see what was going in these screenshots on by increasing the browser window dimensions using e.g. `current_window.resize_to(1920, 1080)`.
* With JS enabled the test successfully finds the checkboxes but only if we include invisible elements, e.g. `check "Active", visible: false`.
* I was able to enable debug-level logging from Selenium-Webdriver using `Selenium::WebDriver.logger.level = :debug; Selenium::WebDriver.logger.output = Rails.root.join("selenium.log")`.
* From this log I could see that finding the checkbox element including invisible elements generated the following XPath query: `./descendant::input[(./@type = 'checkbox')][((((./@id = 'Active') or (./@name = 'Active')) or (./@placeholder = 'Active')) or (./@id = //label[contains(normalize-space(string(.)), 'Active')]/@for))] | .//label[contains(normalize-space(string(.)), 'Active')]//./descendant::input[(./@type = 'checkbox')]` and this returned a value like this: `[{"element-6066-11e4-a52e-4f735466cecf":"849902D65C2D92EADB74FC56B1B7247E_element_22"}]`.
* When querying the visibility of the checkbox, I could see that Selenium was executing a complex and minified/obfuscated JS function called `isDisplayed` and it returned a value of `false`. This seems to correspond to [this documentation](https://www.selenium.dev/documentation/webdriver/elements/information/#is-displayed) and _possibly_ [this function](https://github.com/SeleniumHQ/selenium/blob/de4d2d56390cc17c37a8ce7c31f315e0f95338b9/rb/lib/selenium/webdriver/atoms/isDisplayed.js). The complex and minified/obfuscated nature of the function meant that I didn't take this any further, but it might be worth trying to invoke the function in a Chrome browser...?
* As a separate exercise, I confirmed that the test could successfully check a simple checkbox added near the top of the page. The `option-select` component uses the `checkboxes` component and thence the `checkbox` component to render the individual checkboxes in the filters. I tried changing the `option-select` component to generate the checkboxes directly using HTML/ERB and the test successfully checked these. So I think the problem must lie somewhere in the `checkboxes` and/or `checkbox` component. However, that's when I ran out of time - sorry!


floehopper:
@chrisroos:

I've worked out why Capybara can't see the checkboxes. The `input` elements have [an `opacity` of 0][1] which means they are completely transparent.

The visual box is generated by [a `before` pseudo-element][2] on the `label` element and the tick mark is generated by [an `after` pseudo-element][3] on the `label` element. The transparent `input` element is positioned directly "above" these using [a `z-index` of 1][4].

You can see the `input` if you add the following styles:

```scss
.govuk-checkboxes__input {
  opacity: 1;
  position: relative;
}
```

<p>Screenshot 2023-09-07 at 08 11 31</p>

One solution for a JS-enabled integration test  is to "click" the `label` element using `find("label", text: "Active").click` instead of "checking" the `input` element using `check("Active")` which fails because the element is transparent and therefore treated as "invisible".

As an aside: I haven't investigated this completely, but I'm pretty sure if assets have previously been precompiled (e.g. by having just run a test), I don't think they are re-compiled when you re-run the test. So if you make a change to an asset, it's worth running `rails assets:clobber` before running the test again.

[1]: https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss#L44
[2]: https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss#L58-L69
[3]: https://github.com/alphagov/govuk-frontend/blob/432c8fc14142a91c6b0720dabf0893b334503d8a/packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss#L71-L95
[4]: https://github.com/alphagov/govuk-frontend/blob/432c8fc14142a91c6b0720dabf0893b334503d8a/packages/govuk-frontend/src/govuk/components/checkboxes/_index.scss#L36

floehopper:
@chrisroos:

> Write some Jasmine and/or JS-enabled integration tests to check the auto-submit JS behaviour, e.g. some along the lines of these tests.

I recently came across [this gem][1] which might make writing assertions against the table of users in an integration test much easier/readable.

[1]: https://github.com/jnicklas/capybara_table

floehopper:
@chrisroos:

I just came across the ‘allow_label_click’ option for ‘Capybara::Node::Actions#check’ which might be a better option than finding and clicking the label…

There’s also an ‘automatic_label_click’ configuration option (false by default) which might be worth setting to true…

chrisroos:
> Add before & after screenshots to this PR description and/or the Trello card(s). I've been using temporary CSS styles like filter: blur(4px) to blur the PII columns ("Name" & "Email") in screenshots that appear in public forums like PRs.

I've created a Google Doc of before/after screenshots and linked to them from the Trello card rather.

chrisroos:
> Write some Jasmine and/or JS-enabled integration tests to check the auto-submit JS behaviour, e.g. some along the lines of [these tests](https://github.com/alphagov/signon/blob/35621cd731495122e65fdf7452756c3185a0513d/test/integration/admin_user_index_test.rb#L48-L172).

I've added JavaScript enabled integration tests to test the JavaScript filter functionality.

chrisroos:
> I just came across the ‘allow_label_click’ option for ‘Capybara::Node::Actions#check’ which might be a better option than finding and clicking the label…

Thanks, @floehopper! I'm using this in the integration tests I've added.

chrisroos:
> Check the order of the checkboxes in the "Roles" filter is consistent for current users with different roles. Unfortunately Superadmin.manageable_roles is implemented differently than in the other role classes and it returns roles in the opposite order. It would be nice to make all the implementations of .manageable_roles return roles in the same order, perhaps making use of the level attribute...?

I've confirmed that the order of the Roles filter is the same before/after this PR for the different user roles.

chrisroos (approved):
I've read through the changes in this PR a couple of times, added some JS-enabled tests and played around with the functionality in development. It all looks good to me! Good work, @floehopper!

chrisroos:
I've merged this in @floehopper's absence.

chrisroos:
> Consider landing some of the commits separately as a precursor to this PR

I decided not to do this because it would've taken longer to extract them than it would've done to merge them all in together.

chrisroos:
> Possibly make the values in the "Email" column of the table bold as per the [original Figma design](https://www.figma.com/file/ArJZK2wjwRwIHJItpRTGjq/GOV.UK-Publishing-SignOn---June-2023?type=design&node-id=247-13459&mode=design&t=6eVzWKSt0f2NH8a2-0). Although I'm not sure this will play well with the strikethrough the replacement of which I've extracted into [this Trello card](https://trello.com/c/a5XHI9NS).

I decided not to do this. If it comes up again then we can make the change.

PR #2346, opened 2023-09-07
Improve db seeds

chrisroos:
This PR adds the following users to db/seeds to complement the existing test.admin@gov.uk user:

- test.superadmin@gov.uk
- test.super-organisation-admin@gov.uk
- test.organisation-admin@gov.uk

I found the addition of these users useful while playing with the app in development.

I've also removed some duplication by using `Role::<role>#role_name` instead of hardcoded strings.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -8,10 +8,37 @@ gds = Organisation.create!(
 )
 
 User.create!(
+  name: "Test Superadmin",
+  email: "test.superadmin@gov.uk",
+  password: "6fe552ca-d406-4c54-b7a6-041ed1ade6cd",
+  role: :superadmin,
+  confirmed_at: Time.zone.now,
+  organisation: gds,
+)
+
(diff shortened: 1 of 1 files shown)


chrislo:
Looks good to me @chrisroos. I wonder if we should use `@example.com` domain names rather than `@gov.uk` for these users (I appreciate the existing one already used the latter so it's not something you introduced) - just in case they happen to exist and we end up sending an email to one of them?

chrislo (approved):

chrisroos:
> Looks good to me @chrisroos. I wonder if we should use `@example.com` domain names rather than `@gov.uk` for these users (I appreciate the existing one already used the latter so it's not something you introduced) - just in case they happen to exist and we end up sending an email to one of them?

Thanks, @chrislo :+1: 

Using example.com is probably a good shout but I'm not going to do that as part of this PR.

PR #2350, opened 2023-09-08
Break batch invitations creation into two steps

mike29736:
https://trello.com/c/FzaiUjkY/127-break-batchinvitations-new-into-two-separate-steps-to-prepare-for-migrating-to-the-design-system

This is the first step of moving `batch_invitations#new` to the design system. It breaks the large single page form into two separate steps across their own pages:
  1. Upload user info
  2. Manage permissions

The second step is implemented in its own controller.

## Step 1.
Screenshot from 2023-09-08 20-25-48

## Step 2.
Screenshot from 2023-09-08 20-26-37

### DIFF
--- a/app/controllers/batch_invitation_permissions_controller.rb
+++ b/app/controllers/batch_invitation_permissions_controller.rb
@@ -26,7 +26,7 @@ private
   end
 
   def authorise_to_manage_permissions
-    authorize @batch_invitation
+    authorize @batch_invitation, :manage_permissions?
   end
 
   def grant_default_permissions(batch_invitation)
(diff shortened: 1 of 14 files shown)


chrislo:
I'm going to rebase this branch on `main` to fix the conflict before reviewing.

chrislo:
I've checked out this branch locally and had a play - creating batches of users works as before and the two-stage process seems to work well in the happy path case. 

I noticed that if you start a batch upload, upload a CSV but stop before assigning permissions, the batch upload appears in the list of batch uploads in `/batch_invitations/new`. If you then click on the half-finished batch the error message is "Batch invitation doesn't have any permissions yet.". There's no way to then edit the batch permissions for that batch though, so the error message isn't particularly helpful. I wonder if in that case it should redirect to `/batch_invitations/<id>/permissions/new`.

chrislo:
I've addressed the issue I noticed with not being able to resume an incomplete batch in [c7ea253](https://github.com/alphagov/signon/pull/2350/commits/c7ea253f5297ac5ff3de394e9196e4d8b09bad6b).

chrislo (approved):

PR #2351, opened 2023-09-11
Fix and prevent unnecessary whitespace in organisation names

chrislo:
Trello: https://trello.com/c/rbUCwjHg

This commit removes and prevents unnecessary whitespace being added to `Organisation` `name` in a similar way to #2291 (for `User`). This only affects 1 record in integration and none in production, but it creates confusing ordering when we sort alphabetically by name.


chrisroos (approved):

PR #2354, opened 2023-09-11
Add token expiry timestamp Prometheus metrics

theseanything:
This allows signon to provide prometheus metrics about the expiry time for api user tokens. The `metrics` method is called every time the `/metrics` endpoint is request, hence we cache the expensive database call as tokens are unlikely to change frequently.

These metrics would allow us to monitor and alert on expiring tokens that need rotating.

### DIFF
--- a/lib/collectors/global_prometheus_collector.rb
+++ b/lib/collectors/global_prometheus_collector.rb
@@ -1,0 +1,40 @@
+require "prometheus_exporter"
+require "prometheus_exporter/server"
+
+# Load Rails models as not automatically included for Prometheus Exporter
+require File.expand_path("../../config/environment", __dir__) unless defined? Rails
+
+module Collectors
+  class GlobalPrometheusCollector < PrometheusExporter::Server::TypeCollector
+    def type
+      "signon_global"
+    end
+
+    def metrics
+      token_expiry_timestamp = PrometheusExporter::Metric::Gauge.new("signon_api_user_token_expiry_timestamp_seconds", "Timestamp when API User token expires")
+
+      token_expiry_info.each do |token|
+        token_expiry_timestamp.observe(token[:expires_at], api_user: token[:api_user], application: token[:application_name])
+      end
+
+      [token_expiry_timestamp]
+    end
+
+  private
+
+    def token_expiry_info
+      # Cache metric to prevent needless expensive calls to the database
+      Rails.cache.fetch("token_expiry_info", expires_in: 1.hour) do
+        ApiUser.all.flat_map do |user|
+          user.authorisations.where(revoked_at: nil).map do |token|
+            {
+              expires_at: token.expires_at.to_i,
+              api_user: user.email,
+              application_name: token.application.name.parameterize,
+            }
+          end
+        end
+      end
+    end
+  end
+end
--- a/config/initializers/prometheus.rb
+++ b/config/initializers/prometheus.rb
@@ -1,2 +1,4 @@
 require "govuk_app_config/govuk_prometheus_exporter"
-GovukPrometheusExporter.configure
+require "collectors/global_prometheus_collector"
+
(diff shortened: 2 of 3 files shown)


lib/collectors/global_prometheus_collector.rb
```diff
@@ -0,0 +1,40 @@
...
+  class GlobalPrometheusCollector < PrometheusExporter::Server::TypeCollector
+    def type
+      "app_global"
+    end
+
+    def metrics
+      token_expiry_timestamp = PrometheusExporter::Metric::Gauge.new("signon_api_user_token_expiry_timestamp_seconds", "Timestamp when API User token expires")
+
+      # Cache metric to prevent needless expensive calls to the database
+      token_expiry_info = Rails.cache.fetch("token_expiry_info", expires_in: 1.hour) do
```
chrislo, line 17: How often will this code be called? Do we know whether a 1 hour cache is sufficient to prevent those calls? Also - have you had a look to see how expensive these calls are in practice, I wonder if we could avoid the complication of caching if they're not expensive to run.

lib/collectors/global_prometheus_collector.rb
```diff
@@ -0,0 +1,40 @@
...
+        ApiUser.all.flat_map do |user|
```
chrislo, line 18: Would it be possible to add some tests for this class? In particular, I'm worried that if we make changes to ApiUser/User that this would break without us realising.

chrislo (changes requested):
Thanks for the PR! 

I'm not familiar with Promethus metrics - would you be able to add some more detail on the motivation for this change to the commit note. 

Also, I've had a look in `routes.rb` and we don't define a `/metrics` endpoint, so I can't see how this code is called - could you explain that too?

theseanything:
> I'm not familiar with Promethus metrics - would you be able to add some more detail on the motivation for this change to the commit note.

Updated the PR description to say "These metrics would allow us to monitor and alert on expiring tokens that need rotating."

> Also, I've had a look in `routes.rb` and we don't define a `/metrics` endpoint, so I can't see how this code is called - could you explain that too?

So most(/all) GOV.UK Rails applications are configured to run the Prometheus metrics exporter - this runs a separate web server process (not Rails) that serves the `/metrics` endpoint to be scraped by Prometheus. 

There configuration is in a shared gem: https://github.com/alphagov/govuk_app_config/blob/main/lib/govuk_app_config/govuk_prometheus_exporter.rb
which gets initialised here:
https://github.com/alphagov/signon/blob/main/config/initializers/prometheus.rb

More details about how the Prometheus Export works is here: 
https://github.com/discourse/prometheus_exporter

lib/collectors/global_prometheus_collector.rb
theseanything: So Prometheus will scrape this endpoint every minute. Individual token expiry timestamps never change. Changes in the data would be around tokens being created or deleted - which happen very seldom (a generous estimate would be maybe a couple times a month). Without any caching we're going to be doing a great deal of unnecessary expensive database calls.

1 hour was trade of minimising expensive calls, vs the freshness of the data we can see in our monitoring system - but probably could even be okay bumping caching up to a day.

chrislo: Is the once-per-minute frequency of calls configurable? Would scraping less often would be an alternative to caching (given that there isn't anything to expire that cache other than the expiry time).

theseanything: Not really - the /metrics endpoint has other metrics like http requests total, which we do want per minute resolution on

lib/collectors/global_prometheus_collector.rb
theseanything: Added a simple test case to check the metrics report the correct tokens.

theseanything:
@chrislo Thanks for the review - hopefully have address all the comments now :)

PR #2363, opened 2023-09-18
Small improvements to the Users page

mike29736:
https://trello.com/c/Lzcfwj5c/261-small-improvements-to-the-users-page

The first 2 small improvements to the Users page:

* remove the "Grant access to all users" feature
* move "Active" to the top of the Status filter options

Screenshot from 2023-09-18 16-30-20

(BTW, I'd be happy to squash those BulkGrantPermissionSet commits if that'll leave things neater)


floehopper (approved):
I know you've included it in the commit notes and there's probably info in the Trello card, but I think it would be nice to include at least a summary of the reasoning behind the 2 changes in the PR description.

Otherwise this looks good to me! 👍

PR #2367, opened 2023-09-20
Fix `<h1>` wrapping

mike29736:
Without `#present?`, `yield` seems to generate a false positive. I guess it could be returning an empty string by default.

That false positive was squashing all of our page's `<h1>`s into a one-third column, forcing even fairly short page headings to wrap

### Before
Screenshot from 2023-09-20 11-50-19
Screenshot from 2023-09-20 11-57-01

### After
Screenshot from 2023-09-20 12-09-09
Screenshot from 2023-09-20 12-08-19

### DIFF
--- a/app/views/layouts/admin_layout.html.erb
+++ b/app/views/layouts/admin_layout.html.erb
@@ -36,7 +36,7 @@
       <%= yield(:error_summary) %>
 
       <div class="govuk-grid-row">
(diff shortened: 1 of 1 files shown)


chrisroos (approved):
Looks good to me! (I've just fixed the same problem in my branch but can get rid of that now)

PR #2369, opened 2023-09-20
Prevent users changing their email addresses to non-governmental ones

chrislo:
Trello: https://trello.com/c/dp3QzDlq

We've now chased up all users who were using one of our identified "non-governmental" email domains, so we can enforce this validation rule for all User accounts, not just new ones. This strengthens the validation we introduced in 00e93e9570.


floehopper (approved):
LGTM 👍

PR #2371, opened 2023-09-21
Fixes for batch invitations

floehopper:
Trello: https://trello.com/c/aO4nVCZT

* Avoid displaying validation errors on successful processing of file. Previously if a validation error was displayed and then corrected, the error was incorrectly displayed again on the success page.
* Strip leading & trailing whitespace from `BatchInvitationUser#email` before validation to avoid unnecessarily confusing validation errors.
* Strip leading & trailing whitespace from `BatchInvitationUser#organisation_slug` before save to avoid the potential for failing organisation lookups.
* Improve test coverage of `BatchInvitationUser#email` validation.

### DIFF
--- a/app/controllers/batch_invitations_controller.rb
+++ b/app/controllers/batch_invitations_controller.rb
@@ -15,7 +15,7 @@ class BatchInvitationsController < ApplicationController
     authorize @batch_invitation
 
     unless file_uploaded?
-      flash[:alert] = "You must upload a file"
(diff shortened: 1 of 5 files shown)


mike29736 (approved):
:+1: 

I hadn't seen `ActiveRecord::Base::normalizes`, thanks for dropping that reference in!

I do wonder how much benefit we're getting from re-rendering that view instead of just redirecting back to it, but these `flash.now`s do fix the issue :shrug:

PR #2372, opened 2023-09-22
Fix copy-paste mishap in test setup

mike29736:
It was testing the wrong thing, now it's testing the right thing


floehopper (approved):
Good catch! This looks good to me, although there's a typo in the commit note which makes it a bit confusing!

> not it's testing the right thing

I assume "not" -> "now".

PR #2374, opened 2023-09-25
Fix for previous migrations that didn't strip whitespace for all the relevant records

floehopper:
This includes new migrations to strip leading/trailing whitespace from:
* `User#name`
* `BatchInvitationUser#name`
* `Organisation#name`

In #2373 I noticed that the query we'd used to do this in #2291 & #2351 didn't escape the regular expression properly. ActiveRecord seems to have generated the following SQL `WHERE` expression:

    WHERE (name REGEXP('^ +') OR name REGEXP(' +$')

I _think_ this will only have found names with leading/trailing space characters and not leading/trailing newlines/carriage-returns/tabs. Thus the previous migration didn't do anything bad, it just may have missed _some_ records.

In these new versions of the migration I've constructed the query slightly differently. Firstly I've double-escaped the whitespace character in the regular expression and secondly I've used [an Array Condition vs a Pure String Condition][1] to ensure argument safety. The new query generates the following SQL `WHERE` expression:

    WHERE (name REGEXP '^\\s+' OR name REGEXP '\\s+$')

This has the `REGEXP` string argument escaped correctly.

As per the migrations in #2373 I've use `ActiveRecord::Persistence#update_attribute` in order to skip model validation, in case some of the existing records are invalid for other reasons.

As before I did consider writing the migration as a SQL `UPDATE`, but generating a `TRIM` expression that is exactly equivalent to `String#strip` isn't as trivial as it sounds!

[1]: https://guides.rubyonrails.org/active_record_querying.html#conditions

### DIFF
--- a/db/migrate/20230925104847_strip_whitespace_from_user_name.rb
+++ b/db/migrate/20230925104847_strip_whitespace_from_user_name.rb
@@ -1,0 +1,7 @@
+class StripWhitespaceFromUserName < ActiveRecord::Migration[7.0]
+  def change
+    User.where("name REGEXP ? OR name REGEXP ?", "^\\s+", "\\s+$").each do |u|
+      u.update_attribute(:name, u.name&.strip) # rubocop:disable Rails/SkipsModelValidations
+    end
+  end
+end
--- a/db/migrate/20230925104848_strip_whitespace_from_batch_invitation_user_name.rb
+++ b/db/migrate/20230925104848_strip_whitespace_from_batch_invitation_user_name.rb
@@ -1,0 +1,7 @@
+class StripWhitespaceFromBatchInvitationUserName < ActiveRecord::Migration[7.0]
+  def change
+    BatchInvitationUser.where("name REGEXP ? OR name REGEXP ?", "^\\s+", "\\s+$").each do |u|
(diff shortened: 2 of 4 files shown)


mike29736 (approved):
:flushed: I'm impressed that you spotted this... even just looking into it to verify the changes, there don't seem to be many clues!

FWIW, it looks like we could also use character classes to achieve the same thing (e.g. `"^[[:space:]]+"`, found here: https://dev.mysql.com/doc/refman/8.0/en/regexp.html#regexp-syntax)

floehopper:
> 😳 I'm impressed that you spotted this... even just looking into it to verify the changes, there don't seem to be many clues!

Thanks - I spotted it because I was getting some confusing results when doing counts before & after the migration. It seems as if some of the `BatchUserInvitation` records had trailing newlines in their `email` field - perhaps we didn't have anything like that in the previous fields...?

> FWIW, it looks like we could also use character classes to achieve the same thing (e.g. `"^[[:space:]]+"`, found here: https://dev.mysql.com/doc/refman/8.0/en/regexp.html#regexp-syntax)

Yes, I did try that, but I was so confused by the results I was getting in the DB console at that point, I decided to avoid MySQL altogether! In hindsight, I think it probably would've worked!

PR #2375, opened 2023-09-25
Fix release version in footer

deborahchua:
This has been broken since we moved to EKS and the SHA is no longer provided.

We can use the [SENTRY_RELEASE](https://github.com/alphagov/govuk-helm-charts/blob/077117e7f40c096a958f9e0b853a2bd76cf43f26/charts/generic-govuk-app/templates/deployment.yaml#L92C1-L92C1) environment variable which is defined for every app and takes its value from the container image's tag.

Trello cards:
https://trello.com/c/uoKDUcbV/3182-fix-broken-version-footer-in-publishing-apps-2
https://trello.com/c/h19UQFNE/269-footerversion-not-set-in-govukadmintemplate

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -47,7 +47,7 @@
   <% if content_for?(:thin_form) %></div><% end %>
 <% end %>
 
-<% content_for :footer_version, CURRENT_RELEASE_SHA %>
+<% content_for :footer_version, ENV.fetch("SENTRY_RELEASE", "null") %>
 
 <%# use the govuk_admin_template layout %>
 <%= render :template => 'layouts/govuk_admin_template' %>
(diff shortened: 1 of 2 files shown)


floehopper (approved):
Thanks so much for making this change.

If I'm following the code correctly, I think this means that there will be a slight change in behaviour in development (i.e. when `SENTRY_RELEASE` is not defined) - the footer will now display "Version: " instead of "Version: development". Was that an intentional change? If so, it would be good to explain that in the commit message / PR description.

An alternative implementation preserving the original behaviour would be something like:

```ruby
<% content_for :footer_version, ENV.fetch("SENTRY_RELEASE", "development") %>
```

I'm happy to leave the above to your discretion, so I'm going to mark this as "Approved". Thanks, again!

floehopper:
For reference we are tracking this in https://trello.com/c/h19UQFNE

deborahchua:
Thanks @floehopper ! Your suggestion is much better and clearer to the user if there is an issue obtaining the image tag - I've updated the code and commit message.

PR #2376, opened 2023-09-25
Handle nil name in BatchInvitationUser#strip_whitespace_from_name

floehopper:
This was causing [a migration][1] to fail in integration with:

    NoMethodError: undefined method `strip!' for nil:NilClass

This makes sense from the point of view that there is no validation on BatchInvitationUser#name and so there's nothing forcing it to be present.

I've added a unit test to force me to fix the problem.

[1]: https://github.com/alphagov/signon/blob/afe562d1905c90def59e80a5e483375538fa0935/db/migrate/20230925093143_strip_whitespace_from_batch_invitation_user_organisation_slug.rb

### DIFF
--- a/app/models/batch_invitation_user.rb
+++ b/app/models/batch_invitation_user.rb
@@ -99,7 +99,7 @@ private
   end
 
   def strip_whitespace_from_name
-    name.strip!
+    name&.strip!
   end
 
   def strip_whitespace_from_email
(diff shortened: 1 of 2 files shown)


mike29736 (approved):
It didn't seem obvious that I should use safe navigation when I introduced that method :grimacing: Sorry

floehopper:
> It didn't seem obvious that I should use safe navigation when I introduced that method 😬 Sorry

No - it wasn't obvious to me either - no worries!

PR #2377, opened 2023-09-26
Account page using design system

mike29736:
https://trello.com/c/CTftWKa2/259-update-the-own-account-page-to-use-the-design-system

### What
This pull request introduces the new Account page at /account. It includes 4 or 5 cards (depending on user role) that link to various account-related "sub-pages".

### Summary (in commit-order)
* Move "Change your email or password" page from /users to /account
* Introduce the new Account page and make "Change your email or password" its first sub-page
* Build a temporary "Manage permissions" page and make it the second Account sub-page
* Move "2SV setup" page from /users to /account and make that an Account sub-page, too
* Build "Role and organisation" Account sub-page (the first one that wasn't copy-pasted or otherwise relocated)
* Build "Your account access log" Account sub-page
* Prevent users from editing their details through `/users/<id>/edit` and instead redirecting them `/account`

### Route changes
- Moves `/users/<id>/edit_email_or_password` to `/account/email_password`
- Moves `/users/<id>/update_email` to `/account/email_password/update_email`
- Moves `/users/<id>/update_password` to `/account/email_password/update_password`
- Moves `/users/two_step_verification` to `/account/two_step_verification`
- Copies `/users/<id>/resend_email_change` to `/account/email_password/resend_email_change`
- Copies `/users/<id>/cancel_email_change` to `/account/email_passwords/cancel_email_change`
- Copies `/users/<id>/event_logs` to `/account/activity`
- Adds `/account/manage_permissions`
- Adds `/account/role_organisation`, `/account/role_organisation/update_organisation` and `/account/role_organisation/update_role`

### Things I haven't done:
1. swapped the "2SV setup" link for a `Users#reset_two_step_verification`-like action (or probably more possible, for a link to a new page dedicated to that action), as Joseph suggested when we spoke earlier in the week, but I have let them know
2. improved existing content or probably even maintained the level it was at -- see: headings on "Change your email or password", breadcrumb names, the cards on the Account page that have no description

### Account page (view for a GOV.UK Admin who doesn't have 2SV setup yet)
Screenshot from 2023-09-29 13-40-31

### Account page (view for a normal user who already has 2SV setup)
Screenshot from 2023-09-29 14-31-48

### Breadcrumbs
Screenshot from 2023-09-29 14-26-57

### Manage permissions page (view for a GOV.UK Admin)
Screenshot from 2023-09-29 13-42-10

### Manage permissions page (view for a publishing manager)
Screenshot from 2023-09-29 13-43-12

### Role and organisation page (view for a GOV.UK Admin)
Screenshot from 2023-09-29 13-41-33

### Role and organisation page (view for a user who isn't a GOV.UK Admin)
Screenshot from 2023-09-29 14-27-50


app/views/accounts/show.html.erb
```diff
@@ -0,0 +1,53 @@
...
+      description: "Reset your password or change your email address."
+    },
+    (
+      {
+        link: {
+          text: "Manage permissions",
+          path: edit_user_path(current_user),
+        },
+        description: "Make changes to your permissions across different apps.",
+      } if policy(current_user).edit?
```
mike29736, line 34: I stole this `policy` from the AccountsController but I'm not sure it's what I want here

config/routes.rb
```diff
@@ -24,12 +24,14 @@
   devise_scope :user do
     post "/users/invitation/resend/:id" => "invitations#resend", :as => "resend_user_invitation"
     put "/users/confirmation" => "confirmations#update"
+    resource :two_step_verification_session,
```
chrisroos, line 27: I used `rails routes | grep two_step_verification` before and after this commit to confirm that there are no changes to the Rails-generated URL helper methods :+1:

config/routes.rb
```diff
@@ -30,7 +30,7 @@
              controller: "devise/two_step_verification_session"
     resource :two_step_verification,
              only: %i[show update],
-             path: "/users/two_step_verification",
+             path: "/account/two_step_verification",
```
chrisroos, line 33: It took me a while to realise that this is OK because you can only set-up 2SV for your own account (i.e. admins can't configure it for other users' accounts). It might be worth stating that in the commit note, but I'll leave it to you to decide whether it's useful or not.

app/controllers/account/email_passwords_controller.rb
```diff
@@ -15,7 +15,7 @@ def update_email
     elsif current_user.update(email: new_email)
       EventLog.record_email_change(current_user, current_email, new_email)
       UserMailer.email_changed_notification(current_user).deliver_later
-      redirect_to root_path, notice: "An email has been sent to #{new_email}. Follow the link in the email to update your address."
```
chrisroos: In `email_change_notice` you've changed `#{new_email}` to `#{current_user.unconfirmed_email}`. Presumable these are the same value?

app/controllers/account/email_passwords_controller.rb
```diff
@@ -45,7 +45,7 @@ def resend_email_change
   def cancel_email_change
     current_user.cancel_email_change
 
-    redirect_back_or_to root_path, notice: "You have cancelled your pending email address change."
+    redirect_to account_path, notice: "You have cancelled your pending email address change."
```
chrisroos, line 48: Is there a reason you've changed from `redirect_back_or_to` to `redirect_to`?

app/views/accounts/show.html.erb
```diff
@@ -40,5 +40,14 @@
...
       description: "Change your 2-step verification configuration."
     },
+    (
+      {
+        link: {
+          text: "Personal details",
+          path: account_personal_details_path,
+        },
+        description: "Make changes to your personal details such as your role.",
+      } if current_user.govuk_admin?
```
chrisroos, line 50: Wouldn't it be better to use `if policy([:account, :personal_details]).show?` here?

chrisroos (commented):
This all looks good to me, @mike29736!

I'm not approving it yet because I know you're making other changes.

app/controllers/account/email_passwords_controller.rb
mike29736: I might have understood `redirect_back_or_to` wrong. It only seemed to redirect back to the "Change your email or password" page, anyway. So this change was more accurately stopping `cancel_email_change` redirecting back to either the Dashboard or the Change your email or password page, and redirecting unconditionally to Account instead.

Not necessarily the right commit to make that change in. But if what I'm saying makes sense, at the very least I can explain it in the commit message.

app/controllers/users_controller.rb
```diff
@@ -69,12 +69,7 @@ def unlock
   def resend_email_change
     @user.resend_confirmation_instructions
     if @user.errors.empty?
-      notice = if @user.normal?
-                 "An email has been sent to #{@user.unconfirmed_email}. Follow the link in the email to update your address."
-               else
-                 "Successfully resent email change email to #{@user.unconfirmed_email}"
-               end
-      redirect_to root_path, notice:
+      redirect_to root_path, notice: "Successfully resent email change email to #{@user.unconfirmed_email}"
```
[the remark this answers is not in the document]
mike29736: Thanks, I can definitely mention this in the commit message.

Basically, as you're getting at with both of your points, `normal?` was being used as inaccurate shorthand for: Are they performing this action on their own account or on somebody else's?

There used to be a separate `UsersController` and `Admin::UsersController` and these two notice messages evolved independently, one in either controller. Other than that small difference, though, the controller actions were identical. When the controllers were merged (https://github.com/alphagov/signon/pull/298), the conditional was introduced to preserve the tiny difference in behaviour.

app/controllers/account/personal_details_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+      flash[:alert] = "There was a problem changing your organisation."
+      render :show
+    end
+  end
+
+  def update_role
+    previous_role = current_user.role
+    new_role = params[:user][:role]
+
+    if current_user.update(role: new_role)
```
chrisroos, line 25: I think we should probably use the `UserUpdate` class here, so that we also create the relevant `EventLog`s (see https://github.com/alphagov/signon/blob/main/app/services/user_update.rb#L78).

app/controllers/account/personal_details_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  before_action :authenticate_user!
+  before_action :authorise_user
+
+  def show; end
+
+  def update_organisation
+    new_organisation_id = params[:user][:organisation_id]
+    new_organisation = Organisation.find(new_organisation_id)
+
+    if current_user.update(organisation_id: new_organisation_id)
```
chrisroos, line 13: Similar to the comment about `#update_role` below, I think we should probably use `UserUpdate` here too. Surprisingly, we don't record a specific `EventLog` about changing organisation but I suspect that's an oversight.

I'm hopeful we can get rid of `UserUpdate` at some point but while it's still around I think we should probably consistently use it.

app/controllers/account/email_passwords_controller.rb
chrisroos: Ah, OK, that makes sense to me now. Thanks for explaining :+1:

app/controllers/account/personal_details_controller.rb
mike29736: I've switched to using `UserUpdate` and I don't think you were suggesting that I should, but FYI, I haven't added the missing event type

chrisroos (approved):
This all looks good to me, @mike29736!

I've added a couple of small commits to the branch:

- Ensure only superadmins can edit their role
- Redirect users to `/account` if they attempt to edit their own account using `/users/<id>/edit`

PR #2378, opened 2023-09-26
Remove code related to sending EventLog to Splunk

mike29736:
https://trello.com/c/Zp23Kd6c/274-remove-functionality-to-send-event-log-to-splunk

> [...] we’re now very confident this data has only ever been sent from
> integration and even that was some years ago.


floehopper (approved):
I think you might also be able to remove the `faraday` gem...?

Otherwise this looks good to me 👍

PR #2379, opened 2023-09-26
A heading for the right-hand section of dashboard

mike29736:
https://trello.com/c/1AarwGE2/266-add-a-actions-and-settings-title-on-the-right-side-of-the-dashboard

Based on feedback from user research, without a heading, this section doesn't really look like anything at all.

Our page layout code isn't compatible with the specific layout suggested in the mock up (in the Trello card), so I've taken some cues from [this example Housing and local services page](https://www.gov.uk/get-help-energy-bills), putting the heading below the line.

### How it looks
Screenshot from 2023-09-29 17-24-39

### DIFF
--- a/app/views/root/index.html.erb
+++ b/app/views/root/index.html.erb
@@ -55,16 +55,23 @@
   </div>
 
   <div class="govuk-grid-column-one-third">
-    <ul class="govuk-list app-dashboard-side">
-      <li>
-        <%= link_to "Change your email or password", edit_email_or_password_user_path(current_user), class: "govuk-link" %>
-      </li>
+    <div class="app-dashboard-side">
+      <%= render "govuk_publishing_components/components/heading", {
+        text: "Actions and settings",
+        padding: true,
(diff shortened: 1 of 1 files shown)


jlewr:
Is there a way to keep the top divider line on the apps list (above 'Content Data') and the top line on the right-hand section (blue line above 'Change your email...') in line with each other? I know what you mean about the difference

mike29736:
> Is there a way to keep the top divider line on the apps list (above 'Content Data') and the top line on the right-hand section (blue line above 'Change your email...') in line with each other? I know what you mean about the difference

Yeah, @jlewr, there might be a way to do that, though I'm not 100% sure. My first thought when I realised that it hadn't happened on its own was that it might've undermined the page's visual hierarchy, anyway. If we can fix it with the approach I'm imagining, there's a fair chance that that visual problem will correlate with an accessibility problem, too, so I'd better go get some help!

chrisroos (approved):
This looks good to me!

PR #2381, opened 2023-09-27
Update suspension and unlock email text

chrislo:
Trello: https://trello.com/c/0oTag1bi

This PR updates the content of the suspension reminder, suspension notification and unlock instruction emails following [some user research we did on their content](https://docs.google.com/document/d/1kIedI1QlYwf6RNjbVR3l4HApeRFmoHooShZNrDy7NNk/edit?pli=1).

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -119,11 +119,7 @@ private
   end
 
   def account_name
-    if GovukEnvironment.production?
-      "account"
-    else
-      "#{GovukEnvironment.name} account"
-    end
+    "#{GovukEnvironment.name} account"
   end
 
   def subject_for(key)
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -115,7 +115,11 @@ private
   end
 
   def unlock_time
-    (@user.locked_at + 1.hour).to_fs(:govuk_date)
+    t = (@user.locked_at + User.unlock_in)
+    time_part = t.to_fs(:govuk_time)
+    date_part = t.to_date.to_fs(:govuk_date)
+
+    "#{time_part} UK time on #{date_part}"
   end
 
   def account_name
--- a/test/models/user_mailer_test.rb
+++ b/test/models/user_mailer_test.rb
@@ -166,16 +166,8 @@ class UserMailerTest < ActionMailer::TestCase
       assert_body_includes "for user@example.com"
     end
 
-    should "state when the account was locked" do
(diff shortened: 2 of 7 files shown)


app/mailers/user_mailer.rb
```diff
@@ -115,7 +115,11 @@ def locked_time
   end
 
   def unlock_time
-    (@user.locked_at + 1.hour).to_fs(:govuk_date)
+    t = (@user.locked_at + 1.hour)
```
floehopper, line 118: I know it wasn't you who added it in the first place, but it would be better to use `User.unlock_in` here instead of `1.hour` - I believe the former is provided from the Devise config which I think what this `1.hour` refers to here.

test/models/user_mailer_test.rb
```diff
@@ -156,7 +156,7 @@ def assert_support_present_in_text(link_text, email = @email)
     setup do
       GovukEnvironment.stubs(:production?).returns(false)
       GovukEnvironment.stubs(:name).returns("test")
-      @the_time = Time.zone.now
+      @the_time = Time.zone.local(2023, 10, 31, 2, 0, 0)
```
floehopper, line 159: Subjective: I tend to use `Time.zone.parse("2023-10-31 02:00:00")` in tests, because I think it's more readable.

floehopper (approved):
I've added a couple of minor comments, but otherwise this looks good to me! 👍 

I think it would be good to make the `User.unlock_in` change before merging unless I've misunderstood that bit of code, but I'm happy to leave that to your discretion.

As an aside, I note that we have [Devise's `unlock_strategry` set to `:both`](https://github.com/alphagov/signon/blob/243f9eb898eb70be40cab6ffea55009b16f8e7cf/config/initializers/devise.rb#L184-L189) and my reading of the docs is that the idea of sending them the email is to give them the opportunity to unlock their own account (presumably legitimized by the fact they must have access to their email account to see the email). However, we don't seem to include a link to allow the user to unlock their own account. I think it would be interesting to understand whether not including the link was intentional and/or whether it's even sensible. Perhaps adding such a link in the email would be a simple way to reduce load on the Content 2nd Line Support team...?

app/mailers/user_mailer.rb
chrislo: That's great, thank you for spotting it.

test/models/user_mailer_test.rb
chrislo: Done, that is nicer, thanks!

chrislo:
I've captured the suggestion to allow users to unlock themselves in [this trello card](https://trello.com/c/HovPBBeK).

PR #2399, opened 2023-10-03
Fix "Upload a batch of users" button type

chrislo:
Trello: https://trello.com/c/zNFNt7Oj

This button matches the designs we wanted to use for this page, I picked the wrong one in 7c9864f49b.


floehopper (approved):
LGTM 👍

PR #2400, opened 2023-10-03
WIP: failing tests to reproduce bug

chrislo:
Trello: https://trello.com/c/i27MQaus


floehopper (approved):
Good work! LGTM 👍 

One minor thing - the test names suggest that the problem is specific to multiple permissions on the SAME app, but the commit note / PR description doesn't mention that. I'm guessing that's just an omission in the commit note / PR description, but either way it would be nice to bring the two things into line.

chrislo:
Thanks @floehopper! I'll clarify that in the commit note before merging.

PR #2412, opened 2023-10-06
Move invitations new to use GOV.UK Design System

floehopper:
Trello: https://trello.com/c/umBDQZUc

This moves the "Create new user" page to use the GOV.UK Design System.

For the moment, we've agreed to keep it as one page (rather than splitting the permission assignment into a 2nd page) and we haven't included the filtering of the applications in the permissions assignment section.

I've managed to use [the accordion component](https://components.publishing.service.gov.uk/component-guide/accordion) from the Design System gem for both this page and I've retrofitted it to the batch invitation permissions page to reduce duplication. However, it's worth noting that this component is marked as "experimental"!

I've also taken the opportunity to fix a bug where a number of the values in form fields (e.g. permissions checkboxes) were not being retained when a validation error occurred.

## Screenshots (truncated)

Screen Shot 2023-10-09 at 10 40 46

Screen Shot 2023-10-09 at 10 41 07


app/views/devise/invitations/new.html.erb
```diff
@@ -1,40 +1,120 @@
...
+
+        <%= render "govuk_publishing_components/components/select", {
+          id: "user_role",
+          name: "user[role]",
+          label: "Role",
+          hint: role_hint,
+          options: options_for_role_select(selected: f.object.role),
+        } %>
+      <% end %>
+
```
chrislo, line 83: In the [figma designs](https://www.figma.com/file/ArJZK2wjwRwIHJItpRTGjq/GOV.UK-Publishing-SignOn---June-2023?type=design&node-id=865-1658&mode=design) for this page, when we eventually split it into two, there's a "Manage Permissions" title for the second page with the permission select boxes. I wonder if it's worth adding a similar header here to break up the page and make it more scannable as an interim step.

app/views/devise/invitations/new.html.erb
```diff
@@ -81,32 +81,22 @@
...
-          <div class="govuk-accordion__section-content" aria-labelledby="accordion-heading-<%= idx %>">
-            <%= render "govuk_publishing_components/components/checkboxes", {
+      <%= render "govuk_publishing_components/components/accordion", {
+        items: policy_scope(:user_permission_manageable_application).reject(&:retired?).map do |application|
+          {
+            heading: {
+              text: application.name
+            },
+            content: {
+              html: render("govuk_publishing_components/components/checkboxes", {
```
chrislo, line 91: I didn't know you could do this, thanks for picking it up!

chrislo (approved):
This all looks good to me. I made a suggestion about adding a section header, but we can always come back to that.

app/views/devise/invitations/new.html.erb
floehopper: > In the [figma designs](https://www.figma.com/file/ArJZK2wjwRwIHJItpRTGjq/GOV.UK-Publishing-SignOn---June-2023?type=design&node-id=865-1658&mode=design) for this page, when we eventually split it into two

My current understanding is that we're not _definitely_ going to split this page, but I may have got the wrong end of the stick. Joseph was less concerned about splitting *this* page versus the batch invitation page, because the latter has even more going on with the CSV, etc.

> there's a "Manage Permissions" title for the second page with the permission select boxes. I wonder if it's worth adding a similar header here to break up the page and make it more scannable as an interim step.

Good point. Looking at the diff, I've lost an "Permissions" `h2` heading, so I'll add something like that back before merging.

floehopper: Addressed in https://github.com/alphagov/signon/pull/2412/commits/45c8da72c26ae363ae73f8495455d61b637eff57

PR #2417, opened 2023-10-10
Improve organisation select on role organisations page

chrislo:
Trello: https://trello.com/c/R9bKDyi5

The current "Your Organisation" dropdown on `/account/role_organisation` is hard to navigate. This PR improves things a bit by excluding closed organisations from the list and sorting the organisations alphabetically by name.


app/helpers/role_organisations_helper.rb
```diff
@@ -1,6 +1,6 @@
 module RoleOrganisationsHelper
   def options_for_your_organisation_select(current_user)
-    organisations = Pundit.policy_scope(current_user, Organisation).order(:name)
+    organisations = Pundit.policy_scope(current_user, Organisation).order(:name).where(closed: false)
```
floehopper, line 3: Subjective: I'd expect calls to `#order` to go after calls to `#where`.

floehopper: Subjective: It might be nice to extract an `Organisation.not_closed` scope to make this slightly more readable.

test/helpers/role_organisations_helper_test.rb
```diff
@@ -0,0 +1,29 @@
...
+      assert(options_for_your_organisation_select(@user).detect { |o| o[:value] == @other_organisation.id })
+    end
+
+    should "select the user's organisation" do
+      user_organisation_option = options_for_your_organisation_select(@user).detect { |o| o[:value] == @user_organisation.id }
+      other_organisation_option = options_for_your_organisation_select(@user).detect { |o| o[:value] == @other_organisation.id }
+
+      assert user_organisation_option[:selected]
+      assert_not other_organisation_option[:selected]
+    end
```
floehopper, line 28: Subjective: I find the assertions in these tests hard to read. I think it's partly because they are quite long lines but also if feels as if there's two things going on - (1) invoking the method under test; and (2) digging into the return value to make the assertion. I think I'd find them more readable if they were split up along those lines, e.g.

```
    should "use the name with abbreviation for the organisation as the option text" do
      options = options_for_your_organisation_select(@user)

      assert options.detect { |o| o[:text] == @user_organisation.name_with_abbreviation }
      assert options.detect { |o| o[:text] == @other_organisation.name_with_abbreviation }
    end
```

Note that I've also removed some parens from the assertion which I think are redundant.

Thinking about it a bit more, I think I might be inclined to just assert the whole structure in a single test like I did here:

https://github.com/alphagov/signon/blob/3f25de1c9796a1c5584dc4267c9649965c9cf5c2/test/helpers/users_helper_test.rb#L77-L113

test/helpers/role_organisations_helper_test.rb
```diff
@@ -25,5 +25,14 @@ class RoleOrganisationsHelperTest < ActionView::TestCase
       assert user_organisation_option[:selected]
       assert_not other_organisation_option[:selected]
     end
+
+    should "sort by organisation name alphabetically" do
+      options = options_for_your_organisation_select(@user)
+
+      assert_equal(
+        [@other_organisation.name_with_abbreviation, @user_organisation.name_with_abbreviation],
```
floehopper, line 34: Subjective: It's not immediately obvious to me how this test is testing the ordering. This is because: (a) the use of the instance variables means I have to look up at the setup to see what they are; and (b) because even then neither organisation has its `name` set explicitly - I assume it's relying on the Organisation factory naming sequence. Of the two points, I think (b) is the one I care about more.

I think this (and other tests) in this test case would be easier to follow if the organisation names were set explicitly and those string literals were repeated in the assertions even at the cost of a tiny bit of duplication.

floehopper (approved):
This all looks good to me, but I've added a bunch of subjective comments. I'm happy to leave you to decide whether/how to address them, so I'll mark this as approved.

One thing it might be worth changing is the "Sort organisation options alphabetically by name (with abbreviation)" commit subject line. From what I can tell the sorting is by name and not abbreviation. If that's correct, I found "(with abbreviation)" bit of the commit title confusing!

app/helpers/role_organisations_helper.rb
chrislo: Fixed the method chain order in [09e0e04](https://github.com/alphagov/signon/pull/2417/commits/09e0e04297394e8d1d027c7750eff9a6c17f8cc5)

chrislo: Introduced the new scope in 35b9d0be

test/helpers/role_organisations_helper_test.rb
chrislo: Addressed in [edac91f](https://github.com/alphagov/signon/pull/2417/commits/edac91fefe6c6d3baab9c1e8115ed062f74fbe29)

test/helpers/role_organisations_helper_test.rb
chrislo: Addressed in 44bba6a0

PR #2418, opened 2023-10-10
Create a new access token for the SSO Push API user *before* expiry not *at* expiry

floehopper:
Trello: https://trello.com/c/XfBBRL5M

The main change in this PR is that we will now create a new access token for the SSO Push API user on-demand if the current token is due to expire in the next 4 weeks rather than if it has already expired.

Some Prometheus alerting was added in #2354 but these tokens for the SSO Push API user were excluded in #2413, because, unlike tokens for other API users, they shouldn't need to be manually rotated by 2nd Line Tech. However, @theseanything suggested that if we create the tokens before they expire and before the alert fires, then we can include these tokens in the alerting so that if anything goes wrong with the automatic token creation 2nd Line Tech will receive an alert.

I've also made some minor improvements to the API user edit page which should make it a bit easier to work out what's going on with a user's tokens:
- the tokens are now listed in order of application name and expiry time
- there is a new state column which highlights tokens that have expired

I've also extracted a bunch of scopes to try to reduce duplication and make the code a bit easier to read.


floehopper:
Ugh - failure is due to Brakeman SQL Injection check - I'll sort it out after lunch!

chrislo:
This all looks good to me. Thanks for taking the time to extract some scopes and remove that unnecessary `::`! 

We spoke on slack about the use of `sanitize_sql` in a couple of places to keep Brakeman happy - I don't know if it's worth including a bit of that context in the relevant commit notes (https://github.com/alphagov/signon/pull/2418/commits/05dc3e598ff9902de7911c8af34d94556daa8afc and https://github.com/alphagov/signon/pull/2418/commits/72d31a8f54d3719567edf5ec34e26db316d8122c) but I'll leave that up to you.

chrislo (approved):

floehopper:
> We spoke on slack about the use of `sanitize_sql` in a couple of places to keep Brakeman happy - I don't know if it's worth including a bit of that context in the relevant commit notes ([05dc3e5](https://github.com/alphagov/signon/commit/05dc3e598ff9902de7911c8af34d94556daa8afc) and [72d31a8](https://github.com/alphagov/signon/commit/72d31a8f54d3719567edf5ec34e26db316d8122c)) but I'll leave that up to you.

Good idea - thanks!

PR #2424, opened 2023-10-10
Use option-select vs checkboxes for granting permissions

floehopper:
Trello: https://trello.com/c/SfIc6Q2q

The main motivation for these changes is the Licensing app which has a large number of permissions. The [recent][1] [changes][2] which moved the UI for granting permissions to the GOV.UK Design System have made managing the permissions for the Licensing app unmanageable.

This PR replaces [the accordion component][3] where each item of the accordion was [a checkboxes component][4] (i.e. a set of checkboxes) with a list of [option-select components][5] (each of which has a set of checkboxes within it). We've made this change to *both* the new invitation ("Create new user") page and the new batch invitation permissions page.

The option-select component gives us something similar to the show/hide toggle which was previously provided by the accordion which is why it no longer makes sense to use the accordion. Also it makes the permissions section more compact which is no bad thing given how long it is! Most importantly the option-select component provides a search filter for the permissions for a given app - this is enabled for apps which have more than 4 permissions which is roughly the point that a scrollbar appears. This addresses the primary motivation which is improving the managability of the Licensing app permissions.

We've also set each option-select component to open on load if any of its permissions are selected and moved any checked permission checkboxes to the top of the list. At the moment this is only relevant to the new invitation page when validation errors have occurred and the form has been re-rendered. This behaviour is very similar to what we did on the users index page.

The use of the option-select component isn't really what it was intended for in that it doesn't *control* a list of things. However, I don't think it's too far away and it gives us a much better UI.

In making this change I've worked out we don't really need to set the `id` attribute for each checkbox in `UsersHelper#options_for_permission_option_select` - the checkboxes component within the option-select component will generate unique `id` values automatically and we don't need to reference them anywhere. Also we can set the checkbox name at the option-select component level (via its `key` attribute) rather than having to set it for each of its options in `UsersHelper#options_for_permission_option_select`. This simplifies the code somewhat.

### New invitation page
Screenshot 2023-10-11 at 15 36 48

### New batch invitation permissions page
Screenshot 2023-10-11 at 15 36 48

[1]: https://github.com/alphagov/signon/pull/2361
[2]: https://github.com/alphagov/signon/pull/2412
[3]: https://components.publishing.service.gov.uk/component-guide/accordion
[4]: https://components.publishing.service.gov.uk/component-guide/checkboxes
[5]: https://components.publishing.service.gov.uk/component-guide/option_select

### DIFF
--- a/app/views/devise/invitations/new.html.erb
+++ b/app/views/devise/invitations/new.html.erb
@@ -94,7 +94,7 @@
             options_container_id: "user_application_#{application.id}_supported_permissions",
             show_filter: options.length > 4,
             closed_on_load: options.none? { |o| o[:checked] },
-            options:,
+            options: with_checked_options_at_top(options),
           }) %>
         <% end %>
 
--- a/test/controllers/invitations_controller_test.rb
+++ b/test/controllers/invitations_controller_test.rb
@@ -348,6 +348,22 @@ class InvitationsControllerTest < ActionController::TestCase
         end
       end
 
+      should "put selected permissions at the top if there are validation errors" do
+        application = create(:application)
+        signin_permission = application.signin_permission
+        other_permission = create(:supported_permission, application:)
+        selected_permissions = [other_permission]
+
+        post :create, params: { user: { supported_permission_ids: selected_permissions.map(&:to_param) } }
+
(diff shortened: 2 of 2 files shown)


chrislo (approved):
This all looks good to me 👍

PR #2426, opened 2023-10-11
Improve Organisation select on /users/invitation/new

chrislo:
Trello: https://trello.com/c/R9bKDyi5

Rosa got in touch to say that this Organisation select box is not very usable, particularly in integration. The organisations are sorted in a seemingly random order and they include closed organisations that we never need to assign to new users.

This commit fixes both of those issues.

The previous tests stubbed `policy_scope` which is a Pundit-provided convenience for `Pundit.policy_scope` automatically passing in `current_user` as the first argument. I wanted to actually call the scope, so I could test that the `not_closed` and `order` scopes I've added were working correctly. To get `policy_scope` to work in the isolated tests I've explicitly included `Pundit::Authorization` in `UsersHelper` and stubbed `current_user` instead.

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -1,4 +1,6 @@
 module UsersHelper
+  include Pundit::Authorization
+
   def two_step_status(user)
     user.two_step_status.humanize.capitalize
   end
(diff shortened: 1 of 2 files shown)


floehopper (approved):
LGTM 👍

PR #2430, opened 2023-10-11
Improve usability of organisation select on batch_invitations/new

chrislo:
Trello: https://trello.com/c/gJkINQst

This makes it easier to select the organisation for a batch of users by excluding closed organisations and sorting the remaining ones alphabetically. We did something similar in #2426 and #2417 - there's enough subtle differences in those two select boxes (indicating a current organisation and adding a "None" option respectively) that I haven't attempted to extract the duplication. I think the new scopes are self-explanatory enough that I'm happy to keep them inline in this view.


floehopper (approved):
This looks good to me 👍

It would be nice to add a test (e.g. in `BatchInvitationsControllerTest` to check we're not rendering options for closed organisations. However, I'm going to mark this as approved and you can decide whether it's worthwhile.

chrislo:
Thanks @floehopper - I added a controller test as you suggested, good idea!

PR #2435, opened 2023-10-16
Fix permissions bug in UserUpdate

chrisroos:
We use the `UserUpdate` class in `Account::RoleOrganisationsController` to update the user's role and organisation. We're not passing `supported_permission_ids` in the `params` parameter to `UserUpdate#initialize` because we're not interested in changing a user's permissions in these operations. Unfortunately it turns out that `UserUpdate` expects to have the user's existing permissions passed in if you want them to remain unchanged. Otherwise `UserUpdate#filtered_user_params` will reset/clear all permissions for that user.

I've chosen to fix this in `UserUpdate#filtered_user_params` because it seems reasonable to me that leaving `supported_permission_ids` out of `params` should be interpreted as meaning that supported permissions shouldn't be changed.

Note that this bug would've caused superadmins to lose their permissions if they changed their role or organisation, and admins to lose their permissions if they changed their organisation.

I've also updated `UserUpdate` to record an EventLog when a user's organisation is changed. We appear to record an event when anything else is changed by `UserUpdate` so it seems reasonable that we should also record an event when the organisation is changed.

### DIFF
--- a/app/controllers/account/manage_permissions_controller.rb
+++ b/app/controllers/account/manage_permissions_controller.rb
@@ -4,6 +4,7 @@ class Account::ManagePermissionsController < ApplicationController
 
   before_action :authenticate_user!
   before_action :authorise_user
+  before_action :allow_no_application_access, only: [:update]
 
   def show
     @application_permissions = all_applications_and_permissions_for(current_user)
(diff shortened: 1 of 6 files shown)


floehopper (approved):
This looks good to me! 👍 

I do have a wider concern about all the events that are recorded in this "service" class rather than at what I would regard as a safer lower level on the `User` model itself. However, that's definitely out-of-scope for this PR!

chrisroos:
Thanks, @floehopper :+1:

PR #2436, opened 2023-10-16
Preselect some permissions on users invitation new

chrislo:
Trello: https://trello.com/c/5tHKKbWq

This PR preselects some permissions when inviting a new user. These permissions are commonly added to new users but we want the inviting user to be able to decide whether they are required - so it didn't make sense to just grant them on user creation. By having them preselected we save the inviting user some time as they don't have to find the individual permissions themselves in the list of dropdown select boxes. 

## Before

before

## After

after

### DIFF
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -13,7 +13,9 @@ class InvitationsController < Devise::InvitationsController
 
   def new
     authorize User
-    super
+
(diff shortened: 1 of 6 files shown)


chrislo:
@chrisroos pointed out that `SupportedPermission#default` has some overlap with the functionality added in this PR and it might be better to rework it to use that instead. I'm going to close this PR while I investigate.

PR #2442, opened 2023-10-17
Preselect default supported permissions when creating new users

chrislo:
Trello: https://trello.com/c/5tHKKbWq

This PR ensures that any SupportedPermission with `default` set to `true` will be preselected in the user invitation and batch user upload permissions pages. With this change we no longer grant these permissions by default automatically, allowing the inviting user to deselect some of them if they are not appropriate for the specific user. 

I haven't included a migration to make the permissions listed in the trello card default. I figured we can easily do this through the UI once this change has landed (and double-check that we want the same permissions by default on both integration and production). 

# User Invitation
user_invitation_before

## After
users_invitation_after

# Batch Invitation Permissions

## Before
batch_permissions_before

## After
batch_permissions_after

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -211,7 +211,11 @@ class User < ApplicationRecord
   end
 
   def grant_permission(supported_permission)
-    application_permissions.where(supported_permission_id: supported_permission.id).first_or_create!
+    if persisted?
+      application_permissions.where(supported_permission_id: supported_permission.id).first_or_create!
+    else
+      supported_permissions << supported_permission unless supported_permissions.include?(supported_permission)
+    end
   end
 
   # This overrides `Devise::Recoverable` behavior.
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -122,6 +122,10 @@ class User < ApplicationRecord
     relation
   end
 
+  def self.with_default_permissions
+    new(supported_permissions: SupportedPermission.default)
+  end
+
   def require_2sv?
     return require_2sv unless organisation
 
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -538,6 +538,30 @@ class UserTest < ActiveSupport::TestCase
         assert user.supported_permissions.include?(supported_permission)
       end
     end
+
+    context "where the user is a new user" do
+      should "grant the permission" do
+        user = build(:user)
+        application = create(:application)
+        supported_permission = create(:supported_permission, application:)
+
+        user.grant_permission(supported_permission)
+
+        assert user.has_permission?(supported_permission)
+      end
+
+      should "prevent the permission being granted twice" do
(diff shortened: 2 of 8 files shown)


app/models/user.rb
```diff
@@ -211,7 +211,11 @@ def grant_application_permissions(application, supported_permission_names)
   end
 
   def grant_permission(supported_permission)
-    application_permissions.where(supported_permission_id: supported_permission.id).first_or_create!
+    if persisted?
+      application_permissions.where(supported_permission_id: supported_permission.id).first_or_create!
+    else
+      supported_permissions << supported_permission
+    end
```
chrisroos, line 228: I think there's a potential problem here because it's possible to add the same supported permission multiple times to an unsaved user. Saving that user will then fail with `ActiveRecord::RecordInvalid: Validation failed: Supported permission has already been taken`. I wonder if it's better to use something like:

`supported_permissions << supported_permission unless supported_permissions.include?(supported_permission)`

Which works in both the persisted and unpersisted case.

app/models/user.rb
```diff
@@ -122,6 +122,16 @@ def self.with_2sv_statuses(scope_names)
     relation
   end
 
+  def self.with_default_permissions
```
chrisroos, line 125: I think this method can be simplified as `new(supported_permissions: SupportedPermission.default)`.

test/models/user_test.rb
```diff
@@ -512,6 +512,16 @@ def setup
...
 
+  test "#grant_permission" do
+    user = create(:user)
+    application = create(:application)
+    supported_permission = create(:supported_permission, application:)
+
+    user.grant_permission(supported_permission)
+
+    assert user.has_permission?(supported_permission)
+  end
```
chrisroos, line 544: I think it might be worth adding another test here, to exercise the branch of code where the user already has the supported permission (i.e. the `first` bit of `first_or_create`).

chrisroos (changes requested):
This all looks good apart from the implementation of `User#grant_permission` which I think should work consistently irrespective of whether the user is persisted.

I've suggested another couple of changes but I'll leave it to you to decide whether to make those.

app/models/user.rb
chrislo: Fixed in [cbf2077](https://github.com/alphagov/signon/pull/2442/commits/cbf207774211c8897f2bbef972ffceb5619d2fb1)

test/models/user_test.rb
chrislo: I expanded the test in [62b4a86](https://github.com/alphagov/signon/pull/2442/commits/62b4a86517943f4e0e41aaf583f423ac4934e773) to test this branch.

chrislo:
Thanks @chrisroos - I'll rebase this and get it merged.

PR #2445, opened 2023-10-18
Remove users:create rake task

chrislo:
I noticed the existence of this rake task while working on #2442. As far as I can tell from the history (6333730c, 3432e0c9) this task was added when there was no other mechanism for adding users. We now have the /users/invitation/new and /batch_invitations/new pages to create individual and batches of users respectively. I've asked in the `#govuk-2ndline-tech` slack channel and no-one can remember using this task recently to add users.

I think it's safe to remove it, the UserCreator class that contains its logic, the related tests and documentation.


floehopper (approved):
This looks good to me 👍 

I did have a quick thought that some rake tasks are run by code in https://github.com/alphagov/govuk-helm-charts (mainly as scheduled tasks). However, unsuprisingly there's no sign of this one which would've been weird anyway!

PR #2446, opened 2023-10-18
Hide retired apps on most pages

floehopper:

While the Trello card is about hiding API-only apps, the overall motivation is to reduce the number of apps on various permissions pages and so I thought a good first step would be to exclude permissions relating to retired apps.

This adds a condition to the default scope on the `Doorkeeper::Application` model to only include `not_retired` applications in any query by default. It also incorporates this default scope into the `has_many :application_permissions` & `has_many :supported_permissions` associations on the `User` model. While using a default scope is perhaps a bit controversial, I think the notion of retiring an application is effectively a "soft delete" and this seems like a good use case for a default scope, not least because it should make it less likely a retired app or its permissions show up on any pages added in the future.

All this should mean that retired apps will only be listed in the following places:

1. On the "Retired" tab on the apps index page. It should still be possible to view all the permissions supported by a retired application via this page.
2. On the API user edit page in the "Tokens" section. This is because I haven't (yet) incorporated the default scope on the `Doorkeeper::Application` model to the `has_many :authorisations` association on the `User` model. I've decided to hold off on that for now (a) to reduce the scope of this PR; and (b) because 2nd line might still get alerts for expired tokens associated with retired apps and I want them to be able to still be able to access the UI to revoke/regenerate the token.

Retired apps should no longer appear on any pages where a user's permissions are being added/removed/changed.

The most significant commits are:
* [Add `not_retired` to `Doorkeeper::Application` default scope ](https://github.com/alphagov/signon/pull/2446/commits/7367d2e95bcfd84212395265276cdbcdfae3c102)
* [Exclude retired apps from `User#application_permissions`](https://github.com/alphagov/signon/pull/2446/commits/044a4aabaed3e993a4078a57d70205ffac9405a8)
* [Exclude retired apps from `User#supported_permissions`](https://github.com/alphagov/signon/pull/2446/commits/5bae5c412b66353b3c5437beeb5cb7c4ecc70502)

The other commits are mostly adding test coverage, removing redundant code, or just generally tidying up.

I haven't yet removed the logic to wrap application names in `<del>` tags to strike-through the name, because I want to double-check no retired apps end up appearing anywhere in the integration or production environments before I make that change.

Joseph has recommended that we publicise this change in the `#govuk-publishing` Slack channel before we deploy it.

### DIFF
--- a/test/factories/oauth_applications.rb
+++ b/test/factories/oauth_applications.rb
@@ -18,19 +18,19 @@ FactoryBot.define do
         # this line takes care of tests creating signin in order to look complete or modify delegatable on it.
         app.signin_permission.update(delegatable: false) && next if permission_name == SupportedPermission::SIGNIN_NAME
 
-        create(:supported_permission, application_id: app.id, name: permission_name)
+        create(:supported_permission, application: app, name: permission_name)
       end
 
       evaluator.with_supported_permissions_not_grantable_from_ui.each do |permission_name|
         next if permission_name == SupportedPermission::SIGNIN_NAME
 
-        create(:supported_permission, application_id: app.id, name: permission_name, grantable_from_ui: false)
+        create(:supported_permission, application: app, name: permission_name, grantable_from_ui: false)
       end
 
       evaluator.with_delegatable_supported_permissions.each do |permission_name|
         next if permission_name == SupportedPermission::SIGNIN_NAME
 
-        create(:delegatable_supported_permission, application_id: app.id, name: permission_name)
+        create(:delegatable_supported_permission, application: app, name: permission_name)
       end
     end
   end
--- a/app/controllers/account/applications_controller.rb
+++ b/app/controllers/account/applications_controller.rb
@@ -13,6 +13,6 @@ class Account::ApplicationsController < ApplicationController
     authorize [:account, Doorkeeper::Application]
 
     @applications_with_signin = Doorkeeper::Application.can_signin(current_user)
-    @applications_without_signin = Doorkeeper::Application.not_retired.without_signin_permission_for(current_user)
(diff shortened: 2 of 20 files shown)


test/factories/oauth_applications.rb
```diff
@@ -18,19 +18,19 @@
...
 
-        create(:supported_permission, application_id: app.id, name: permission_name)
+        create(:supported_permission, application: app, name: permission_name)
       end
 
       evaluator.with_supported_permissions_not_grantable_from_ui.each do |permission_name|
         next if permission_name == SupportedPermission::SIGNIN_NAME
 
-        create(:supported_permission, application_id: app.id, name: permission_name, grantable_from_ui: false)
+        create(:supported_permission, application: app, name: permission_name, grantable_from_ui: false)
```
chrislo, line 27: I appreciate you didn't introduce this change, but if we were to name the block parameter`application` we could pass `application:` instead of `application: app` here (and elsewhere in this commit) to further simplify things.

chrislo (approved):
This all looks good to me @floehopper - thanks for the detailed commit messages and easy-to-follow history. I've made one small suggestion that shouldn't block merging this.

test/factories/oauth_applications.rb
floehopper: It's a good thought, but I'm a bit nervous that it might be confusing given the factory is called `application` too. So I think I'll leave it for now. Also I'd prefer to get rid of the `after(:create)` hook altogether and I've got a personal to-do note about that!

floehopper:
> This all looks good to me @floehopper - thanks for the detailed commit messages and easy-to-follow history. I've made one small suggestion that shouldn't block merging this.

Thanks, @chrislo!

PR #2451, opened 2023-10-19
Button to clear permissions on new user pages

chrislo:
Trello: https://trello.com/c/c6TxIASX

In #2436 we ensured that the default permissions for new users were checked in the permissions checkboxes on when inviting new users (singly or in a batch). This is a big time-saver for some inviting users who frequently add these
permissions. However there's a subset of inviting users who set up new user accounts who don't need these default permissions (e.g. when setting up users who use the Licensing app).

This PR adds a "Clear selected permissions" button to the `/users/invitations/new` and `/batch_invitations/:id/permissions/new` pages to make it easier for those inviting users to complete their task. 

Clicking the button clears the checked boxes and resets the "selected" counter for each app

https://github.com/alphagov/signon/assets/16707/570af1c2-a658-4003-8735-511ca5ef5bde

### DIFF
--- a/app/assets/javascripts/modules/clear-selected-permissions.js
+++ b/app/assets/javascripts/modules/clear-selected-permissions.js
@@ -1,0 +1,34 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  'use strict'
+
+  function ClearSelectedPermissions (module) {
+    this.module = module
+    this.clearButton = this.module.querySelector('button[data-action="clear"]')
+    this.checkboxes = this.module.querySelectorAll('.govuk-checkboxes__input')
+    this.checkboxLists = this.module.querySelectorAll('.gem-c-checkboxes__list')
+  }
+
+  ClearSelectedPermissions.prototype.init = function () {
+    this.clearButton.addEventListener('click', this.clear.bind(this))
+  }
+
+  ClearSelectedPermissions.prototype.clear = function (event) {
+    event.preventDefault()
+    this.uncheckBoxes()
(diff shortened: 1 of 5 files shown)


floehopper (approved):
This all looks great to me - good work! Updating the number of selected items is a nice catch and your solution is very neat! 🚀 

One question: Does it definitely make sense to have the button *outside* the form? And if so, would it make sense to set the `aria_controls` option (see [these docs](https://components.publishing.service.gov.uk/component-guide/button/with_aria_controls)) for the component to point at the form?

floehopper:
> One question: Does it definitely make sense to have the button *outside* the form? And if so, would it make sense to set the `aria_controls` option (see [these docs](https://components.publishing.service.gov.uk/component-guide/button/with_aria_controls)) for the component to point at the form?

Looking at this again I can see the button is *inside* the `form` (and even the relevant `fieldset`) - I must've got confused when I looked at the diff last week. I did still contemplate adding an `id` to the `fieldset` and setting the button's `aria-controls` attribute to point to it, but I decided it probably wasn't worth it since the button is so "near" the checkboxes that it controls.

floehopper:
I've just noticed some single quotes which have been accidentally added in the HTML, so I'm going to fix up the relevant commits.

floehopper:
> I've just noticed some single quotes which have been accidentally added in the HTML, so I'm going to fix up the relevant commits.

Addressed in:
* https://github.com/alphagov/signon/pull/2451/commits/196e09a3ff5288440ca552842c731ddeb64926f1
* https://github.com/alphagov/signon/pull/2451/commits/00c42dab9915089b44024a3bf3d7eff080a2ee74

I'm happy to get this merged now!

PR #2452, opened 2023-10-19
Hide API-only apps on most pages

floehopper:

This follows on from #2446 in which we hid retired apps from most pages.

The idea in this PR is to hide API-only apps from most (but not all) pages in order to reduce the clutter on a bunch of pages - mainly (only?) pages which involve viewing/managing permissions for a user. It should no longer be possible to view or edit permissions for API-only apps on any pages except for the API user edit page and the "Edit supported permissions" page accessible via the application link in the "API-only" tab on the applications index page. We think this makes sense, because a non-API user should never need permissions for an API-only application.

Screenshot 2023-10-19 at 17 49 02

Note that I've replaced `Doorkeeper::Application#show_on_dashboard` with `Doorkeeper::Application#api_only` and inverted the values. I think this better reflects what the boolean flag means.

The commits which make significant changes are:
* [Replace `Application#show_on_dashboard` -> `#api_only`](https://github.com/alphagov/signon/pull/2452/commits/b2bff6a9172378f2cf23070804a7808fb942ed0c)
* [Add new tab for API-only applications on applications index page](https://github.com/alphagov/signon/pull/2452/commits/680bce76ec3f49234781dfb83edad97aec25772d)
* [Do not display API-only apps on account applications page](https://github.com/alphagov/signon/pull/2452/commits/d36723a7e14a7ca95b99b3c5e15161df3b49fb3b)
* [Do not display API-only apps on account permissions page](https://github.com/alphagov/signon/pull/2452/commits/8c6a56bf0a4513085b4ad32d2ff4c2bf0f345ef3)
* [Do not display API-only apps on a bunch of pages](https://github.com/alphagov/signon/pull/2452/commits/858309d4f8a09bd3bf877b271d042814dfc92475)
    * `/users/invitation/new`
    * `/batch_invitations/:batch_invitation_id/permissions/new`
    * `/account/manage_permissions`
* [Do not display API-only apps on two more pages](https://github.com/alphagov/signon/pull/2452/commits/92cf7eb232f4d53239b9422121993010cb8bdc15)

Most of the rest of the commits are adding test coverage to (hopefully) catch any regressions.

I would dearly like to have done some more refactoring of the code, particularly that involving `UserPermissionsControllerMethods` and `UserPermissionManageableApplicationPolicy`, but the scope of these changes was already large. Hopefully the test coverage I've put in place would make any such refactoring a bit easier the future.

### DIFF
--- a/test/controllers/account/manage_permissions_controller_test.rb
+++ b/test/controllers/account/manage_permissions_controller_test.rb
@@ -44,6 +44,20 @@ class Account::ManagePermissionsControllerTest < ActionController::TestCase
           assert_select "td", count: 0, text: retired_app.name
         end
       end
+
+      should "not list API-only applications" do
+        user = create(:admin_user, email: "admin@gov.uk")
+        sign_in user
+
+        api_only_app = create(:application, api_only: true)
+        user.grant_application_signin_permission(api_only_app)
+
+        get :show
+
+        assert_select ".container" do
+          assert_select "td", count: 0, text: api_only_app.name
+        end
+      end
     end
 
     context "organisation admin" do
--- a/test/controllers/account/manage_permissions_controller_test.rb
+++ b/test/controllers/account/manage_permissions_controller_test.rb
@@ -30,6 +30,20 @@ class Account::ManagePermissionsControllerTest < ActionController::TestCase
           assert_select "td", count: 1, text: non_delegatable_no_access_to_app.name
         end
       end
+
+      should "not list retired applications" do
+        user = create(:admin_user, email: "admin@gov.uk")
+        sign_in user
+
+        retired_app = create(:application, retired: true)
+        user.grant_application_signin_permission(retired_app)
+
+        get :show
+
+        assert_select ".container" do
+          assert_select "td", count: 0, text: retired_app.name
+        end
+      end
     end
 
     context "organisation admin" do
--- a/test/controllers/account/manage_permissions_controller_test.rb
+++ b/test/controllers/account/manage_permissions_controller_test.rb
@@ -106,6 +106,20 @@ class Account::ManagePermissionsControllerTest < ActionController::TestCase
           assert_select "td", count: 1, text: "GDS Admin"
         end
       end
+
+      should "not list API-only applications" do
+        user = create(:organisation_admin_user)
+        sign_in user
+
+        api_only_app = create(:application, api_only: true)
+        user.grant_application_signin_permission(api_only_app)
+
+        get :show
+
+        assert_select "#all-permissions" do
+          assert_select "td", count: 0, text: api_only_app.name
+        end
+      end
     end
 
     context "super organisation admin" do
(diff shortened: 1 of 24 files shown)


test/controllers/account/manage_permissions_controller_test.rb
```diff
@@ -30,6 +30,20 @@ class Account::ManagePermissionsControllerTest < ActionController::TestCase
...
+        user = create(:admin_user, email: "admin@gov.uk")
+        sign_in user
+
+        retired_app = create(:application, retired: true)
+        user.grant_application_signin_permission(retired_app)
+
+        get :show
+
+        assert_select ".container" do
+          assert_select "td", count: 0, text: retired_app.name
```
chrislo, line 44: Is it possible to drop the `text` parameter from this assertion? It might make the test even stronger as I don't think we should expect any `td` given the setup?

test/controllers/users_controller_test.rb
```diff
@@ -541,6 +541,22 @@ class UsersControllerTest < ActionController::TestCase
...
+
+          sign_in super_org_admin
+          user = create(:user_in_organisation, organisation: super_org_admin.organisation)
+          create(:user_application_permission, application: retired_app, user:)
+
+          get :edit, params: { id: user.id }
+
+          assert_select "h2", "All Permissions for this user"
+          assert_select "#all-permissions" do
+            assert_select "td", text: retired_app.name, count: 0
```
chrislo, line 557: Similar thought here re `text:`. Also there's a typo in the commit message `s/scenraio/scenario`.

chrislo (approved):
This all looks great to me @floehopper. I made a couple of small suggestions that I'm not even sure are an improvement, so feel free to ignore those!

test/controllers/users_controller_test.rb
floehopper: I see what you mean, but for me explicitly assert that we're not seeing the retired app in the table makes the intention of the test a bit clearer.

However, you have made me notice a couple of typos in the test names: `should "be not able to see permissions..." do` -> `should "not be able to see permissions..." do`

I'll fix those and the typos in the commit note too - thanks!

test/controllers/account/manage_permissions_controller_test.rb
floehopper: While you're probably right that the test setup means there won't be any rows in the table, I think the only thing I care about in this test is that the retired app doesn't show up. And I think asserting that explicitly makes the intention of the test clearer. So I think I'm going to leave this as it is.

test/controllers/users_controller_test.rb
floehopper: Addressed in https://github.com/alphagov/signon/pull/2452/commits/4ca60b45ed71a4e5bda51a2fb39681b80e018d59 & https://github.com/alphagov/signon/pull/2452/commits/689573bc4f65f33a00650528a5b9d4a0524499e8.

PR #2457, opened 2023-10-24
Hide permissions for API-only apps on users index page

floehopper:

Previously, despite #2452, we were still showing permissions for API-only apps in the permissions filter on the users index page. Joseph spotted that I had missed it. This PR should fix it!

### Before
Screenshot 2023-10-24 at 18 07 43

### After
Screenshot 2023-10-24 at 18 07 56


chrislo (approved):
Good catch 👍

PR #2458, opened 2023-10-24
Hide access tokens for retired apps

floehopper:
Trello: https://trello.com/c/kvmb5OHO, https://trello.com/c/pN8KOiQV & https://trello.com/c/huqPdMv8

This follows on from #2452 & #2446. It's main focus is to ensure that we're not using access tokens for retired apps anywhere, e.g.

* Access tokens for retired apps no longer appear on the API user edit page.
* No longer send SSO Push requests to retired apps. Addresses [this Trello card][1].
* No longer include token expiry times to Prometheus for retired apps. This should mean that 2nd Line Tech do not see alerts for such tokens. At least partly addresses [this Trello card][2].
* No longer sync app secrets to K8s for retired apps. This was actually changed in #2446, but I've added a test for it in this PR.

[1]: https://trello.com/c/pN8KOiQV
[2]: https://trello.com/c/huqPdMv8

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -56,7 +56,7 @@ class User < ApplicationRecord
   validate :exemption_from_2sv_data_is_complete
   validate :organisation_has_mandatory_2sv, on: :create
 
-  has_many :authorisations, class_name: "Doorkeeper::AccessToken", foreign_key: :resource_owner_id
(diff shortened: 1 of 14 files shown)


chrislo (approved):
Looks good to me!

PR #2460, opened 2023-10-25
Upgrade sidekiq to v6.5.10 to fix DoS vulnerability

floehopper:
See https://github.com/alphagov/signon/security/dependabot/70

For some reason Dependabot couldn't do this upgrade, so I've done it manually.


floehopper:
It turns out I did this wrong - closing!

PR #2464, opened 2023-10-25
Remove tasks and scripts that require a writeable filesystem

floehopper:
All these rake tasks and script try to write CSV file(s) to the local filesystem. However, on the containers on the EKS platform have a read-only filesystem. This means that these scripts cannot have worked for at least the best part of a year.  This in turn suggests that the data in the generated CSV files were not that important and so I think it's better to delete this unused code for now.

We already have an upcoming [Trello card][1] about collecting some metrics. I'll mention this code on that card in case there's some useful stuff in these CSV files.

[1]: https://trello.com/c/2SRZa6cA


test/lib/user_permissions_exporter_test.rb
```diff
@@ -1,82 +0,0 @@
-require "test_helper"
-
-class UserPermissionsExporterTest < ActionView::TestCase
-  def setup
-    @chips_org = create(:organisation, name: "Ministry of chips")
```
chrislo, line 5: Slightly sad to lose a reference to the "ministry of chips" 😂

chrislo (approved):
Great work tidying this up and explaining why they're safe to remove, thanks @floehopper! Having read the deleted code briefly I can't see anything that would solve the issue in https://trello.com/c/2SRZa6cA (tracking which users signed in to which apps), but it's good to know we can dig these out of the history if there are any patterns we want to borrow.

test/lib/user_permissions_exporter_test.rb
floehopper: 😢

PR #2471, opened 2023-10-27
Log application access

chrislo:
Trello: https://trello.com/c/2SRZa6cA

We've had some requests from other teams who want to know how often users log in to the applications they maintain.

We've decided to create an EventLog record each time a user is successfully authorized and an access token is created. This should be sufficient to answer the question "how many unique users have accessed a specific application in a given time period?".

I'll address the displaying of this data to users in a follow up PR once we've started collecting the data. At the moment the user-visible change as a consequence of this PR is the additional of a new event log item in the user's "access log":

Screenshot 2023-10-31 at 15 09 39

### DIFF
--- a/app/models/event_log.rb
+++ b/app/models/event_log.rb
@@ -49,6 +49,7 @@ class EventLog < ApplicationRecord
     ACCESS_TOKENS_DELETED = LogEntry.new(id: 44, description: "Access tokens deleted", require_uid: true),
     ACCOUNT_DELETED = LogEntry.new(id: 45, description: "Account deleted", require_uid: true),
     ORGANISATION_CHANGED = LogEntry.new(id: 46, description: "Organisation changed", require_uid: true, require_initiator: true),
+    SUCCESSFUL_USER_APPLICATION_AUTHORIZATION = LogEntry.new(id: 47, description: "Successful user application authorization", require_uid: true, require_application: true),
   ].freeze
 
   EVENTS_REQUIRING_UID = EVENTS.select(&:require_uid?)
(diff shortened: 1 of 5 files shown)


app/models/event_log.rb
```diff
@@ -49,9 +49,6 @@ class EventLog < ApplicationRecord
     ACCESS_TOKENS_DELETED = LogEntry.new(id: 44, description: "Access tokens deleted", require_uid: true),
     ACCOUNT_DELETED = LogEntry.new(id: 45, description: "Account deleted", require_uid: true),
     ORGANISATION_CHANGED = LogEntry.new(id: 46, description: "Organisation changed", require_uid: true, require_initiator: true),
-
-    # We no longer expire passwords, but we keep this event for history purposes
-    PASSWORD_EXPIRED = LogEntry.new(id: 6, description: "Password expired"),
```
floehopper, line 54: Thanks for cleaning this up - good catch!

floehopper (commented):
Typo in 2nd commit note: `/users.json` -> `/user.json`.

Using requests to the `/users.json` endpoint to record when a user is authorized to access an application seems rather indirect. While it may well not be the case at the moment, I can't see any reason a client app should not make requests to this endpoint whenever it feels like it.

I'm not 100% sure, but I think it might be better to hook into the creation of an access token, perhaps using [the Doorkeeper `after_successful_authorization` configuration option](https://github.com/doorkeeper-gem/doorkeeper/blob/v5.6.6/lib/generators/doorkeeper/templates/initializer.rb#L422-L442) if it's appropriate.

It might be worth having a chat about this!

chrislo:
@floehopper - I've re-written the history with the alternative approach we paired on together earlier.

floehopper (approved):
This looks great - good work! In particular both the unit & integration tests look fab!

Minor/pedantic: Change this text in the PR description:

> each time an access token is successfully authorized

To something like:

> each time a user is successfully authorized and an access token is created

i.e. make it clearer it's the user that is being authorized not the access token

chrislo:
Thank @floehopper! I've ammended the commit note and PR description and rebased on main before merging.

PR #2476, opened 2023-10-30
Remove redundant EKS redirect for application URIs

floehopper:
This effectively reverts the changes in #1911 which seem to relate to the migration of applications to the EKS platform. This migration happened about a year ago and as far as I can see the relevant env vars are not currently set:

* https://github.com/search?q=repo%3Aalphagov%2Fgovuk-helm-charts%20SIGNON_APPS_URI_SUB_PATTERN&type=code
* https://github.com/search?q=repo%3Aalphagov%2Fgovuk-helm-charts+SIGNON_APPS_URI_SUB_REPLACEMENT&type=code

In fact, it's not obvious the two env vars have *ever* been set:

* https://github.com/search?q=org%3Aalphagov+SIGNON_APPS_URI_SUB_PATTERN&type=code
* https://github.com/search?q=org%3Aalphagov+SIGNON_APPS_URI_SUB_REPLACEMENT&type=code

Either way, since they're not currently set, this functionality is redudant and unnecessarily complicating the code.

I've also removed the two env vars from the docs which were added more recently.

### DIFF
--- a/app/models/doorkeeper/application.rb
+++ b/app/models/doorkeeper/application.rb
@@ -66,26 +66,7 @@ class Doorkeeper::Application < ActiveRecord::Base # rubocop:disable Rails/Appli
     ].include?(name)
   end
 
-  def redirect_uri
-    substituted_uri(self[:redirect_uri])
-  end
-
-  def home_uri
-    substituted_uri(self[:home_uri])
-  end
-
 private
-
-  def substituted_uri(uri)
-    uri_pattern = Rails.configuration.oauth_apps_uri_sub_pattern
-    uri_sub = Rails.configuration.oauth_apps_uri_sub_replacement
-
-    if uri_pattern.present? && uri_sub.present?
-      uri.sub(uri_pattern, uri_sub)
-    else
-      uri
-    end
-  end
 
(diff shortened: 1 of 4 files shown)


sengi (commented):
I think we're still gonna need this for staging very soon. Right now we're still relying on some [hacky and brittle scripts](https://github.com/alphagov/govuk-puppet/blob/aa75027/modules/govuk_env_sync/files/govuk_env_sync.sh#L537-L573) that try to patch the production data before restoring the prod data into staging, but we'll be ditching those when we move the production->staging copy job to Kubernetes.

sengi:
FWIW there were definitely used in the past too: https://sourcegraph.com/search?q=repo:alphagov/govuk-%28infrastructure%7Chelm-charts%29+URI_SUB_PATTERN+type:diff

sengi:
I think what happened was we went a bit too far when cleaning up after the switch to Kubernetes — we probably should have left the staging config in there for clarity (even though the old, bodgy script was making it temporarily redundant).

floehopper:
@sengi OK - I'll close this - thanks for explaining

floehopper:
See #1407

PR #2479, opened 2023-10-31
Run brakeman in local development build

floehopper:
This is already running [as part of the CI build][1] and I've been caught out before when the local build passed, but the CI build failed which is sub-optimal. I copied the actual command from [the one that runs in the CI build][2].

Adding this to the `lint` rake task means that it will run when you run the default rake task locally, but it won't run twice in CI, because the latter doesn't run the default rake task.

[1]: https://github.com/alphagov/signon/blob/acae02f346e3880368f47622fcad2aa80558eb42/.github/workflows/ci.yml#L14-L16
[2]: https://github.com/alphagov/govuk-infrastructure/blob/6b6fe37cfd6af8ef57eff983e4356e96f7b3fd14/.github/workflows/brakeman.yml#L20C14-L20C71


chrislo (approved):
Great! Thank you for catching this.

PR #2484, opened 2023-11-01
Separate navigational and authorization concerns in UsersController

floehopper:
The main thing I'm trying to achieve in here is to separate navigational concerns into the controller and authorization concerns into the policy class.

My motivation for doing this is that in [moving the "edit user" page to use the GOV.UK Design System](https://trello.com/c/uZD2I9dj), we're planning to have a separate page for changing just the email and I want to re-use the `Account::EmailsController` for users changing *another* user's email. I'm planning to move/rename `Account::EmailsController` -> `UserEmailsController` and share it between two sets of routes, one under the `account` namespace and one under the `users` resource, e.g. we will end up with paths like:

* `/account/emails/edit` (which already exists; to be used when a user edits their own email)
* `/users/:user_id/emails/edit` (which doesn't currently exist; to be used when a user edits another user's email)

In this newly named `UserEmailsController`, I'll need to authorize requests both in the context of a user changing their own email *and* in the context of a user changing another user's email. I'm planning to re-use the `UserPolicy` to achieve that and the changes in this PR will make that easier.

I'm slightly nervous that this is going against the thinking in [this PR](https://github.com/alphagov/signon/pull/2377), in particular these two commits (https://github.com/alphagov/signon/commit/acad605426e87bd6c8cdd4f18da7f3da425ce57f & https://github.com/alphagov/signon/commit/bcde44227c48a7a29cba0d925d6a740587adeb00), but it feels like it moves us in the direction of using Pundit policies in a more idiomatic way.

To elaborate on that, in my mind the `UserPolicy` should represent the *application-wide* rules for which actions a given user can carry out on another user irrespective of which controller is being used to perform that action, i.e. the `User` in `UserPolicy` refers to the `User` class and instances of that class rather than to `UsersController`.

I think we've diverged from this approach recently, particularly with the controllers and corresponding policies in the `Account` namespace. This has led to a proliferation of policy classes that are *only* concerned with scenarios where a user is editing their own user, but that condition is not built into the policy class; it's just *implied* by the choice of policy in the relevant controller. This seems a bit dangerous to me, because a developer might miss this implicit condition and incorrectly apply that policy in an inappropriate context. Also it means we don't have a single canonical place which defines all the rules about which actions a given user can carry out on another (or their own) user.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -84,7 +84,7 @@ class UsersController < ApplicationController
 private
 
   def load_and_authorize_user
-    @user = current_user.normal? ? current_user : User.find(params[:id])
+    @user = User.find(params[:id])
     redirect_to(account_path) and return if current_user == @user && action_name == "edit"
 
     authorize @user
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -7,6 +7,7 @@ class UsersController < ApplicationController
 
   before_action :authenticate_user!
   before_action :load_user, except: %i[index]
+  before_action :redirect_to_account_page_if_acting_on_own_user, only: %i[edit]
   before_action :authorize_user, except: %i[index]
   before_action :allow_no_application_access, only: [:update]
   before_action :redirect_legacy_filters, only: [:index]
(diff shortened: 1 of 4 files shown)


chrislo (approved):
Thanks for taking me through this on a call too @floehopper. I think this is a good change and I agree with principal that the navigational concerns shouldn't be in the Pundit classes as much as possible. Looks good to me 👍

PR #2487, opened 2023-11-02
Split account update role & organisation page into two separate pages

floehopper:
Trello: https://trello.com/c/sxgHE1Xx

This main aim of this PR is to split the "update role & organisation" page into two separate pages one for updating your role and one for updating your organisation. This is in preparation for [moving the user edit page to use the GOV.UK Design System](https://trello.com/c/pws9UYJz) which will require a separate page for updating a user's role and another for updating a user's organisation.

Unlike in #2470, I haven't spend any time trying to sort out the validation errors. This is because as far as I can see there's not really a way to trigger a validation error on either page without hacking the submitted form. Note that if somehow a validation error does occur both pages just show a flash alert saying "There was a problem changing your role" or "There was a problem changing your organisation". This isn't inline with [these recommendations](https://design-system.service.gov.uk/patterns/validation/#how-to-tell-the-user-about-validation-errors) from the GOV.UK Design System and so we might want to have another look at them at some point.

I noticed that unlike for the change email/password PR, in this case there are no links on the dashboard sidebar, so I didn't have to make any changes there.

I am slightly concerned that the read-only version of these pages (where the user doesn't have permission to edit their role/organisation) look a bit empty, but I don't think it should block this being merged. On a similar front, I think it _might_ make sense to split the read-only version of the edit role and the edit organisation pages into separate `show` actions & templates. However, given the already significant scope of [moving the user edit page to use the GOV.UK Design System](https://trello.com/c/pws9UYJz), I've left this for now.

I also wondered whether the organisation edit page should allow you to "unset" your organisation, i.e. whether we should include a "None" option in the list like we do for the new invitation page. Either way, I think it would be good to remove the duplication between `UsersHelper#options_for_organisation_select` & `RoleOrganisationsHelper#options_for_your_organisation_select`. However, again I'm deeming that out-of-scope for this PR.

### Changes to account/settings page (for a user with permission to edit both role & organisation)
Screenshot 2023-11-02 at 14 47 53

### New update role page (for a user with permission to edit role)
Screenshot 2023-11-02 at 14 48 14

### New update organisation page (for a user with permission to edit organisation)
Screenshot 2023-11-02 at 14 48 31

### Changes to account/settings page (for a user without permission to edit either role or organisation)
Screenshot 2023-11-02 at 14 49 11

### New (read-only) update role page (for a user without permission to edit role)
Screenshot 2023-11-02 at 14 49 39

### New (read-only) update organisation page (for a user without permission to edit organisation)
Screenshot 2023-11-02 at 14 49 22


chrislo (approved):
This all looks good to me. I agree the "view" pages are a little spartan, but I also agree that can wait!

PR #2488, opened 2023-11-02
Add account application permissions edit page

chrislo:
Trello: https://trello.com/c/wVeHYiqF

This PR allows admins and publishing managers to update their own permissions via the `/account/applications` path. Publishing managers can only update their permissions if the application has delegatable permissions, if not they can view their permissions using the "View permissions" link that existed before this PR. 

With this PR we are able to remove the older non-design system "manage permissions" page.

## Link to update permissions 

Screenshot_2023-11-08_12-32-50

## Updating permissions

Screenshot_2023-11-08_12-33-01

## Success message

Screenshot_2023-11-08_12-33-10


app/controllers/account/permissions_controller.rb
```diff
@@ -14,6 +14,15 @@ def show
...
   def edit
     authorize [:account, @application], :edit_permissions?
+
+    @permissions = @application.sorted_supported_permissions_grantable_from_ui
+  end
+
+  def update
+    authorize [:account, @application], :edit_permissions?
+
+    current_user.replace_application_permissions(@application, params[:application][:permissions])
```
chrisroos, line 24: I think we need to use `UserUpdate` here because that already handles:

- adding/removing only the permissions that the user is allowed to modify (through the [`SupportedPermissionParameterFilter`](https://github.com/alphagov/signon/blob/de64c49d9017518421ad0844ef4ec811068a2c57/app/services/user_update.rb#L35))
- logging the addition/removal of permissions

I guess you'll need to merge all existing permissions with the permissions that the user has added/removed through the UI for this particular application. The [`Account::SigninPermissionsController#create`](https://github.com/alphagov/signon/blob/de64c49d9017518421ad0844ef4ec811068a2c57/app/controllers/account/signin_permissions_controller.rb#L9-L10) is doing something similar so might give you some clues.

app/controllers/account/permissions_controller.rb
```diff
@@ -22,6 +22,7 @@ def update
     authorize [:account, @application], :edit_permissions?
 
     current_user.replace_application_permissions(@application, params[:application][:permissions])
+    flash[:success] = @application.id
```
chrisroos, line 25: If we're going to store an ID in the flash then I think we should probably give it quite a specific key. Otherwise it's feasible that some other future action might populate the `success` with something other than an ID (e.g. with a string), and redirect to `account_applications_path` at which point the template will error in the `message_for_success` message. I half wonder whether there's also a risk that the application has been removed between populating the flash and then calling `message_for_success` in the subsequent request. If you don't want to store the full HTML in the flash then maybe storing the application name and set of permissions might be a bit more robust?

chrisroos (changes requested):
This mostly looks good but I've noted two things that I think we should change before merging. I'm very happy to talk through them or pair if that's easier.

app/controllers/account/permissions_controller.rb
chrislo: @chrisroos and I have paired on this and addressed this issue in [a524c11](https://github.com/alphagov/signon/pull/2488/commits/a524c11ff7ea94d980c29d74f6cdb0047e695a14)

chrisroos:
@chrislo - I've noticed that the flash message is a bit confusing if a user removes all their permissions for an app:

image

Maybe there's an argument for including the "signin" permission in that list so that there's always something to display? Or should we display a custom message when all non-signin permissions have been removed?

chrislo:
@chrisroos good catch. I've addressed that in [05d7afc](https://github.com/alphagov/signon/pull/2488/commits/05d7afcca5063913d88a07e624f7ae250f73fbaa)

floehopper:
Well done, both! 🎉

PR #2497, opened 2023-11-08
Add edit user name page

floehopper:


chrislo (approved):
This all looks good to me and I think moving a field at a time is a useful way of keeping the changes manageable. I checked out the branch locally and tested the error states and navigation on a couple of browsers and screen widths.

floehopper:
Thanks, @chrislo! 🙏

floehopper:
@chrisroos

I just noticed that I missed this test:

https://github.com/alphagov/signon/blob/55adcdf42e41d0ef9167def81a49a1f715e96ff8/test/controllers/users_controller_test.rb#L663-L668

And hence this behaviour:

https://github.com/alphagov/signon/blob/55adcdf42e41d0ef9167def81a49a1f715e96ff8/app/services/user_update.rb#L127-L129

When I inlined `UserUpdate` into `Users:NamesController#update` in this PR. I don't really understand why we'd want to send a permission update out to the users apps just because we've changed their name - probably just because this has got lumped into the `UserUpdate` mess. Anyway, I don't think it's a big deal, but thought I'd highlight it just in case something comes up about it next week.

chrisroos:
> @chrisroos
> 
> I just noticed that I missed this test:
> 
> https://github.com/alphagov/signon/blob/55adcdf42e41d0ef9167def81a49a1f715e96ff8/test/controllers/users_controller_test.rb#L663-L668
> 
> And hence this behaviour:
> 
> https://github.com/alphagov/signon/blob/55adcdf42e41d0ef9167def81a49a1f715e96ff8/app/services/user_update.rb#L127-L129
> 
> When I inlined `UserUpdate` into `Users:NamesController#update` in this PR. I don't really understand why we'd want to send a permission update out to the users apps just because we've changed their name - probably just because this has got lumped into the `UserUpdate` mess. Anyway, I don't think it's a big deal, but thought I'd highlight it just in case something comes up about it next week.

I've been doing some digging and I think we might want to update the client app when the name changes, in order to retain the previous behaviour.

Prior to this change, when a user's name was updated in Signon it would also be updated in the client app (e.g. Whitehall) very soon after (triggered by the [`PermissionUpdater`](https://github.com/alphagov/signon/blob/c59965aa6dd4846d11c44767a257c07257c6ea4c/app/jobs/permission_updater.rb)). After this change, I think the user's name will only be updated in the client app when they sign in again, or when the user stored in the session expires ([default expiry is 20 hours](https://github.com/alphagov/gds-sso/blob/3aa846de2d71c1f90c120546ade3595e0883c60c/lib/gds-sso/config.rb#L26)) and the user is authenticated again by gds-sso.

It looks as though gds-sso has supported the updating of name and email, along with permissions, since it was [first introduced in 2012](https://github.com/alphagov/gds-sso/commit/8c0888cda78b2d0323da26ff4c8da362074168d7). [Signon was updated to use this new user update endpoint](https://github.com/alphagov/signon/blob/f339848ac0c98188b0fc746935324f59215d26f2/lib/propagate_permissions.rb#L40) a few days later, and has also been sending name and email along with permissions, since then.

It feels as though the name of `PermissionsUpdater` is incorrect and that it should be something more like `UserUpdater` to indicate that it pushes user updates out to client apps.

I've captured the requirement for updating the name in client apps in [this Trello card](https://trello.com/c/2t9zx0Aq).

PR #2500, opened 2023-11-08
Change default name in User factory to be less confusing

floehopper:
I've seen the string "A name is now required" on a number of occasions when trying to debug a failing test and incorrectly assumed that it related to some kind of unexpected validation error which has wasted a chunk of time.

This changes the name to use a sequence a bit like the email attribute which should be a lot less confusing in similar circumstances. 🤞


chrisroos (approved):
I _might've_ been tempted to set it to `user-name` given that it doesn't need to be unique, but there's no harm in it using a sequence, and it's definitely better than the previous value.

floehopper:
> I _might've_ been tempted to set it to `user-name` given that it doesn't need to be unique, but there's no harm in it using a sequence, and it's definitely better than the previous value.

Thanks - it's been driving me nuts!

PR #2502, opened 2023-11-09
Fix redirect after admin updates another user's name

floehopper:
Trello: https://trello.com/c/DjdHEqxG

This was a mistake in #2497.

When the update completes successfully, the admin should be taken back to the "edit user" page for the user in question.

Previously they were redirected to the *non-existent* path for a non-existent `UsersController#show` action which was somehow (🤷) being handled by `UsersController#edit` as I intended!


floehopper:
> I suspect it's [this redirect in the routes](https://github.com/alphagov/signon/blob/de64c49d9017518421ad0844ef4ec811068a2c57/config/routes.rb#L97) that was making it work.

Ah, yes - I'd completely missed that! Thanks.

PR #2510, opened 2023-11-10
SPIKE: How we might be able to simplify/improve UserUpdate

floehopper:
Using `ActiveSupport::CurrentAttributes` means that we can have access to the current user in models; not just in controllers. This in turn means we ought to be able to move all of the logic in `UserUpdate` which is recording events in the activity log into model callbacks. These ought to be a lot more reliable in the sense that we don't need to keep remembering to use `UserUpdate` everywhere.

/cc @chrisroos

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,7 @@ class ApplicationController < ActionController::Base
 
   include TwoStepVerificationHelper
 
+  before_action :set_current_user_and_ip, if: :user_signed_in?
   before_action :handle_two_step_verification
   after_action :verify_authorized, unless: :devise_controller?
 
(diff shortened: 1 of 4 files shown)


chrisroos:
Thanks @floehopper. I hadn't seen `ActiveSupport::CurrentAttributes`! I can see that it's being [used in whitehall and e-petitions](https://github.com/search?q=org%3Aalphagov+CurrentAttributes+language%3ARuby&type=code) so this looks promising :-)

floehopper:
> I can see that it's being [used in whitehall and e-petitions](https://github.com/search?q=org%3Aalphagov+CurrentAttributes+language%3ARuby&type=code) so this looks promising :-)

Ah, that's good to know - thanks!

Closing for now, because I think it's served it's purpose.

PR #2513, opened 2023-11-13
Add help text to explain "Send push updates to this app"

chrisroos:
It took quite a bit of digging to understand what this setting does. Adding this will hopefully save someone else some time in future.

## Before

image

## After

image

### DIFF
--- a/app/views/doorkeeper_applications/edit.html.erb
+++ b/app/views/doorkeeper_applications/edit.html.erb
@@ -93,6 +93,14 @@
         <label>
           <%= f.check_box :supports_push_updates %> Send push updates to this app
         </label>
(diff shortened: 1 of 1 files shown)


chrislo (approved):
This looks useful to me @chrisroos 👍 

I wondered if it would be worth also linking to the documentation to further describe how those endpoints can be implemented. We have [this doc](https://docs.publishing.service.gov.uk/repos/signon/oauth.html) but it doesn't really describe those endpoints particularly clearly - perhaps better than nothing and we could beef it up later?

chrisroos:
Thanks @chrislo! I've amended the commit to include a link to the doc as you've suggested :+1:

PR #2523, opened 2023-11-16
Update client apps when user's name changes

chrisroos:
Trello - https://trello.com/c/2t9zx0Aq

Prior to https://github.com/alphagov/signon/pull/2497 changing a user's name in Signon would queue a job to push that update out to the applications the user has access to. This PR reinstates the previous behaviour by using `UserUpdate` in `Users::NamesController#update`, which in turn uses `PermissionUpdater` to push updates to the user's apps.

I can't find any documentation that describes the changes that result in push updates to a user's applications, but as far as I can see, changing a user's name has resulted in a push update since the relevant functionality was added to [Signon][1] and [gds-sso][2] in 2012 so it seems reasonable that we retain this behaviour.

I've also added a test to ensure we continue to push email updates out to a user's applications.

[1]: https://github.com/alphagov/signon/commit/f339848ac0c98188b0fc746935324f59215d26f2
[2]: https://github.com/alphagov/gds-sso/commit/8c0888cda78b2d0323da26ff4c8da362074168d7

### DIFF
--- a/app/controllers/users/names_controller.rb
+++ b/app/controllers/users/names_controller.rb
@@ -9,8 +9,8 @@ class Users::NamesController < ApplicationController
   def edit; end
 
   def update
-    if @user.update(user_params)
-      EventLog.record_event(@user, EventLog::ACCOUNT_UPDATED, initiator: current_user, ip_address: user_ip_address)
+    updater = UserUpdate.new(@user, user_params, current_user, user_ip_address)
+    if updater.call
       redirect_to edit_user_path(@user), notice: "Updated user #{@user.email} successfully"
     else
       render :edit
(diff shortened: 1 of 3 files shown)


floehopper (approved):
Good catch! I think adding the controller tests with expectations on `PermissioUpdater.perform_on` are particularly useful, because (a) they better document the somewhat hidden behaviour of `UserUpdate#call`; and (b) they should make it easier to refactor `UserUpdate` in the future.

I think the reason I missed this in #2497 was (a) because `UserUpdate` does so much but only a small part of it is relevant to this controller action, so it's difficult to see the wood for the trees; and (b) the name of `PermissionUpdater` implies that it only affects a user's permissions. These are both things it would be good to rectify as soon as possible to avoid similar confusion for others.

chrisroos:
Thanks, @floehopper! :+1:

PR #2529, opened 2023-11-20
Copy tweaks for account application permissions

chrislo:
Trello: https://trello.com/c/vInbSoSc

After reviewing the new account permissions pages we spotted a couple of small tweaks: 
- Change the wording of the "success" flash message
- Don't link to "update permissions" when there are no permissions to update.


chrislo:
Thanks @floehopper! I had another chat with Etain and we decided to remove the link entirely when there are no "additional" permissions to edit, which I think addresses your final 3 points. I'll have a look at converting the helper method to a partial in another PR.

PR #2536, opened 2023-11-21
Fixes for edit user name & edit user email form actions

floehopper:
The `#user_name_path` & `#user_email_path` route helper methods should require a `user` argument. However, the argument was not provided in either of the PRs where these forms were added (#2497 & #2509).

The assertions in `Users::NamesControllerTest` & `Users::EmailsControllerTest` were already comparing with the correct values and *somehow* the calls to `#user_name_path` & `#user_email_path` without any arguments were magically returning the correct value, even though when I try them in a Rails console, I get an `ActionController::UrlGenerationError`. So this mistake wasn't actually causing any problems.

### DIFF
--- a/app/views/users/emails/edit.html.erb
+++ b/app/views/users/emails/edit.html.erb
@@ -51,7 +51,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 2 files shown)


chrislo (approved):
Looks good to me!

PR #2537, opened 2023-11-21
Split out edit user role page from user email page

floehopper:


app/controllers/users/roles_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      current_user_role: current_user.role.to_sym,
+    ).sanitise
+  end
+
+  def permitted_user_params
+    @permitted_user_params ||= params.require(:user).permit(:role).to_h
+  end
+
+  def redirect_to_account_page_if_acting_on_own_user
+    redirect_to account_path if current_user == @user
```
chrisroos, line 36: Given that a superadmin user can edit their own role, I wonder whether it'd be better to redirect to `edit_account_role_path` here? Non superadmins will have already been rejected by `authorize_user` so we don't need to worry about them.

app/views/users/roles/edit.html.erb
```diff
@@ -46,22 +46,11 @@
...
-          <%= render "govuk_publishing_components/components/list", {
-            visible_counters: true,
-            items: [
-              "<strong>Superadmins</strong> can create and edit all user types and edit applications.",
-              "<strong>Admins</strong> can create and edit normal users.",
-              "<strong>Super Organisation Admins</strong> can unlock and unsuspend their organisation and related organisation accounts.",
-              "<strong>Organisation Admins</strong> can unlock and unsuspend their organisation accounts.",
-            ]
-          } %>
-        <% end %>
```
chrisroos: Not one for this PR but I wonder whether this highlights that we've missed this from the account edit role page?

app/controllers/users/roles_controller.rb
floehopper: That makes sense - I thought we'd changed the equivalent page for editing user email to redirect to the account page rather than the account edit email page, but I was mistaken. I will change this to match as you suggest.

app/controllers/users/roles_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+
+private
+
+  def load_user
+    @user = User.find(params[:user_id])
+  end
+
+  def authorize_user
+    authorize(@user)
+    authorize(@user, :assign_role?)
```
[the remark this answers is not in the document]
floehopper: I was trying to replicate the authorization that was happening in `UsersController` as closely as possible as a first step. As I think I mentioned in a commit note, I was imagining this could be simplified, but I decided against doing anything about it in this PR, because I think there are wider problems in this area.

I would argue against just getting rid of `authorize(@user)`, because it makes assumptions about the implementations of `UserPolicy#edit?`, `UserPolicy#update?` & `UserPolicy#assign_role?`.

I did contemplate calling `UserPolicy#edit?` from `UserPolicy#assign_role?` and then only calling `authorize(@user, :assign_role?)` here, but the `UserPolicy#assign_role?` method already feels like it's diverging from the idiomatic use of Pundit and I didn't want to make that worse.

I think I'd like to leave this for now and revisit the Pundit-related functionality for all these `Users::XXX` controllers when we've split out all the relevant pages from the "edit user" page. Does that make sense?

app/views/users/roles/edit.html.erb
floehopper: Yeah, maybe - although I think that page was more explicitly "designed", so it might have been intentional. In this case I've just tried to preserve the existing functionality/copy as closely as possible while still using the GOV.UK Design System.

app/controllers/users/roles_controller.rb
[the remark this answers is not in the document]
floehopper: @chrisroos & I have just discussed this and while we think there's definitely more work to do in this area, we've agreed that it's reasonable for it to be out-of-scope for this PR.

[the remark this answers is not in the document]
chrisroos: Yep, that makes sense to me :-)

app/views/users/roles/edit.html.erb
floehopper: Addressed in https://github.com/alphagov/signon/pull/2537/commits/a0e540a4a064eaff66e206892822649c278020a1

app/controllers/users/roles_controller.rb
floehopper: Addressed in https://github.com/alphagov/signon/pull/2537/commits/a0e540a4a064eaff66e206892822649c278020a1

floehopper:
Applied fixup commits and force-pushed in preparation for merging

PR #2540, opened 2023-11-22
Split out edit user organisation page from user edit page

floehopper:
Trello: https://trello.com/c/uZD2I9dj & https://trello.com/c/fA6dRVSm

This splits out the editing of another user's organisation into a separate page. The latest design calls for the splitting out the editing of a bunch of user fields into separate pages like this. The new "edit user organisation page uses the GOV.UK Design System, but the "edit user" page still does not. c.f. #2497,  #2509 & #2537.

The new "Change (organisation)" link on the "edit user" page only shows up if the current user is an Admin or a Superadmin, i.e. they have permission to change another user's organisation.

In my use of the [`error_summary` component](https://components.publishing.service.gov.uk/component-guide/error_summary) I have also brought the display of validation errors more into line with [this Design System guidance](https://design-system.service.gov.uk/patterns/validation/#how-to-tell-the-user-about-validation-errors). The most obvious validation error that can occur is if you attempt to change an Organisation Admin or Super Organisation Admin's organisation to None. Although I do wonder if this would be better tackled by excluding the "None" option from the select box in that scenario. However, I suggest we tackle that separately.

### New "Change" link on "edit user" page
Screenshot 2023-11-23 at 09 06 10

### New "edit user organisation" page
Screenshot 2023-11-23 at 09 06 28

### New "edit user organisation" page with validation error
Screenshot 2023-11-23 at 09 06 58

### Planned design for the "edit user" page (not part of this PR)
282090000-d8b41904-f74c-4896-8c35-3902beaebbbc

### DIFF
--- a/app/controllers/users/organisations_controller.rb
+++ b/app/controllers/users/organisations_controller.rb
@@ -1,0 +1,43 @@
+class Users::OrganisationsController < ApplicationController
+  before_action :authenticate_user!
+  before_action :load_user
+  before_action :authorize_user
+  before_action :redirect_to_account_page_if_acting_on_own_user, only: %i[edit]
+
+  def edit; end
+
+  def update
+    updater = UserUpdate.new(@user, user_params, current_user, user_ip_address)
+    if updater.call
+      redirect_to edit_user_path(@user), notice: "Updated user #{@user.email} successfully"
+    else
+      render :edit
+    end
+  end
+
+private
+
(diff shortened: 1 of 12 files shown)


chrisroos (approved):
This all looks good to me, @floehopper. Good work!

floehopper:
Thanks, @chrisroos!

PR #2545, opened 2023-11-23
Add separate page for resetting another user's 2SV

floehopper:
Trello: https://trello.com/c/uZD2I9dj & https://trello.com/c/caETStGa

This splits out the resetting of another user's 2SV into a separate page. The latest design calls for the splitting out of a bunch of user operations into separate pages like this. The new "reset user 2SV" page uses the GOV.UK Design System, but the "edit user" page still does not. c.f. https://github.com/alphagov/signon/pull/2497, https://github.com/alphagov/signon/pull/2509 & https://github.com/alphagov/signon/pull/2537.

Previously the "Reset 2-step verification" link on the "edit user" page was enhanced by UJS to prompt the user with a confirmation dialog and then submit a PATCH HTTP request to the controller. Since you now have to click the "Reset 2-step verification" link on the "edit user" page AND then click the "Reset 2-step verification" button on the new page, I don't think we need the confirmation dialog any more. However, I've used the `destructive` option on the button component to indicate that some information is lost when you click it.

All this means we no longer need to use UJS for this operation which is a plus for two reasons: (a) UJS is deprecated in Rails v7.0 and we're about to upgrade to Rails v7.1; and (b) this functionality did not work without JS enabled which was an accessibility issue.

Previously there wasn't anything stopping a hand-crafted request to `UsersController#reset_two_step_verification` from calling `User#reset_2sv!` & `UserMailer.two_step_reset` for a user that didn't have 2SV setup. And I haven't added any such check into the new controller. Since the "Reset 2-step verification" link on the "edit user" page still only shows up if the user has 2SV setup, it's not possible to _navigate_ to the new "Reset 2SV" page if the user doesn't have 2SV setup. However, I suppose it's now slightly easier to hack the URL to get to the new "Reset 2SV" page and submit the form. So it _might_ be worth protecting against this somehow, although I don't think it will do any harm if `User#reset_2sv!` & `UserMailer.two_step_reset` are called for a user that doesn't have 2SV setup - it might just be a bit confusing. What do you think?

### Unchanged "Reset 2-step verification" link on "edit user" page
Screenshot 2023-11-23 at 16 25 50

### New "reset 2SV" page
Screenshot 2023-11-23 at 16 26 05

### DIFF
--- a/test/controllers/users/two_step_verification_resets_controller_test.rb
+++ b/test/controllers/users/two_step_verification_resets_controller_test.rb
@@ -1,0 +1,173 @@
+require "test_helper"
+
+class Users::TwoStepVerificationResetsControllerTest < ActionController::TestCase
+  include ActiveJob::TestHelper
+
+  context "GET edit" do
+    context "signed in as Admin user" do
+      setup do
+        @admin = create(:admin_user)
+        sign_in(@admin)
+      end
+
+      should "display form with submit button & cancel link" do
+        user = create(:two_step_enabled_user)
+
+        get :edit, params: { user_id: user }
+
+        assert_template :edit
+        assert_select "form[action='#{user_two_step_verification_reset_path(user)}']" do
+          assert_select "button[type='submit']", text: "Reset 2-step verification"
+          assert_select "a[href='#{edit_user_path(user)}']", text: "Cancel"
+        end
+      end
+
+      should "authorize access if UserPolicy#reset_2sv? returns true" do
+        user = create(:two_step_enabled_user)
+
+        user_policy = stub_everything("user-policy", reset_2sv?: true)
+        UserPolicy.stubs(:new).returns(user_policy)
+
+        get :edit, params: { user_id: user }
+
+        assert_template :edit
+      end
+
+      should "not authorize access if UserPolicy#reset_2sv? returns false" do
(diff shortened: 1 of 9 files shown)


test/controllers/users/two_step_verification_resets_controller_test.rb
```diff
@@ -0,0 +1,159 @@
...
+        assert_select "form[action='#{user_two_step_verification_reset_path(user)}']" do
+          assert_select "button[type='submit']", text: "Reset 2-step verification"
+          assert_select "a[href='#{edit_user_path(user)}']", text: "Cancel"
+        end
+      end
+
+      should "authorize access if UserPolicy#reset_2sv? returns true" do
+        user = create(:two_step_enabled_user)
+
+        user_policy = stub_everything("user-policy", reset_2sv?: true)
```
chrislo, line 26: Yes, @chrisroos and I have done some similar stubbing in the branch we're working on. Seems better than testing all the combinations, like you say.

chrislo (approved):
> So it might be worth protecting against this somehow, although I don't think it will do any harm if `User#reset_2sv!` & `UserMailer.two_step_reset` are called for a user that doesn't have 2SV setup - it might just be a bit confusing. What do you think?

Agreed, I don't think it'll cause any issues.

PR #2547, opened 2023-11-24
Add separate page for mandating 2SV for another user

floehopper:
Trello: https://trello.com/c/uZD2I9dj & https://trello.com/c/caETStGa

This splits out mandating 2SV for another user into a separate page. The latest design calls for the splitting out of a bunch of user operations into separate pages like this. The new "mandate 2SV" page uses the GOV.UK Design System, but the "edit user" page still does not. c.f. https://github.com/alphagov/signon/pull/2497, https://github.com/alphagov/signon/pull/2509, https://github.com/alphagov/signon/pull/2537 & https://github.com/alphagov/signon/pull/2545.

Previously this functionality was implemented via a checkbox on the "edit user" page. This checkbox was only displayed when 2SV was not already mandated for the user either via an organisation mandate or an individual mandate, i.e. when `User#require_2sv?` was `false`. I've implemented the new page as just a button (no checkbox), because we only ever need to set `User#require_2sv` to `true`. If the user has a 2SV exemption, then some additional text is displayed in both the link and the form to explain that the exemption will be removed. Also in this case the `destructive` option on the button is set to `true` to highlight that the exemption data will be lost.

### Notes
* `User#require_2sv` can be set to `false` by adding a 2SV exemption for the user.
* For similar reasons as in this PR, we can probably get rid of the checkbox on the `UsersController#require_2sv` page which is part of the inviting a new user workflow.
* We can probably remove some duplication by combining `UsersController#require_2sv` page with the page in this PR (i.e. `Users::TwoStepVerificationMandations#edit`), but there were enough differences that I decided it was out-of-scope in this PR.

### New "Mandate 2SV" link on "edit user" page
Screenshot 2023-11-26 at 18 37 55

### New "mandate 2SV" page
Screenshot 2023-11-26 at 18 38 11

### New "Mandate 2SV" link on "edit user" page (when user has 2SV exemption)
Screenshot 2023-11-26 at 18 34 48

### New "mandate 2SV" page (when user has 2SV exemption)
Screenshot 2023-11-26 at 18 35 01

### DIFF
--- a/app/controllers/users/two_step_verification_mandations_controller.rb
+++ b/app/controllers/users/two_step_verification_mandations_controller.rb
@@ -1,0 +1,29 @@
+class Users::TwoStepVerificationMandationsController < ApplicationController
+  layout "admin_layout"
+
+  before_action :authenticate_user!
+  before_action :load_user
+  before_action :authorize_user
+
+  def edit; end
+
+  def update
+    user_params = { require_2sv: true }
+    updater = UserUpdate.new(@user, user_params, current_user, user_ip_address)
+    if updater.call
+      redirect_to edit_user_path(@user), notice: "Updated user #{@user.email} successfully"
+    else
+      render :edit
+    end
+  end
+
+private
+
+  def load_user
+    @user = User.find(params[:user_id])
+  end
+
+  def authorize_user
+    authorize(@user, :mandate_2sv?)
+  end
+end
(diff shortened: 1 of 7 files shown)


chrislo (commented):
Very minor: `s/separatly/separately` in 434430cbd commit note

floehopper:
I've made the copy changes requested by Etain in https://github.com/alphagov/signon/pull/2547/commits/d618833371929d49bbf9dc32fc6867ceb0fb793e.

Screenshot 2023-11-28 at 12 29 08

Screenshot 2023-11-28 at 12 27 42

Force-pushed in preparation for merging.

PR #2559, opened 2023-11-28
 Add separate page for resending signup email for another user

floehopper:
Trello: https://trello.com/c/uZD2I9dj & https://trello.com/c/K3oa6wAj

This splits out resending a signup email for another user into a separate page. The latest design for the "edit user" page calls for the splitting out of a bunch of user operations into separate pages like this. The new "resend signup email" page uses the GOV.UK Design System, but the "edit user" page still does not. c.f. https://github.com/alphagov/signon/pull/2497, https://github.com/alphagov/signon/pull/2509, https://github.com/alphagov/signon/pull/2537, https://github.com/alphagov/signon/pull/2545 & https://github.com/alphagov/signon/pull/2547.

Previously this functionality was implemented by a button on the "edit user" page, but now the button has been replaced by a link to a new "resend signup email" page. This page includes some explanatory text and a button to actually resend the signup email for another user.

Rather than further complicating the existing `InvitationsController`, I've moved `InvitationsController#resend` -> `Users::InvitationResendingsController#update`. Since `InvitationsController` was already significantly customizing the default behaviour of `Devise::InvitationsController`, I think this makes a lot of sense. See the commit titled "Extract page for resending invitation for a user" for more details.

I've also tidied up a couple of things along the way. Notably we now record an event in the activity log when we resend a signup email and we keep track of which admin user resent an invitation. See individual commit notes for details.

### "Resend signup email" link on "edit user" page
Screenshot 2023-11-28 at 17 02 18

### New "resend signup email" page
Screenshot 2023-11-28 at 14 31 36


app/views/users/invitation_resendings/edit.html.erb
```diff
@@ -0,0 +1,42 @@
...
+       }
+     ]
+   })
+%>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for @user, url: user_invitation_resending_path(@user) do %>
+      <%= render "govuk_publishing_components/components/hint", {
+        text: "Invitation not accepted yet. This user hasn't clicked on the link in their signup mail yet."
```
chrislo, line 31: Minor: I appreciate this text is copied from the previous flash/info message, but now it's a single paragraph the repeated "yet" feels a bit weird. I'd be tempted to drop the first "yet", i.e. : "Invitation not accepted. This user hasn't clicked on the link in their signup mail yet.". But you could also run this past Etain if you wanted, even after this PR is merged.

app/controllers/users/invitation_resendings_controller.rb
```diff
@@ -0,0 +1,33 @@
+class Users::InvitationResendingsController < ApplicationController
```
chrislo, line 1: Minor: I think `InvitationResendController` is a littler easier to read, but I don't feel strongly about it -  I appreciate it's a bit of an invasive change to make.

app/controllers/users/invitation_resendings_controller.rb
```diff
@@ -10,6 +10,7 @@ def edit; end
 
   def update
     @user.invite!
+    EventLog.record_account_invitation(@user, current_user)
```
chrislo, line 13: Good idea 👍

chrislo (approved):
All looks good to me! I made a couple of minor suggestions but none of them are important enough to block merging this.

app/controllers/users/invitation_resendings_controller.rb
floehopper: I dithered over this for ages, but you've convinced me. It'll be quite a faff, but I'll change it to `InvitationResendsController` (note the "Resend" will need to be plural).

floehopper: Addressed in https://github.com/alphagov/signon/pull/2559/commits/e23e19a4d353139f90d7e22e9dd9d49e9e7470bf.

app/views/users/invitation_resendings/edit.html.erb
floehopper: Addressed in https://github.com/alphagov/signon/pull/2559/commits/a6e562ce5dc98751d454987a2008e7f2227fd893.

PR #2561, opened 2023-11-29
 Add separate page for unlocking another user's account

floehopper:


app/controllers/users/unlockings_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  before_action :authenticate_user!
+  before_action :load_user
+  before_action :authorize_user
+  before_action :redirect_if_already_unlocked
+
+  def edit; end
+
+  def update
+    EventLog.record_event(@user, EventLog::MANUAL_ACCOUNT_UNLOCK, initiator: current_user, ip_address: user_ip_address)
```
chrislo: I appreciate you didn't introduce this but I wonder if the event log should happen after the call to `unlock_access!` in case the latter fails for some reason (now or with a future implementation).

chrislo (approved):

app/controllers/users/unlockings_controller.rb
floehopper: No worries - I should probably have mentioned in the earlier commit note that I was going to do that!

PR #2570, opened 2023-12-04
Fix breadcrumbs for pages shared with API users

floehopper:
The "access log" and "suspend/unsuspend user" pages are linked to from both the "edit user" page *AND* the "edit API user" page and so we should take that into account in the breadcrumb links to avoid confusing the user.

### New "access log" page
Screenshot 2023-12-04 at 17 03 27

### New "suspend/unsuspend user" page
Screenshot 2023-12-04 at 17 03 53

### DIFF
--- a/app/views/suspensions/edit.html.erb
+++ b/app/views/suspensions/edit.html.erb
@@ -8,8 +8,8 @@
          url: root_path,
        },
        {
-         title: "Users",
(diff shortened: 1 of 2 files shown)


chrisroos (approved):
Looks good to me, @floehopper. Is it worth adding a test to avoid this problem in future?

PR #2573, opened 2023-12-05
Increase number of minitest jobs in ci build

floehopper:


floehopper:
It looks like it takes ~1 min to "start" each Minitest job and increasing the number of jobs hasn't appreciably sped up the build as a whole. I think the main thing it might do is reduce the chances of one job having a lot of slow tests, i.e. it might make the build time more consistent, but I don't think that's worth it. So I'm closing this. cc @chrislo @chrisroos

PR #2582, opened 2023-12-07
Move `UpdateUser` methods into `after_update` callbacks on `User`

floehopper:
## Questions
1. Does this approach broadly make sense?
2. I'm fairly confident that moving the creation of `EventLog` records makes sense to do in callbacks on `User`, but I'm slightly less sure about sending email notifications. I think it still makes sense, but I'll probably need to be very careful that the logic is implemented correctly so that e.g. we don't end up sending out emails multiple times.
3. I think the user's permissions are going to be slightly more awkward to deal with, because everything else relates to attributes directly on the `User` record. However, I think we can either use callbacks on the `UserApplicationPermission` join model or implement some custom `ActiveModel::Dirty`-esque methods to detect changes to a user's permissions. The former would be conceptually simpler, but it would mean that we'd lose the ability to group permission changes together, so we'd end up e.g. creating multiple `EventLog` records (one for each user permission that's changed) and sending multiple SSO Push Updates out to each app. The latter might not be a problem though.

### DIFF
--- a/app/controllers/account/organisations_controller.rb
+++ b/app/controllers/account/organisations_controller.rb
@@ -10,7 +10,7 @@ class Account::OrganisationsController < ApplicationController
     organisation_id = params[:user][:organisation_id]
     organisation = Organisation.find(organisation_id)
 
-    if UserUpdate.new(current_user, { organisation_id: }, current_user, user_ip_address).call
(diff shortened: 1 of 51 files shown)


floehopper:
I'm going to try to get the bulk of this merged in #2623. Closing this PR for now.

PR #2587, opened 2023-12-11
Split manage permissions and manage tokens out of "edit API user" page

floehopper:

This is a small step towards moving this page to use the GOV.UK Design System. Splitting the page up like this should mean we can work on different bits of the page in parallel.

Note that I've added breadcrumbs for the page as part of this.

### "edit API user" page split into 3 parts

Screenshot 2023-12-11 at 12 06 36

### New "manage permissions" page
Screenshot 2023-12-11 at 12 19 32

### New "manage tokens" page
Screenshot 2023-12-11 at 12 20 53


chrislo (approved):
Looks good to me 👍

PR #2590, opened 2023-12-12
Extract `User#web_user?` predicate method

floehopper:
To be the inverse of `User#api_user?` and corresponding with the existing `User.web_users` scope.

We can use `User#web_user?` to avoid unnecessary inverse logic and make the code more readable by using `if user.web_user?` vs `unless user.api_user?` and `user.web_user?` vs `!user.api_user?`.

Also the callsite in `UserPolicy#exempt_from_two_step_verification?` was using the non-predicate attribute accessor which wasn't very idiomatic.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -285,7 +285,7 @@ class User < ApplicationRecord
   def status
     return USER_STATUS_SUSPENDED if suspended?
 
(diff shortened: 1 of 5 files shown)


chrisroos (approved):

PR #2591, opened 2023-12-12
Remove vestiges of `permissions_attributes` param

floehopper:
These relate to a time when the `User` model had `accepts_nested_attributes_for :permissions`. However, these were renamed to `:application_permissions` in [this commit][1] and removed entirely in [this commit][2] all as part of #311 which was merged 8 years ago!

The controller test about pushing permission changes out to apps wasn't relying on the `permissions_attributes` param being set. And none of the form fields under `app/views/api_users/*` use the name.

[1]: https://github.com/alphagov/signon/pull/311/commits/e2126f392b312a05263b5ca26c13e6c68ad002fc
[2]: https://github.com/alphagov/signon/pull/311/commits/d0df5b1b0569ffc08f04cace7e8665102188e138

### DIFF
--- a/app/controllers/api_users_controller.rb
+++ b/app/controllers/api_users_controller.rb
@@ -65,6 +65,6 @@ private
   end
 
   def permitted_user_params(params)
-    params.permit(:email, :name, permissions_attributes: {}, supported_permission_ids: [])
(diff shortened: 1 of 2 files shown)


chrisroos (approved):

PR #2592, opened 2023-12-12
Prevent Admins from suspending/unsuspending API users

floehopper:
Trello: https://trello.com/c/s9Qdhk0Y

Previously, `SuspensionsController#load_and_authorize_user` was only ever checking `UserPolicy#suspension?` to determine whether the current user was authorized to suspend/unsuspend an API user. This meant that although it hasn't been possible for Admins to *navigate* to the suspend/unsuspend user page for an API user, it would have been easy enough to hack the URL to reach the page and then there was nothing preventing them from suspending/unsuspending the API user.

I think this was an oversight when the `UserPolicy#suspension?` was introduced almost 9 years ago in https://github.com/alphagov/signon/pull/298. Since it was calling `User.find` it would never have returned an instance of an `ApiUser` and thus it would never have tried to use `ApiUserPolicy#suspension?` instead of `UserPolicy#suspension?`. I've implemented the former by aliasing it to `ApiUserPolicy#new?` as per all the other `ApiUser`-related predicates which means that only Superadmins have permission.

I've also hidden the suspend/unsuspend API user link on the edit API user page if the current user doesn't have the relevant permission (even though this doesn't make any practical difference) and added some missing test coverage that I missed in #2587.

### DIFF
--- a/app/controllers/suspensions_controller.rb
+++ b/app/controllers/suspensions_controller.rb
@@ -28,7 +28,9 @@ class SuspensionsController < ApplicationController
 private
 
   def load_and_authorize_user
-    @user = User.find(params[:id])
+    @user = ApiUser.find_by(id: params[:id]) || User.find_by(id: params[:id])
+    raise ActiveRecord::RecordNotFound if @user.blank?
+
     authorize @user, :suspension?
   end
 end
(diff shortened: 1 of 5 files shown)


chrislo (approved):
Good catch @floehopper 👍

PR #2567, opened 2023-11-30
Split "edit user" page and move top to GOV.UK Design System

floehopper:
Trello: https://trello.com/c/uZD2I9dj

This splits out the permissions-related functionality from the "edit user" page into a separate "manage permissions" page and then changes the "edit user" page to use the GOV.UK Design System. This is a culmination of a bunch of previous PRs that split out separate pages for various actions an admin can perform on another user, e.g. #2497 & #2509. This means that the "edit user" page no longer contains any forms, but it's now more like a "show user" page which displays various attributes of the user along with a bunch of links to pages for specific actions.

Although there are a *lot* of commits in this PR, almost all of them are small/mechanical preparatory refactorings intended to make the actual changeover relatively straightforward. The key commits are probably:

* [Split manage permissions page out of edit user page ](https://github.com/alphagov/signon/pull/2567/commits/5678e933ddec577fc7e3ad0e1b4bfa580673ec27)
* [Move edit user page to use GOV.UK Design System 🎉](https://github.com/alphagov/signon/pull/2567/commits/4590bb2de22c5e55f4344acc090ad475a1f26031)

And then there are a few commits at the end which make small copy improvements suggested by Etain, Joseph et al, and some tweaks to make things a bit more consistent. In a bid to keep this PR as small (!) as possible, I haven't implemented the copy changes which would've meant slightly more pervasive changes to keep everything consistent.

### Old "edit user" page
Screenshot 2023-12-04 at 16 02 13

### New "edit user" page
Screenshot 2023-12-04 at 15 45 19

### New "manage permissions" page
Screenshot 2023-12-04 at 15 47 29

### Old "suspend user" page
Screenshot 2023-12-04 at 16 02 33

### New "suspend user" page
Screenshot 2023-12-04 at 15 45 40

### Old "access log" page
Screenshot 2023-12-04 at 16 02 51

### New "access log" page
Screenshot 2023-12-04 at 15 46 51

PR #2593, opened 2023-12-12
Move "edit API user" page to use GOV.UK Design System

floehopper:
Trello: https://trello.com/c/bhPUU9qo

This moves what was (prior to #2587) the top of the "edit API user" page to use the GOV.UK Design System along similar lines to what we did for the "edit user" page in #2567.

The new "edit API user" page is in three sections:
* A [summary list](https://components.publishing.service.gov.uk/component-guide/summary_list) component displayed attributes of the API user and links to change some of them.
* A [list](https://components.publishing.service.gov.uk/component-guide/list) component containing action links.
* Buttons linking to the "Manage permissions" & "Manage tokens" pages.

I've reused the "change name" & "change user" pages, although they had to be adapted for use with API users rather than web users. At the suggestion of @chrisroos, I've made the relevant controllers (`Users::NamesController` & `Users::EmailsController`) available via routes under `/api_users` as well as under `/users`. As we discussed at the time, I think we should probably use this approach more widely for web-user- & API-user-specific pages, but in an effort to keep control of the scope of this PR, I've only tackled the controllers directly related to this PR.

I've also been able to reuse a bunch of methods in `UsersHelper`.

Although there are a lot of commits in this PR, many of them are small/mechanical preparatory refactorings intended to make the actual changeover relatively straightforward. The actual changeover happens in [this commit](https://github.com/alphagov/signon/pull/2593/commits/56f8b5ca0ed83a9d5aa53f8af47e9a79789a67f2).

### Old "edit API user" page
Screenshot 2023-12-13 at 13 44 12

### New "edit API user" page
Screenshot 2023-12-13 at 13 43 33

### Adapted "change name" page
Screenshot 2023-12-13 at 10 20 41

### Adapted "change email" page
Screenshot 2023-12-13 at 10 20 48


test/factories/users.rb
```diff
@@ -71,13 +71,15 @@
     expiry_date_for_2sv_exemption { (Time.zone.today + 1).to_date }
   end
 
-  factory :user_with_pending_email_change, parent: :user do
+  trait :with_pending_email_change do
```
chrisroos, line 74: I think it might be better to nest this, and the `invited_user` trait, inside the `:user` factory rather than making them top-level. Although they shouldn't cause any problem being at the top-level I think there's an implication that they're useful outside of the `user` factories and I'm not sure that's true in these cases.

I only discovered that top-level traits were possible when they broke in a recent upgrade of factory_bot - ee01a1b318aed3a961c8660f7d0a2110c32d66ac.

test/controllers/users/emails_controller_test.rb
```diff
@@ -57,8 +92,30 @@ class Users::EmailsControllerTest < ActionController::TestCase
         assert_not_authorised
       end
 
+      should "authorize access if ApiUserPolicy#edit? returns true when user is an API user" do
+        user = create(:api_user)
+
+        api_user_policy = stub_everything("api-user-policy", edit?: true)
+        ApiUserPolicy.stubs(:new).returns(api_user_policy)
```
chrisroos, line 99: It'd be good to consistently use `PunditHelpers#stub_policy` but I appreciate that it might be easier to make that change in one go outside of this PR.

app/services/user_update.rb
```diff
@@ -140,6 +140,8 @@ def record_email_change_and_notify
 
     EventLog.record_email_change(user, email_change.first, email_change.last, current_user)
 
+    return if user.api_user?
```
chrisroos, line 143: I know you're exercising this in controller tests but did you consider/is it worth unit testing it in `UserUpdateTest` too?

app/controllers/api_users_controller.rb
```diff
@@ -64,7 +63,8 @@ def api_user_params
     ).sanitise
   end
 
-  def permitted_user_params(params)
-    params.permit(:email, :name, supported_permission_ids: [])
+  def permitted_user_params(filtered_params)
```
chrisroos, line 66: If I understand correctly then the `create` action only allows `email` and `name` params, and the `update` action only allows `supported_permission_ids` param. If that is correct then it might be clearer to either specify those allowed  params in the actions, or split this into something like `permitted_user_params_for_create` and `permitted_user_params_for_update`.

app/controllers/users/emails_controller.rb
```diff
@@ -6,12 +6,14 @@ class Users::EmailsController < ApplicationController
   before_action :authorize_user
   before_action :redirect_to_account_page_if_acting_on_own_user, only: %i[edit]
 
+  helper_method :return_path
```
chrisroos, line 10: There's a typo in the message for the commit titled "Adapt Users::EmailsController to work with API users". The message refers to `Users::NamesController` but I think it should be `Users::EmailsController`.

chrisroos (approved):
This all looks good to me, @floehopper!

test/factories/users.rb
floehopper: I'd completely missed that - good catch!

test/controllers/users/emails_controller_test.rb
floehopper: Sure thing - I don't think it existed when I started this branch!

app/services/user_update.rb
floehopper: That's a reasonable request - I think I'd got lulled into not doing it because there aren't many such unit tests at the moment!

app/controllers/api_users_controller.rb
floehopper: Good idea!

app/controllers/users/emails_controller.rb
floehopper: Well spotted! Too much copy/pasting!

test/factories/users.rb
floehopper: As we just discussed in Slack, I need these traits to be available to the `api_user` factory and not just the `user` factory and so without introducing some kind of base class, I can't see how to do this. Marking as resolved.

app/services/user_update.rb
floehopper: Addressed in https://github.com/alphagov/signon/pull/2593/commits/77b6ce3c63a8068b131aa1bb2060025c512970fa

app/controllers/api_users_controller.rb
floehopper: I've remembered more about why I did this. The actions are actually using another method, `#api_user_params` which in turn calls this method, `#permitted_user_params`, so it's quite awkward to do what you're suggesting. I'll give it one more go...

floehopper: I think I've cracked it and in doing so spotted a minor security issue!

* https://github.com/alphagov/signon/pull/2593/commits/e74494c5aa1d48834a5c7d8b2612791571430d76
* https://github.com/alphagov/signon/pull/2593/commits/e9b3acbfc0ce31bea6f7b1151ca90f7d57c4f369
* https://github.com/alphagov/signon/pull/2593/commits/58d0433fceadefb114eae14996caf745e004f809
* https://github.com/alphagov/signon/pull/2593/commits/33bffd1fbc03143f78616da8cf1ce31d49788ade

app/controllers/users/emails_controller.rb
floehopper: Addressed in https://github.com/alphagov/signon/pull/2593/commits/7434b1abdc8fc842892f532cb7b3cc1d98d6c406

PR #2596, opened 2023-12-13
Tidy up users controllers

floehopper:
Just a few small tidying up commits for controllers under the `Users` namespace. See individual commit notes for details.

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,7 +1,29 @@
 {
   "ignored_warnings": [
-
+    {
+      "warning_type": "Mass Assignment",
+      "warning_code": 105,
+      "fingerprint": "767cb68f550b9a322a8dc311cf19b77e058eac5a546698967b4245c64cb8c32f",
+      "check_name": "PermitAttributes",
+      "message": "Potentially dangerous key allowed for mass assignment",
+      "file": "app/controllers/users/roles_controller.rb",
+      "line": 31,
+      "link": "https://brakemanscanner.org/docs/warning_types/mass_assignment/",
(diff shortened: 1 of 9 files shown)


config/brakeman.ignore
```diff
@@ -1,7 +1,29 @@
...
+    {
+      "warning_type": "Mass Assignment",
+      "warning_code": 105,
+      "fingerprint": "767cb68f550b9a322a8dc311cf19b77e058eac5a546698967b4245c64cb8c32f",
+      "check_name": "PermitAttributes",
+      "message": "Potentially dangerous key allowed for mass assignment",
+      "file": "app/controllers/users/roles_controller.rb",
+      "line": 31,
+      "link": "https://brakemanscanner.org/docs/warning_types/mass_assignment/",
+      "code": "params.require(:user).permit(:role)",
```
chrislo, line 12: Huh, Brakeman checks for this specific parameter name (`role`) which surprised me a bit! https://github.com/presidentbeef/brakeman/blob/391c95e6b8859934c7cb1ec8335b5c78af233dc9/lib/brakeman/checks/check_permit_attributes.rb#L8

chrislo (approved):
All looks like an improvement to me!

config/brakeman.ignore
floehopper: Yeah. I didn't really explain it very well in the commit note, but this isn't actually a problem in this scenario, because only Superadmins can update `User#role` through this controller.

Ironically, this commit doesn't change the behaviour, it just unobfuscates the code so Brakeman can detect the potential problem.

PR #2597, opened 2023-12-14
Consistently use PunditHelpers#stub_policy in controller tests

floehopper:
This tidies up a bunch of controller tests as follows:
* [Extract `PunditHelpers#stub_policy_for_navigation_links`](https://github.com/alphagov/signon/pull/2597/commits/ab2f9261c53d09cf5e85518b056c7ddd24b8474b).
* [Stub all the policies used in the navigation items](https://github.com/alphagov/signon/pull/2597/commits/71929fd102f4373740a3fdb65bbd18906095db3c).
* A bunch of commits changing controller tests to use `PunditHelpers#stub_policy` instead  of stubbing `UserPolicy.new` directly.

### DIFF
--- a/test/controllers/users/applications_controller_test.rb
+++ b/test/controllers/users/applications_controller_test.rb
@@ -254,10 +254,6 @@ class Users::ApplicationsControllerTest < ActionController::TestCase
 
 private
 
-  def stub_policy_for_navigation_links(current_user)
-    stub_policy(current_user, User, index?: true)
-  end
-
   def stub_user_application_permission(user, application)
     permission = UserApplicationPermission.new
     UserApplicationPermission.stubs(:for).with(user, application).returns(permission)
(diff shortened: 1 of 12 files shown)


chrisroos:
We're using `stub_policy` with a class to satisfy `NavigationItemsHelper#navigation_items` in [`ApplicationsControlllerTest#stub_policy_for_navigation_links`](https://github.com/alphagov/signon/blob/9df64e6c14f5c40d7d06a5066df93cc5e7e06ecf/test/controllers/users/applications_controller_test.rb#L257) so that scenario shouldn't require the fallback you've added.

> Another alternative would be to stub the authorize method on the controller. I haven't actually tried this, but it ought to be simpler and more robust and I think it should give us almost as much confidence as the current tests.

I don't think that will help in the cases where we want to stub a policy in order to test a conditional in a template.

chrisroos (approved):

PR #2598, opened 2023-12-14
Fix Users and Account::PermissionsController#update

chrisroos:
Trello: https://trello.com/c/09yMSdcU

We noticed some problems with the implementation of both `Users::PermissionsController#update` and `Account::PermissionsController#update`:

1. The signin permission would be removed unless explicitly added in the call to `update`
2. Permissions not grantable from the UI would be removed unless explicitly added in the call to `update`
3. Permissions not grantable from the UI could be added if explicitly added in the call to `update`
4. Permissions could be passed to `UserUpdate` for applications other than the application whose permissions were being viewed

See the individual commits for more detail.

We're fixing these problems before moving the API User edit page to use similar functionality.

### DIFF
--- a/app/controllers/account/permissions_controller.rb
+++ b/app/controllers/account/permissions_controller.rb
@@ -3,6 +3,7 @@ class Account::PermissionsController < ApplicationController
 
   before_action :authenticate_user!
   before_action :set_application
+  before_action :set_permissions
 
   def show
     authorize [:account, @application], :view_permissions?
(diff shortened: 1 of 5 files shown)


floehopper (approved):
The 3rd commit note "Fix issues in Users::PermissionsController#update" looks like it's got the remnants of a squashed commit at the bottom - it would be nice to tidy that up.

As I mentioned earlier today, I'm nervous about the complexity of `PunditHelpers#stub_policy` but I think that's something we can tackle separately.

Otherwise this all looks great to me - good work!

chrislo:
Thanks @floehopper - we've fixed that commit message. Rebasing on main before merging.

PR #2599, opened 2023-12-14
Manage permissions for api users

chrislo:


floehopper:
I think the breadcrumb link text could be terser, because later links are in the context of the previous links, so something like this:

Dashboard > API users > Test API User > Test API User's applications > Update Test API User's permissions for Test Application 1

Could become:

Dashboard > API users > Test API User > Applications > Update permissions for Test Application 1

I think this is more consistent with what we've done elsewhere.

test/integration/manage_api_users_test.rb
```diff
@@ -121,10 +123,12 @@ class ManageApiUsersTest < ActionDispatch::IntegrationTest
...
   end
 
-  def assert_has_signin_permission_for(application_name)
+  def assert_user_has_signin_permission_for(user, application_name)
+    assert user.has_access_to?(Doorkeeper::Application.find_by!(name: application_name))
+  end
+
+  def assert_has_access_token_for(application_name)
     within "table#editable-permissions" do
-      # The existence of the <tr> indicates that the API User has "singin"
```
floehopper, line 126: 🎶

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -0,0 +1,347 @@
...
+
+      current_user = create(:superadmin_user)
+      sign_in current_user
+
+      stub_policy current_user, api_user, update?: false
+
+      patch :update, params: { api_user_id: api_user, application_id: application, application: { supported_permission_ids: [] } }
+
+      assert_not_authorised
+    end
```
floehopper, line 342: Minor: It might be nice to move this test nearer the top of this context under the `should "prevent unauthenticated users"` test like in the context for the `#edit` action. At first I thought you'd missed out a test like this, because it wasn't where I expected it.

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -0,0 +1,347 @@
...
+      application = create(:application, api_only: true, with_supported_permissions: %w[permission])
+      api_user = create(:api_user)
+      create(:access_token, resource_owner_id: api_user.id, application:)
+
+      permission = application.supported_permissions.find_by(name: "permission")
+
+      assert_nothing_raised do
+        patch :update, params: { api_user_id: api_user, application_id: application, application: { supported_permission_ids: [permission.id] } }
+      end
+    end
```
floehopper, line 356: The name of this test and the setup somewhat implies to me that users can *only* edit API-only applications, but I'm not sure that's the case with the current implementation. If it wsa, then I'd probably expect to see a test demonstrating that editing permissions for a non-API-only app is not allowed.

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -343,5 +368,32 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
+        patch :update, params: { api_user_id: api_user, application_id: application, application: { supported_permission_ids: [] } }
+      end
+    end
+
+    should "include applications with revoked access tokens when there is at least one non-revoked access token" do
+      sign_in create(:superadmin_user)
+
+      application = create(:application, with_supported_permissions: %w[permission])
+      api_user = create(:api_user)
+      create(:access_token, resource_owner_id: api_user.id, application:, revoked_at: Time.zone.now)
```
floehopper, line 389: Minor: You can use `Time.current` instead of `Time.zone.now` - I thought rubocop enforced that, but I must've imagined it!

app/views/api_users/applications/index.html.erb
```diff
@@ -24,6 +24,13 @@
    })
 %>
 
+<% if flash[:application_id] %>
+<%= render "govuk_publishing_components/components/success_alert", {
+    message: "Permissions updated",
+    description: message_for_success(flash[:application_id], @api_user),
+} %>
+<% end %>
+
```
floehopper, line 33: I think it would be more consistent to put this in a `content_for(:custom_alerts)` block so that this appears above the heading, etc.

See:
* https://github.com/alphagov/signon/blob/ca17a91ad436f0c059f13a683fedafc14b9652fa/app/views/layouts/admin_layout.html.erb#L36
* https://github.com/alphagov/signon/blob/ca17a91ad436f0c059f13a683fedafc14b9652fa/app/views/api_users/manage_tokens.html.erb#L28-L37

app/controllers/api_users_controller.rb
```diff
@@ -38,17 +36,6 @@ def create
...
-    if @api_user.update(api_user_params_for_update)
-      @api_user.application_permissions.reload
-      PermissionUpdater.perform_on(@api_user)
-
-      redirect_to api_users_path, notice: "Updated API user #{@api_user.email} successfully"
-    else
-      render :manage_permissions
-    end
-  end
-
```
floehopper, line 51: 🎉 🎉 🎉

floehopper (approved):
This looks great - good work! I particularly like the change to `User#authorised_applications` and it feels really good to get rid of a bunch of code in `ApiUsersContoller`. However, while I understand why you've done it, I do have significant concerns about the amount of duplication that's been introduced. I worry that unless we do something about it now (when we're arguably in the best position to understand it all), it'll never get done.

Otherwise I've added a bunch of mostly minor comments. I think the ones I care most about are:
* Only testing editing permissions for API-only apps - this might just be a misunderstanding on my part
* Shortening the breadcrumb text
* Moving the flash alert

I think I might've asked you about this before but would it possible to use the [table component](https://components.publishing.service.gov.uk/component-guide/table) instead of using the lower-level CSS classes...?

Anyway, overall I think it looks good and I'm happy to approve it! ✅

chrisroos:
> I think the breadcrumb link text could be terser, because later links are in the context of the previous links, so something like this:
> 
> Dashboard > API users > Test API User > Test API User's applications > Update Test API User's permissions for Test Application 1
> 
> Could become:
> 
> Dashboard > API users > Test API User > Applications > Update permissions for Test Application 1
> 
> I think this is more consistent with what we've done elsewhere.

I agree that your suggestion is an improvement but these breadcrumbs for /api_users are consistent with /users so I think we'll leave this for now and tackle it when we address breadcrumbs site-wide in https://trello.com/c/dB0VDo9a.

test/controllers/api_users/permissions_controller_test.rb
chrisroos: We copied these tests from the equivalent users tests so inherited the problem from there. We're not going to move the test for now but think we might tackle this as part of trying to remove duplication between the 3 permissions controllers we've ended up with.

test/controllers/api_users/permissions_controller_test.rb
chrisroos: A quick search in the project suggests that we're using `Time.zone.now` in more places than `Time.current` so I'm not going to change this for now. Thanks for the heads up, though :+1:

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -0,0 +1,347 @@
...
+    should "push permission changes out to apps" do
+      sign_in create(:superadmin_user)
+
+      application = create(:application, with_supported_permissions: %w[permission])
+      api_user = create(:api_user)
+      create(:access_token, resource_owner_id: api_user.id, application:)
+
+      permission = application.supported_permissions.find_by(name: "permission")
+
+      PermissionUpdater.expects(:perform_on).with(api_user).once
```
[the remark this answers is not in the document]
chrisroos: Good spot. We copied this from an equivalent test in api_users_controller_test and didn't spot that unnecessary call. A quick search in the project reveals that there are quite a few instances of calling `#once` in our tests so I think I'll address this separately.

app/views/api_users/applications/index.html.erb
chrisroos: We've addressed this for /account, /users and /api_users in the commit titled, "Display permission flash notice in custom_alerts container".

chrisroos:
> However, while I understand why you've done it, I do have significant concerns about the amount of duplication that's been introduced. I worry that unless we do something about it now (when we're arguably in the best position to understand it all), it'll never get done.

Agreed! I think we should address this asap now that we've got 3 places (/account, /users and /api_users) all doing very similar things.

> I think I might've asked you about this before but would it possible to use the [table component](https://components.publishing.service.gov.uk/component-guide/table) instead of using the lower-level CSS classes...?

Good question! In this PR we've copied existing code which isn't using the table component but I don't know whether there are any reasons that we can't use it. We'll investigate using the table component outside of this PR.

chrisroos:
Thanks, @floehopper!

floehopper:
> I've added the following as checklist items on [this Trello card](https://trello.com/c/wSKG73MP):
> 
>     * Improve breadcrumbs as per [James's comment](https://github.com/alphagov/signon/pull/2599#issuecomment-1882883722)
> 
>     * Remove redundant calls to `Mocha::ObjectMethods#once` as per [James's comment](https://github.com/alphagov/signon/pull/2599#pullrequestreview-1810980907)
> 
>     * Use `Time.current` instead of `Time.zone.now` as per [James's comment](https://github.com/alphagov/signon/pull/2599#pullrequestreview-1810984127)
> 
>     * Consistently render flash messages in `content_for(:custom_alerts)` as per [James's comment](https://github.com/alphagov/signon/pull/2599#pullrequestreview-1810988174)
> 
>     * Use the table component as per [James's comment](https://github.com/alphagov/signon/pull/2599#pullrequestreview-1811008504)
> 
>     * Remove the duplication we've introduced as per [James's comment](https://github.com/alphagov/signon/pull/2599#pullrequestreview-1811008504)

Fab! 😍

PR #2620, opened 2024-01-03
Remove button for re-generating access token for API user

floehopper:
The "Re-generate" button revokes the access token and creates a new one for the same API user and application. However, in the meantime application(s) will continue to use the old (now revoked) access token and thus API requests will fail. A separate step is needed to sync the new valid access token to the relevant application container(s).

Screenshot 2024-01-03 at 11 37 02

I suspect this is why [the documentation][1] used by #govuk-2ndline-tech recommends first creating a new access token and only revoking the old one once the new one has been synced to the relevant application container(s) and is confirmed to be working OK.

I've also looked at the count of `EventLog` records in production which seems to confirm that the "Re-generate" button has rarely been used:
* `EventLog::ACCESS_TOKEN_GENERATED`: 88
* `EventLog::ACCESS_TOKEN_REVOKED`: 74
* `EventLog::ACCESS_TOKEN_REGENERATED`: 4

I asked about this in Slack and @theseanything confirmed that the button would not be missed, because it's still possible to achieve the same effect via the UI albeit with more clicks.

Ideally we'd be able to trigger the syncing of access tokens from within the Signon codebase and fully automate the procedure that 2nd Line use. However, that's a bigger piece of work and out-of-scope for the moment.

I'm about to [move the "Manage tokens for API user" page to use the GOV.UK Design System][2]. Doing this first will make that easier.

I've left the `EventLog::ACCESS_TOKEN_REGENERATED` constant defined in order to support the historical records. However, I've changed an unrelated test to use a different constant and added a comment to make it clear that this constant is deprecated.

[1]: https://docs.publishing.service.gov.uk/manual/alerts/signon-api-user-token-expires-soon.html
[2]: https://trello.com/c/75Jyg8zR

### DIFF
--- a/app/controllers/authorisations_controller.rb
+++ b/app/controllers/authorisations_controller.rb
@@ -29,19 +29,8 @@ class AuthorisationsController < ApplicationController
   def revoke
     authorisation = @api_user.authorisations.find(params[:id])
     if authorisation.revoke
-      if params[:regenerate]
-        regenerated_authorisation = @api_user.authorisations.create!(
-          expires_in: ApiUser::DEFAULT_TOKEN_LIFE,
-          application_id: authorisation.application_id,
-        )
-
-        EventLog.record_event(@api_user, EventLog::ACCESS_TOKEN_REGENERATED, initiator: current_user, application: authorisation.application, ip_address: user_ip_address)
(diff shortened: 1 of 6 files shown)


chrisroos (approved):

PR #2636, opened 2024-01-09
Replace Time.zone.now with Time.current

chrislo:
Arises from: #2599 

rubocop-govuk enables the "strict" style of the Rails/TimeZone cop[1]. This cop doesn't have a preference[2] for either of the equivalent methods `Time.zone.now` or `Time.current`. Since the latter is shorter and we're currently using a mixture of both I've gone with `Time.current`. I don't think we can enforce this without deviating from rubocop-govuk but hopefully being consistent now will help keep things that way.

[1]
https://github.com/alphagov/rubocop-govuk/blob/97fcbc866aa10a75c7c1c89fc3c2c562cc04fc26/config/rails.yml#L89 [2] https://github.com/rubocop/rubocop-rails/blob/cb01e4d6a33b0e4285b2972f9c64fc45e5776429/lib/rubocop/cop/rails/time_zone.rb#L26

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -183,7 +183,7 @@ class User < ApplicationRecord
   end
 
   def permissions_synced!(application)
-    application_permissions.where(application_id: application.id).update_all(last_synced_at: Time.zone.now)
+    application_permissions.where(application_id: application.id).update_all(last_synced_at: Time.current)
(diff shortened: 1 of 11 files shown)


floehopper (approved):
😍

PR #2638, opened 2024-01-10
Use table component on API users application index

chrislo:
Arises from: #2599 

To ensure we keep up with changes to the design system this PR replaces the table markup on `/app/views/api_users/applications/index.html.erb` with the `table` design system component.

## Before 
Screenshot 2024-01-10 at 15 08 28

## After
Screenshot 2024-01-10 at 15 08 13

### DIFF
--- a/app/views/api_users/applications/index.html.erb
+++ b/app/views/api_users/applications/index.html.erb
@@ -33,22 +33,18 @@
   <% end %>
 <% end %>
 
-<table class="govuk-table">
-  <caption class="govuk-table__caption govuk-table__caption--m">Apps <%= @api_user.name %> has access to</caption>
-  <thead class="govuk-table__head">
-    <tr class="govuk-table__row">
-      <th scope="col" class="govuk-table__header govuk-!-width-one-quarter">Name</th>
-      <th scope="col" class="govuk-table__header govuk-!-width-one-third">Description</th>
-      <th scope="col" class="govuk-table__header"><span class="govuk-visually-hidden">Permissions</span></th>
-    </tr>
-  </thead>
-  <tbody class="govuk-table__body">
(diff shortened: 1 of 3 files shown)


app/views/api_users/applications/index.html.erb
```diff
@@ -33,22 +33,18 @@
...
+    head: [
+      { text: "Name" },
+      { text: "Description" },
+      { text: content_tag(:span, "Permissions", class: "govuk-visually-hidden") },
+    ],
+    rows: @applications.map do |application|
+    [
+      { text: application.name },
+      { text: application.description },
+      { text: update_permissions_link(application, @api_user) || "" }
```
floehopper, line 47: Minor: Do you definitely need the trailing ` || ""`...? I'm wondering whether the `nil` might just magically turn into ""...?

floehopper (approved):
I added one very minor question/suggestion, but otherwise this looks good to me!

floehopper:
Oh, I just noticed the change in size of the table heading/caption. Is that consistent with what we've got elsewhere...?

app/views/api_users/applications/index.html.erb
chrislo: I think so, otherwise `nil` is replaced (by the component code) with `Not set` which I thought looked a bit confusing when combined with the visually hidden header for that column.

chrislo:
> Oh, I just noticed the change in size of the table heading/caption. Is that consistent with what we've got elsewhere...?

Yes, I think so - we're using the caption option for the table component in several other places in the code.

app/views/api_users/applications/index.html.erb
floehopper: Ah, yes - sorry for the noise!

PR #2643, opened 2024-01-11
Improve consistency of breadcrumbs

chrislo:
Trello: https://trello.com/c/dB0VDo9a

We've taken a look at the site as a whole and identified some places where the consistency of the breadcrumb trail can be improved. The main changes here are in the first two commits where we:
- Standardise on "Dashboard" as the first crumb in the trail
- Remove the final crumb from every trail to be consistent with the design system

The other changes are mostly small tweaks to make things more consistent.

### DIFF
--- a/app/views/account/activities/show.html.erb
+++ b/app/views/account/activities/show.html.erb
@@ -11,10 +11,7 @@
        {
          title: "Settings",
          url: account_path,
-       },
-       {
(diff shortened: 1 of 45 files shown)


floehopper (approved):
I haven't looked at all the breadcumbs across the site to check you haven't missed any, but all these changes make sense to me! 👍

PR #2648, opened 2024-01-15
Remove legacy layout

chrislo:
Trello: https://trello.com/c/kAZCL59w

This PR removes the old layout and related code now that the application has been migrated to the design system. 

The first two commits migrate an error page to the design system that we'd previously missed (see screenshots below for the before and after). We remove an unused route provided by Devise so that we don't have to migrate it to the design system. We then switch to using the design system as the default layout in the application and remove dead code. Finally we remove two gems that are no longer used in the application. 

## Error page before

error_before

## Error page after

error_after


floehopper (approved):
This makes me very happy! 🙌  Good work! 🎉 

This isn't anything you've done, but when I was looking at the old layout code you removed, I noticed that in [this commit](https://github.com/alphagov/signon/commit/6d910f1e58848286100bff655d9ea4d15e77f158) where the 2SV prompt page was converted to use the Design System layout and in [this commit](https://github.com/alphagov/signon/commit/5e43361755e806ec03c3ebf4443fce5db0f20340) where the new 2SV session page was converted to use the Design System layout, the `suppress_navbar_items` functionality was lost. I haven't looked into it, so I don't know how important that functionality was, but we should either (a) reinstate it; or (b) remove the relevant bits of code.

Similarly, in [this commit](https://github.com/alphagov/signon/commit/5e43361755e806ec03c3ebf4443fce5db0f20340) where all the `Devise::TwoStepVerificationSessionController` templates are switched to use the GOV.UK Design System layout, the `content_for(:thin_form)` seems to have been accidentally left in the `app/views/devise/two_step_verification_session/max_2sv_login_attempts_reached.html.erb` template. I'm pretty sure it can just be removed.

Also are you planning to remove the `govuk_admin_template` gem and all it's associated stuff in a separate PR? I wonder if doing that _might_ make more sense as part of this PR. What do you think...?

Anyway, I'm happy to mark this as approved and let you decide!

floehopper:
@chrislo:

Just looking at https://github.com/alphagov/signon/pull/2648/commits/8500c2100aed3fe8e601a2d854641a45fe842802, I'm a bit nervous that deleting `config/initializers/content_security_policy.rb` altogether might mean a regression in behaviour. I haven't looked into it in detail, but we might still need to call `GovukContentSecurityPolicy.configure` to get the default behaviour...?

Also related to the above, although unrelated `govuk_admin_template`, we are still using Rails UJS in a few places, e.g. [here](https://github.com/alphagov/signon/blob/6faa31579c47bed0b875e714f6f67a9a7932edbe/app/views/account/emails/edit.html.erb#L26-L27) and I have a suspicion that they rely on inline scripts. So we might need to keep some of the other config...?

> This isn't anything you've done, but when I was looking at the old layout code you removed, I noticed that in [this commit](https://github.com/alphagov/signon/commit/6d910f1e58848286100bff655d9ea4d15e77f158) where the 2SV prompt page was converted to use the Design System layout and in [this commit](https://github.com/alphagov/signon/commit/5e43361755e806ec03c3ebf4443fce5db0f20340) where the new 2SV session page was converted to use the Design System layout, the suppress_navbar_items functionality was lost. I haven't looked into it, so I don't know how important that functionality was, but we should either (a) reinstate it; or (b) remove the relevant bits of code.

Did you have any luck working out what to do about this? ☝️

floehopper:
Further to my point about Content Security Policy and given that we don't report on JS exceptions from production environments, I wonder if it might be sensible to deploy this branch to integration and check that a few things (like those Rails UJS links) are working and nothing is raising exceptions...?

chrislo:
After discussing with @floehopper I've removed the content security changes, we can see if they're still required and relevant if we decide to port the remaining uses of UJS to turbo (or remove them in favour of not using js at all) as part of the rails upgrade. 

There's still some places where we use `suppress_navbar_items` so I'm not going to tackle changing that in this PR.

chrislo:
Rebasing this on `main` before merging.

PR #2652, opened 2024-01-17
Add link to GOV.UK status page to footer

floehopper:
c.f. https://github.com/alphagov/whitehall/pull/8727

I considered using `Plek.external_url_for` to generate the URL based for the "status" app (like the code is already doing for the "support" app), but it doesn't look like there's an equivalent of that app in any other environments, so it didn't seem worthwhile.

Screenshot 2024-01-17 at 10 09 31


davidtrussler (approved):
LGTM! 👍

PR #2657, opened 2024-01-22
Refactor build user update params

chrislo:
Trello: https://trello.com/c/yDM25WlZ

This PR is an attempt to remove some of the duplication we added in c7236440f, edf2e83e and 92c5f5c0 with the `build_user_update_params` controller method. I did consider folding some of this logic into `UserUpdate` but as we're currently thinking about simplifying or removing that altogether (e.g. #2623) I think for now it's an improvement to have the code shared between the three controllers via a simple Ruby object. It'll at least make it easier to move somewhere else in the future.

### DIFF
--- a/app/controllers/account/permissions_controller.rb
+++ b/app/controllers/account/permissions_controller.rb
@@ -18,7 +18,7 @@ class Account::PermissionsController < ApplicationController
   def update
     authorize [:account, @application], :edit_permissions?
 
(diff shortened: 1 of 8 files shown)


chrisroos (approved):

chrislo:
Thanks @chrisroos - rebasing before merging.

PR #2664, opened 2024-01-25
Refactor applications tables

chrislo:
Trello: https://trello.com/c/yDM25WlZ

This PR removes some duplication of logic in the views that render the `/account/applications`, `/users/:id/applications` and `/api_users/:id/applications` pages. It also removes the hard-coded HTML and instead relies on the `govuk_publishing_components` `table` component to generate the markup. 

In the final commit I introduce a small CSS override modelled on a similar pattern used in Whitehall to align the final "actions" column of the table to the right. 

There are some small visual differences after this refactor (notably we've lost the ability to specify the widths of the first two columns of the tables), but I've gone over these with Calum and we're happy that the changes introduced are acceptable given the overall improvement to consistency and benefits of using the table component. 

## /account/applications After

Screenshot_2024-01-31_11-14-27

## /account/applications Before

Screenshot_2024-01-31_11-15-03

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -85,3 +85,12 @@ $govuk-page-width: 1140px;
     float: right;
   }
 }
+
+.govuk-table--with-actions {
+  .govuk-table__cell:last-child {
+    display: flex;
+    justify-content: flex-end;
+    align-items: center;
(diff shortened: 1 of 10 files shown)


chrisroos (approved):
This looks good to me, @chrislo! Nice work :+1:

PR #2684, opened 2024-02-05
Content changes

chrisroos:
Trello: https://trello.com/c/GFazegOU and https://trello.com/c/W1L8p8d9

Etain and I paired on making these content changes. See the individual commits for more information.

### DIFF
--- a/test/models/user_mailer_test.rb
+++ b/test/models/user_mailer_test.rb
@@ -178,11 +178,7 @@ class UserMailerTest < ActionMailer::TestCase
     end
 
     should "mention that reset is disallowed because their account is suspended" do
-      assert_body_includes "You can't request a password reset on a suspended account."
-    end
-
-    should "include correct support links" do
-      assert_support_present_in_text "support form"
+      assert_body_includes "You cannot reset the GOV.UK Signon development password for user@example.com while the account is suspended."
     end
   end
 end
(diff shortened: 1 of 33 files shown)


test/models/user_mailer_test.rb
```diff
@@ -178,11 +178,7 @@ def assert_support_present_in_text(link_text, email = @email)
     end
 
     should "mention that reset is disallowed because their account is suspended" do
-      assert_body_includes "You can't request a password reset on a suspended account."
-    end
-
-    should "include correct support links" do
-      assert_support_present_in_text "support form"
+      assert_body_includes "You cannot reset your GOV.UK Signon development account, for user@example.com while your account is suspended."
```
chrislo, line 181: Does this need an additional comma after the email address, i.e.:

> You cannot reset your GOV.UK Signon development account, for user@example.com, while your account is suspended.

?

chrislo (approved):
Looks good to me - a small query on the copy in one commit, but I'll let you both decide whether to address that.

test/models/user_mailer_test.rb
EtainNi: Yes, although on second thoughts it would read better as "You cannot reset the GOV.UK Signon password for user@example.com while the account is suspended." So remove all commas, add the word 'password' and change 'your' to 'the'. Sorry!

chrisroos: Thanks both! I've updated the text as requested.

PR #2689, opened 2024-02-06
Use table component everywhere

CristinaRO:
Part of https://trello.com/c/r53vhSZn/408-add-information-into-aria-to-assist-screen-reader-users

While looking into adding ARIA elements for tables, we realised that signon has three ways of rendering tables - hardcoded, using the govuk table component, and using the custom signon table component (which is derived from the govuk one).

This preliminary work seeks to unify the approach to tables, and use the signon table component everywhere.

Some formatting is lost, as illustrated in the attached screenshots. (No way to make the header three quarters width.)

### Before
Screenshot 2024-02-06 at 17 10 18

### After
Screenshot 2024-02-06 at 16 29 39

### Before
Screenshot 2024-02-06 at 17 10 10

### After
Screenshot 2024-02-06 at 16 29 12


test/helpers/application_helper_test.rb
```diff
@@ -45,4 +45,12 @@ class ApplicationHelperTest < ActionView::TestCase
...
     end
   end
+
+  test "#govuk_tag returns a strong tag with the given text, and the class govuk-tag" do
+    assert_equal "<strong class=\"govuk-tag\">something</strong>", govuk_tag("something")
+  end
+
+  test "#govuk_tag returns a strong tag with the given text, and the classes govuk-tag plus any provided classes" do
+    assert_equal "<strong class=\"govuk-tag some-class another--class\">something</strong>", govuk_tag("something", "some-class another--class")
+  end
```
yndajas, line 55: Are you allowing the `classes` argument to be an array (given the `.compact`)? If so, is it worth adding a test for when an array is passed in?

yndajas (approved):
Just one suggestion. Generally looks good!

test/helpers/application_helper_test.rb
CristinaRO: No, the intention was to allow a string of space-separated classes, or nil, and the compact is there to protect against the default nil value of `classes`.

CristinaRO: I wonder if this is making it too flexible, when we don't yet have any other uses for it.

I also notice that none of the other, newer, tables are using the tag class for their Yes/No strings. So we might want to remove it altogether...

yndajas: Ah, my Ruby is rusty: I'm confusing `.compact` and `.flat`!

PR #2694, opened 2024-02-07
Remove cookie rotate code

chrislo:
We introduced this cookie rotation code in e711982abbf as part of the upgrade to Rails 7. Since our session cookies do not have a `Max-Age` or `Expires` attribute they are treated by the browsers as Session cookies and expire when the browser window is closed. I think it's safe to assume that all cookies issued under the old scheme (sometime before approximately April 13th 2022 when e711982abbf was made) have now been rotated, so this code can be removed.


chrisroos (approved):
Looks good to me, @chrislo

PR #2713, opened 2024-02-13
Show user status as tag instead of strikethrough

CristinaRO:
The strikethrough is not part of the design system, so we've decided to use the govuk-tag style instead.

The design on the card showed green for active users, grey for suspended users, but there were no examples of locked or invited users.

I've chosen to only use green for active users, and grey for all other statuses.

[Trello](https://trello.com/c/ZW8yl3g8/280-review-discuss-other-options-possible-for-displaying-suspended-users-rather-than-strikethrough-since-this-isnt-inline-with-the-d)

### Users - before
Screenshot 2024-02-13 at 20 10 33

### Users - after
Screenshot 2024-02-13 at 19 28 38

### API users - before
Screenshot 2024-02-14 at 17 00 25

### API users - after
Screenshot 2024-02-15 at 12 41 14

### DIFF
--- a/app/helpers/api_users_helper.rb
+++ b/app/helpers/api_users_helper.rb
@@ -4,8 +4,7 @@ module ApiUsersHelper
   end
 
   def api_user_name(user)
-    anchor_tag = link_to(user.name, edit_api_user_path(user), class: "govuk-link")
-    user.suspended? ? content_tag(:del, anchor_tag) : anchor_tag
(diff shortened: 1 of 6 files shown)


chrislo (approved):
Looks like you might need to make `govuk_tag` available in the tests but otherwise this looks good to me 👍

chrislo (approved):
Looks good to me!

PR #2717, opened 2024-02-14
Improve User validation error messages

chrisroos:
This PR moves the remaining validation error strings from the `User` class into the `en.yml` locale file. This should make it easier to see at a glance the differences in style between the `User` validation error messages (i.e. some are of the form "Enter an email" and others are of the form "Email can't contain non-ascii characters").

Summary of commits:

- Use `assert_includes` when checking validation error messages in users_test.rb
- Correct a mistake in the validation error when `User#reason_for_2sv_exemption` is present but shouldn't be
- Add tests for 3 validation error messages that weren't tested
- Move remaining validation error messages from strings in `User` to the `en.yml` locale file

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -449,7 +449,7 @@ private
   end
 
   def organisation_has_mandatory_2sv
-    errors.add(:require_2sv, "2-step verification is mandatory for all users from this organisation") if organisation && organisation.require_2sv? && !require_2sv
+    errors.add(:require_2sv, :blank) if organisation && organisation.require_2sv? && !require_2sv
   end
 
   def fix_apostrophe_in_email
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -430,7 +430,7 @@ private
   end
 
   def user_can_be_exempted_from_2sv
-    errors.add(:reason_for_2sv_exemption, "cannot be present for #{role_display_name} users. Remove the user's exemption to change their role.") if exempt_from_2sv? && role.require_2sv?
+    errors.add(:reason_for_2sv_exemption, :present, role_display_name:) if exempt_from_2sv? && role.require_2sv?
   end
 
   def organisation_admin_belongs_to_organisation
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -430,7 +430,7 @@ private
   end
 
   def user_can_be_exempted_from_2sv
-    errors.add(:reason_for_2sv_exemption, "cannot be blank for #{role_display_name} users. Remove the user's exemption to change their role.") if exempt_from_2sv? && role.require_2sv?
+    errors.add(:reason_for_2sv_exemption, "cannot be present for #{role_display_name} users. Remove the user's exemption to change their role.") if exempt_from_2sv? && role.require_2sv?
   end
 
   def organisation_admin_belongs_to_organisation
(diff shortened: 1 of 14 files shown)


app/models/user.rb
```diff
@@ -435,7 +435,7 @@ def user_can_be_exempted_from_2sv
 
   def organisation_admin_belongs_to_organisation
     if publishing_manager? && organisation_id.blank?
-      errors.add(:organisation_id, "can't be 'None' for #{role_display_name}")
+      errors.add(:organisation_id, :blank, role_display_name:)
```
floehopper, line 438: Nice!

floehopper (commented):
I'm probably being dim, but I don't understand how Rails is distinguishing between validation messages of the form "Enter an email" and those of the form "Email can't contain non-ascii characters". In the former case, it doesn't need to prefix the message with the humanized attribute name, but in the latter it does. How does it tell the difference?

It's been a long day and I'm tired, so apologies if this is really obvious!

chrisroos:
> I'm probably being dim, but I don't understand how Rails is distinguishing between validation messages of the form "Enter an email" and those of the form "Email can't contain non-ascii characters". In the former case, it doesn't need to prefix the message with the humanized attribute name, but in the latter it does. How does it tell the difference?

Rails isn't distinguishing. We have to do that in the templates where these messages are being displayed, e.g. 1eb45b8ce50223816ff7ebbe77d997310a1b2576. I mentioned in standup today that we're now in a bit of an inconsistent state with error messages because some will require us to use `full_message` (i.e. attribute + message) and others will require us to use `message`. Joseph suggested that we might want to change them all in one go so that they're in the new format but that will have to wait. Having typed that response I realise that I don't understand where this new format has come from and whether it justifies the work required to make everything consistent.

floehopper:
> > I'm probably being dim, but I don't understand how Rails is distinguishing between validation messages of the form "Enter an email" and those of the form "Email can't contain non-ascii characters". In the former case, it doesn't need to prefix the message with the humanized attribute name, but in the latter it does. How does it tell the difference?
> 
> Rails isn't distinguishing. We have to do that in the templates where these messages are being displayed, e.g. [1eb45b8](https://github.com/alphagov/signon/commit/1eb45b8ce50223816ff7ebbe77d997310a1b2576). I mentioned in standup today that we're now in a bit of an inconsistent state with error messages because some will require us to use `full_message` (i.e. attribute + message) and others will require us to use `message`. Joseph suggested that we might want to change them all in one go so that they're in the new format but that will have to wait.

Thanks for explaining - that makes more sense now.

> Having typed that response I realise that I don't understand where this new format has come from and whether it justifies the work required to make everything consistent.

I suspect it comes from [this documenation](https://design-system.service.gov.uk/components/error-message/#use-instructions-and-descriptions) from the GOV.UK Design System.

chrisroos:
Hi @floehopper - Is there anything else you think I should do before this can be approved/merged?

floehopper:
> Hi @floehopper - Is there anything else you think I should do before this can be approved/merged?

I don't think so - I'll mark it as approved now.

chrisroos:
This PR was an attempt to fix the inconsistencies introduced by changing some of the User validation error messages in 1eb45b8ce50223816ff7ebbe77d997310a1b2576.

The intention was to:

- move all User validation error messages to the en.yml locale file
- make each message self-contained (so that it didn't require the attribute to be prefixed to the error for it to make sense)
- change the format of `full_message` so that it didn't include the attribute prefix
- update the templates so that they used `full_message` everywhere

I think it's probably fairly close to being complete but I've run out of time so have instead ended up reverting the change that introduced the problem (see PR #2736). 

Closing.

PR #2719, opened 2024-02-15
Add new permissions form layout

yndajas:
[Trello](https://trello.com/c/xzUooeJS)

This adds a new layout for editing permissions when there are over eight supported permissions for a given app. This applies to the account path page for editing your own permissions and the user path page for editing another user's permissions, but not the API user path page

I tried writing some Jasmine tests for this, but came across issues when trying to select elements. Given the majority of this is from a library, I wonder if the JavaScript testing is really necessary. There don't appear to be examples of other `alphagov` repositories testing the accessible autocomplete functionality. I thought it might be good to test the clear button functionality, but perhaps it's unnecessary. The underlying select box functionality is integration tested

### Current layout - now the layout for 8 or fewer permissions

image

### New layout for 9 or more permissions

image

### Error when trying to add a permission without selecting one

image


app/controllers/account/permissions_controller.rb
```diff
@@ -13,15 +13,46 @@ def show
...
   end
 
   def update
     authorize [:account, @application], :edit_permissions?
 
+    selected_permission_ids = []
+
+    if update_params[:supported_permission_ids]
+      selected_permission_ids = update_params[:supported_permission_ids]
+    elsif update_params[:new_permission_id]&.length&.> 0
```
Gweaton, line 44: Would `length` here ever return `nil`? Can we remove the `&. > 0` or am I misunderstanding here?

Gweaton: I've found this condition a little tricky to read, though I know it's hard as I've spent half the day trying to come up with an alternative 😅...

I don't know if this is any clearer, I initially tried an early return but couldn't get it to work:

```
    selected_permission_ids = update_params[:supported_permission_ids] || []

    if selected_permission_ids.empty?
      if update_params[:new_permission_id]&.length > 0
        selected_permission_ids.concat(update_params[:current_permission_ids] || [], [update_params[:new_permission_id]])
      else
        flash[:alert] = "You must select a permission."
        redirect_to edit_account_application_permissions_path(@application)
        return
      end
    end

```

Gweaton (commented):
Looks like this was hard work - well done on it. I've left a few comments, happy to discuss them on Monday.

That large commit in the middle has taken me a while to review, as there was a lot going on so hopefully I've understood what it's doing there. I think the approach is solid, but could possibly be a bit more readable (although I don't know if my suggestion makes all that much difference!)

I know we mentioned the Jasmine tests possibly not being necessary, but I wonder if it's worth trying to get them into shape for the custom behaviour we've added to the autocomplete, rather than the standard behaviour we get from the component - sorry, I must have misunderstood this yesterday.

app/controllers/account/permissions_controller.rb
```diff
@@ -13,15 +13,46 @@ def show
...
+    @shared_permissions_form_locals = {
+      action: account_application_permissions_path(@application),
+      application: @application,
+      cancel_path: account_applications_path,
+      user: current_user,
+    }
+
+    if @permissions.count > 8
+      @current_permissions = []
+      @new_permission_options = []
```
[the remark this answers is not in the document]
yndajas: Assigned and unassigned

Gweaton (approved):
Looks good, thanks for taking the feedback on board. I'd suggest squashing the commit introducing the JS tests for the accessible autocomplete into the commit where we add the custom functionality.

It's a shame we have to resort to timeouts, it could introduce some flakiness in tests, but let's see how things go on CI. Having that documented as it is in the commit message should clarify things to future readers.

PR #2735, opened 2024-02-22
Remove bootstrap:all rake task

floehopper:
This doesn't appear to be used any more and so it and its associated code and configuration can be removed.


sengi (approved):
Thanks!

PR #2737, opened 2024-02-22
Fix NoMethodError in SessionsController#create

floehopper:
Trello: https://trello.com/c/qo01Wq1w

We were seeing [the following exception][1] when there was no `user` param in the request:

    NoMethodError: undefined method `[]' for nil:NilClass (NoMethodError)

I've added a test to demonstrate the problem and fixed it by using [`Devise::SessionsController#sign_in_params`][2].

Since `SessionsController` inherits from `Devise::SessionsController`, we probably should have been using this method anyway to properly sanitize the `params`. Using that method also has the benefit of fixing the problem, because it will always return an empty `Hash` rather than `nil`.

I've also taken the opportunity to replace a comment in the implementation with an explanatory test.

As noted in the Trello card, I think we could make further improvements in this controller by making use of [`Warden::Manager` hooks][3] instead of overriding `Devise::SessionsController` action c.f. [these changes][4]. However, I'm short of time, so I'm regarding that work as out-of-scope.

[1]: https://govuk.sentry.io/issues/4838149311/?project=202251
[2]: https://www.rubydoc.info/github/plataformatec/devise/Devise%2FSessionsController:sign_in_params
[3]: https://www.rubydoc.info/github/hassox/warden/Warden/Hooks
[4]: https://github.com/alphagov/signon/pull/2623/commits/9114bdb6971e573e763b9f36d1d02a8398731426#diff-441d219db83f427847d3ecd7a9beb578d99794170944de35500114746651c606

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -28,7 +28,7 @@ private
         user_agent_id: user_agent_record_id,
       )
     else
-      email = params[:user][:email].to_s
+      email = sign_in_params[:email].to_s
       user = User.find_by(email:)
       if user
         EventLog.record_event(
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -28,7 +28,6 @@ private
         user_agent_id: user_agent_record_id,
       )
     else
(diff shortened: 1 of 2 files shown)


chrislo (approved):
Looks good to me!

PR #2833, opened 2024-04-24
Bump Ruby to 3.2.3 for ARM builds.

dj-maisy:
## What?
This bumps the version of Ruby to 3.2.3, as we don't have any ARM base/builder images for 3.2.2. This should un-break the build hopefully.


sengi (approved):
Good catch!

PR #2850, opened 2024-05-01
Add button to delete application supported permissions

brucebolt:
It is not currently possible to delete a supported permission through the user interface.

This has led to code being run on a Rails console when permissions need to be deleted, which has led to other issues because the wrong method was invoked (`delete` instead of `destroy`, which didn't delete the associated `UserApplicationPermission` records).

Therefore adding a delete button (including confirmation view) to allow to do be done in the user interface.

| Action | Screenshot |
| -- | -- |
| Index for supported permissions, showing new 'Delete' link | Screenshot showing the index for supported permissions for a fake application.  There is now a red 'Delete' link to the far side of the table entry. |
| Confirmation view for deleting a permission | Screenshot of the confirmation view for deleting a permission; Main body reads: Are you sure you want to delete the "new" permission for "Test Application 1"?  There is a red 'Delete' button and link labelled 'Cancel'. |
| Success flash notice after deleting a permission | Screenshot showing index page with flash notice that reads Successfully deleted permission new |

[Trello card](https://trello.com/c/95Xkrscy)

### DIFF
--- a/app/views/supported_permissions/confirm_destroy.erb
+++ b/app/views/supported_permissions/confirm_destroy.erb
@@ -1,0 +1,42 @@
+<% content_for :title, "Delete #{@supported_permission.name} permission for #{@application.name}" %>
+
+<% content_for :breadcrumbs,
+   render("govuk_publishing_components/components/breadcrumbs", {
+     collapse_on_mobile: true,
+     breadcrumbs: [
+       {
+         title: "Dashboard",
+         url: root_path,
+       },
+       {
+         title: "Applications",
+         url: doorkeeper_applications_path,
+       },
+       {
+         title: @application.name,
+         url: edit_doorkeeper_application_path(@application),
+       },
+       {
+         title: "Permissions",
+         url: doorkeeper_application_supported_permissions_path(@application),
+       }
+     ]
+   })
+%>
+
+<div class="govuk-grid-row">
+  <section class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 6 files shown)


app/views/supported_permissions/index.html.erb
```diff
@@ -61,7 +61,7 @@
         text: supported_permission.default? ? 'Yes' : 'No',
       },
       {
-        text: link_to('Edit', edit_doorkeeper_application_supported_permission_path(@application, supported_permission)),
+        text: link_to('Edit', edit_doorkeeper_application_supported_permission_path(@application, supported_permission)) + (link_to("Delete", confirm_destroy_doorkeeper_application_supported_permission_path(@application, supported_permission), class: "govuk-link gem-link--destructive govuk-!-margin-left-2")),
```
yndajas, line 64: Not related to this work, but the edit link is missing the `govuk-link` class

yndajas (approved):

app/views/supported_permissions/index.html.erb
brucebolt: Good spot, I'll add that on as another commit.

app/views/supported_permissions/confirm_destroy.erb
```diff
@@ -0,0 +1,42 @@
+<% content_for :title, "Delete #{@supported_permission.name_was} permission for #{@application.name}" %>
```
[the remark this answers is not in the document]
brucebolt: I just copied this from the edit view.  But I don't think it's needed here, so I'll change it to use `name`.

PR #2855, opened 2024-05-02
Make secrets sync zero-configuration.

sengi:
We used to have a lot of cruft lying around in Signon, but now it's much tidier. It's now feasible for us to just import the secrets for all active (non-retired) Doorkeeper applications and all non-suspended API users.

This removes an awkward step in setting up or renaming an application which tends to trip people up.

### DIFF
--- a/lib/tasks/sync_kubernetes_secrets.rake
+++ b/lib/tasks/sync_kubernetes_secrets.rake
@@ -1,13 +1,9 @@
 namespace :kubernetes do
   desc "Synchronise OAuth Token secrets in Kubernetes"
-  task :sync_token_secrets, %i[config_map_name] => :environment do |_, args|
+  task sync_token_secrets: :environment do
     client = Kubernetes::Client.new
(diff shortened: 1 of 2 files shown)


lib/tasks/sync_kubernetes_secrets.rake
```diff
@@ -4,10 +4,7 @@ namespace :kubernetes do
     client = Kubernetes::Client.new
     config_map = client.get_config_map(args[:config_map_name])
```
theseanything, line 5: We don't need to fetch the config map in this case?

theseanything (approved):

lib/tasks/sync_kubernetes_secrets.rake
sengi: Ah yup sorry just realised that too! Fixed now I think.

sengi:
Whoops forgot to update the tests. Hopefully fixed now.

PR #2891, opened 2024-05-16
Only show current permissions form where there are some

yndajas:
It's a little jarring to see an empty current permissions form, so we now only show it where there are some

#### Before

image

#### After

image

### With some existing permissions (no change)

image

---


test/integration/granting_permissions_test.rb
```diff
@@ -125,39 +92,6 @@ class GrantingPermissionsTest < ActionDispatch::IntegrationTest
...
-
-      select "adding"
-      click_button "Add permission"
-      assert_includes @user.permissions_for(app), "pre-existing"
-      assert_includes @user.permissions_for(app), "adding"
-      %w[removing never-1 never-2 never-3 never-4 never-5 never-6].each do |permission|
-        assert_not_includes @user.permissions_for(app), permission
-      end
-    end
-
```
JonathanHallam, line 160: is this test exactly the same as the one you've removed on lines 43 - 75?

JonathanHallam: and indeed 236 - 269?

yndajas: Yup, only differences are:
- the type of admin we're testing with. Given all these test the same actions, the other tests feel sufficient for testing authorisation/privileges. These 9+ permissions tests are more about presentation and user interaction so I feel like they only really need one test per function, rather than one per function per admin type
- the third one seems to grant the admin sign in permissions for the app, though I think that line might not be doing anything useful: `@super_org_admin.grant_application_signin_permission(app)`

test/integration/account_applications_test.rb
```diff
@@ -224,6 +224,72 @@ class AccountApplicationsTest < ActionDispatch::IntegrationTest
       end
     end
 
+    context "when the user already has some permissions" do
+      should "show the new and current permissions forms" do
+        # when I'm signed in as a user who can edit their own permissions
```
JonathanHallam, line 229: I think that these comments would normally be context blocks. That way they show up in the stack trace if they fail. Do you think this can be rejigged a bit to make that work?

yndajas: I might be misunderstanding the suggestion, but breaking this one down into five context blocks (one per comment) would break this test? Throwing all five comments into the context might also make it a little long, even if condensed?

I felt the purpose here was a little different to what we'd typically use context blocks for in any case: it's breaking down a fairly long test by labelling what each chunk/action is doing, rather than saying what the overall context of the test is. I'm typically an anti-commenter, but it felt useful to add in some feature test style explanation here. Open to suggestions though!

JonathanHallam: No, not suggesting a load of inline blocks, more that maybe the comments could be represented by the names of the existing ones. So instead of with comments
```
    context "when the user already has some permissions" do
      should "show the new and current permissions forms" do
```
Something like 
```
context "when a user with more than 8 permissions visits an app with more than 8 supported permissions"
should "show the new and current permissions forms"
```

JonathanHallam: but that may be too wordy

yndajas: I've removed the comments but left the context as is, since the "with apps that have greater than eight permissions" is in an outer context. These match the style of the other tests now

test/integration/granting_permissions_test.rb
```diff
@@ -286,6 +286,69 @@ class GrantingPermissionsTest < ActionDispatch::IntegrationTest
...
+        visit edit_user_path(user)
+        click_on "Manage permissions"
+        click_on "Update permissions for MyApp"
+        assert_current_url edit_user_application_permissions_path(user, app)
+
+        # the new permissions form exists
+        assert_selector "#new_permission_id", visible: false
+
+        # the current permissions form does not exist
+        assert_selector "legend", text: "Current permissions", count: 0
```
JonathanHallam, line 346: could we use `asser_no_selector` here? https://rubydoc.info/gems/capybara/Capybara/Minitest/Assertions#assert_no_selector-instance_method

yndajas: Indeed we can and done. The codebase seems to use a mixture of `count: 0` and `assert_no_selector` but the latter feels cleaner

JonathanHallam (approved):
Nice one, thank you :)

PR #2954, opened 2024-06-13
Hide certain account activity events from non-admins

callumknights:
Response to some user concerns about often automated account activity. Decided to restrict visibility of certain events `Access token generated`, `Access token revoked`, `Access grants deleted`, `Access tokens deleted`.

Visibility of these events was causing confusion for some users, so restricting them to admin/superadmin only visibility seemed sensible.

Trello: https://trello.com/c/Vv4Dn2O7/1193-only-display-technical-events-in-user-activity-log-for-admins-superadmins

### DIFF
--- a/app/models/event_log.rb
+++ b/app/models/event_log.rb
@@ -31,8 +31,8 @@ class EventLog < ApplicationRecord
     TWO_STEP_PROMPT_DEFERRED = LogEntry.new(id: 26, description: "2-step prompt deferred", require_uid: true),
     API_USER_CREATED = LogEntry.new(id: 27, description: "Account created", require_uid: true, require_initiator: true),
     ACCESS_TOKEN_REGENERATED = LogEntry.new(id: 28, description: "Access token re-generated", require_uid: true, require_application: true), # deprecated
-    ACCESS_TOKEN_GENERATED = LogEntry.new(id: 29, description: "Access token generated", require_uid: true, require_application: true, require_initiator: true),
-    ACCESS_TOKEN_REVOKED = LogEntry.new(id: 30, description: "Access token revoked", require_uid: true, require_application: true, require_initiator: true),
+    ACCESS_TOKEN_GENERATED = LogEntry.new(id: 29, description: "Access token generated", require_uid: true, require_application: true, require_initiator: true, access_limited: true),
+    ACCESS_TOKEN_REVOKED = LogEntry.new(id: 30, description: "Access token revoked", require_uid: true, require_application: true, require_initiator: true, access_limited: true),
     PASSWORD_RESET_LOADED_BUT_TOKEN_EXPIRED = LogEntry.new(id: 31, description: "Password reset page loaded but the token has expired", require_uid: true),
     SUCCESSFUL_PASSWORD_RESET = LogEntry.new(id: 32, description: "Password reset successfully", require_uid: true),
     ROLE_CHANGED = LogEntry.new(id: 33, description: "Role changed", require_uid: true, require_initiator: true),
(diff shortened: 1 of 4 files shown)


app/models/event_log.rb
```diff
@@ -76,6 +77,10 @@ def event
     entry.description
   end
 
+  def requires_admin?
+    ACCESS_LIMITED_EVENTS.include? entry
```
Gweaton, line 81: Do you think there's any value in adding a test for this method?

app/views/components/_table.html.erb
```diff
@@ -39,7 +39,7 @@
     <%= t.body do %>
       <% rows.each_with_index do |row, rowindex| %>
         <%= t.row do %>
-          <% row.each_with_index do |cell, cellindex| %>
+          <% row&.each_with_index do |cell, cellindex| %>
```
Gweaton, line 42: I might have missed something, but would `row` ever be `nil` here if we're looping over a thing?

app/views/shared/_event_logs_table.html.erb
```diff
@@ -1,18 +1,18 @@
-<% if logs.any? %>
+<% if @logs.any? %>
```
Gweaton, line 1: Can we use the local variable rather than the instance (`@`) for both `user` and `logs` here if we're passing them in in `app/views/account/activities/show.html.erb` or is there a reason we've gone for the instance var here?

Gweaton (commented):
This is a neat approach, nice work.

I've left a couple of small comments/questions - apologies if I've got the wrong end of the stick on any of them!

I'd suggest adding some of that useful PR description to the commit message here too, as there's some handy detail in there that could be interesting to future readers.

app/views/components/_table.html.erb
callumknights: I actually left this in by accident, before it occurred to me to use `compact` on the loop in `_event_log_table.rb`

app/models/event_log.rb
callumknights: I don't necessarily think this is necessary given it only refers to the content of a constant on line 58

app/views/shared/_event_logs_table.html.erb
callumknights: There we go, sorry, sloppy git fu

app/controllers/account/activities_controller.rb
```diff
@@ -4,6 +4,7 @@ class Account::ActivitiesController < ApplicationController
 
   def show
     @logs = current_user.event_logs.page(params[:page]).per(100)
+    @user = current_user
```
Gweaton, line 7: Now we're calling `current_user` directly in the view, we're not using this instance variable anymore, so it can be removed.

app/views/shared/_event_logs_table.html.erb
```diff
@@ -8,8 +8,8 @@
     ],
     rows: logs.map do |log|
     [ { text: formatted_date(log), format: "event-log-date" },
-      { text: formatted_message(log) } ]
-    end
+      { text: formatted_message(log) } ] unless log.requires_admin? && !user.govuk_admin?
```
Gweaton, line 11: I think we can just call `&& !current_user.govuk_admin?` here and remove the passing of the user local to the views here, actually. I think it'd read a bit more clearly as it's unclear which user this is - is it the user being viewed or the user doing the viewing. I think this is where our confusion came from yesterday.

Gweaton (commented):
One last thing I've noticed that might help readability

app/views/shared/_event_logs_table.html.erb
callumknights: Sorted both, this makes perfect sense. Got a bit lost with the `user` confusion to notice this

Gweaton (approved):
Thanks for making those changes 👍

PR #2988, opened 2024-06-27
Fix applications tables column alignment

yndajas:
[Trello](https://trello.com/c/Vn5vDNV8/1232-fix-applications-tables-column-alignment)

In #2341, the table component from `govuk_publishing_components` was copied over and customised

In #2664, we migrated the applications tables to use this. In the process, we lost the custom classes on the first two columns, which set the widths to 1/4 and 1/3 respectively. This was noted as "acceptable" in the PR body in exchange for improved consistency and other benefits from using the component

However, since we have a custom version of the table component, it's very easy for us to support custom classes. This was in fact done in the same PR that copies the component into the Signon codebase, adding custom class support for the outer table element in 4f09fbc

This adds support for custom classes in the headers, which is then used to restore the old classes and therefore consistent column widths between tables (irrespective of content width)

In #2341, it was suggested that we might try to get the customisations merged upstream - it might be worth exploring that with both this and earlier changes as a separate piece of work

**Note:** this makes existing row alignment issues more apparent in some circumstances. These are fixed separately in #2984 

### Before

image

### After

image

---

### DIFF
--- a/app/helpers/components/table_helper.rb
+++ b/app/helpers/components/table_helper.rb
@@ -50,6 +50,7 @@ module Components
 
       def header(str, opt = {})
         classes = %w[govuk-table__header]
+        classes << opt[:classes] if opt[:classes]
         classes << "govuk-table__header--#{opt[:format]}" if opt[:format]
         classes << "govuk-table__header--active" if opt[:sort_direction]
         link_classes = %w[app-table__sort-link]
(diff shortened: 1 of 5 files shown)


callumknights (approved):
Thanks for the note about the row alignment issues 👍

PR #2989, opened 2024-06-27
Make applications table vertical on small screens

yndajas:
[Trello](https://trello.com/c/9pwUMehD/1233-use-vertical-layout-for-applications-table-on-small-screens)

Since https://github.com/alphagov/signon/pull/2341, we've had support for making tables use a vertical layout on smaller viewports, similar to how the summary list reflows on smaller viewports (the end result is something of a hybrid of table and summary list layouts)

We already use this on the users screen. This feels like something that would be useful on the applications screen too

#### Before

image

#### After

image

#### Before

image

#### After

image

---

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -8,6 +8,8 @@ $govuk-page-width: 1140px;
 @import "components/contact-details";
 @import "components/table";
 
+$applications-table-actions-gap: 10px;
+
 // TODO: move into component
 .gem-c-success-alert,
 .gem-c-error-alert {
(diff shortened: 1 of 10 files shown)


callumknights (commented):
Absolutely get the need to change this, as the comment is wrong. But it's also not super relevant to the PR. At the same time it is such a minor fix it doesn't seem worth it's own, I'm uncertain...

test/integration/granting_permissions_test.rb
```diff
@@ -340,10 +340,10 @@ class GrantingPermissionsTest < ActionDispatch::IntegrationTest
         click_on "Update permissions for MyApp"
         assert_current_url edit_user_application_permissions_path(user, app)
 
-        # the new permissions form exists
+        # the new permissions does not exist
```
callumknights, line 343: Absolutely get the need to change this, as the comment is wrong. But it's also not super relevant to the PR. At the same time it is such a minor fix it doesn't seem worth it's own, I'm uncertain...

callumknights: (sorry for the double comment)

yndajas: Yeah, I'm not otherwise changing this file in this PR, but given it's just two line fix of comments, I'm not sure it warrants its own PR. Happy to split it out though if preferred!

yndajas: I've separated that out here: https://github.com/alphagov/signon/pull/2991

callumknights (approved):

PR #3019, opened 2024-07-11
Show view permissions link even when updatable

yndajas:
[Trello](https://trello.com/c/VNAR8G4o/1250-show-view-permissions-link-even-when-updatable)

Previously on the applications pages the view permissions link wouldn't show if the user could edit permissions. However, we'll soon fix delegatable permissions, which will mean that publishing managers will only see delegatable permissions on the update permissions page. It would be useful for them to be able to see all permissions, even if they can't grant non-delegatable ones, so we should show the view link regardless of update authorisation

### Before

image

### After

image

---

### DIFF
--- a/app/helpers/application_table_helper.rb
+++ b/app/helpers/application_table_helper.rb
@@ -37,22 +37,23 @@ module ApplicationTableHelper
     end
   end
 
-  def account_applications_permissions_link(application)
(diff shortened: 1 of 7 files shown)


yndajas (commented):
The account and users applications controller tests should be updated here

PR #3022, opened 2024-07-15
Upgrade to govuk_publishing_components version that uses GOV.UK Frontend V5.1

mike21573:
I've followed the instructions for [Upgrading to GOV.UK Frontend v5 in the govuk_publishing_components gem](https://docs.google.com/document/d/1uwip7pzQwM7t5ghn9_8KrsWXLePSRUltFPZ5fYw6Ab8/edit). And, as predicted, this app didn't require _all that many_ changes

https://trello.com/c/aZ4vNT80/1240-upgrade-govukpublishingcomponents-for-signon

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,5 +1,15 @@
 //= require govuk_publishing_components/dependencies
-//= require govuk_publishing_components/all_components
+//= require govuk_publishing_components/components/button
+//= require govuk_publishing_components/components/checkboxes
+//= require govuk_publishing_components/components/copy-to-clipboard
+//= require govuk_publishing_components/components/details
+//= require govuk_publishing_components/components/error-summary
+//= require govuk_publishing_components/components/layout-header
+//= require govuk_publishing_components/components/option-select
+//= require govuk_publishing_components/components/skip-link
+//= require govuk_publishing_components/components/table
+//= require govuk_publishing_components/components/tabs
+
 //= require_tree ./modules
 //= require_tree ./components
 //= require rails-ujs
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/application.js
```diff
@@ -1,5 +1,10 @@
 //= require govuk_publishing_components/dependencies
-//= require govuk_publishing_components/all_components
+//= require govuk_publishing_components/components/checkboxes
```
mike21573, line 2: according to [a conversion on Transition's gem upgrade pull request](https://github.com/alphagov/transition/pull/1647#discussion_r1681137070), `checkboxes` needs moving into the ES6-specific file, too

mike21573: OK, that's sorted now

app/views/layouts/application.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= render "layouts/google_tag_manager" %>
 
-<%= render 'govuk_publishing_components/components/layout_for_admin',
+<%= render "govuk_publishing_components/components/layout_for_admin",
```
yndajas, line 3: Is this a requirement for the upgrade?

yndajas (approved):

app/views/layouts/application.html.erb
mike21573: Not at all! It is introduced in its very own commit, though, if that makes it any more palatable?

yndajas: Yeah I think it's fine - it might just be good to mention that this is just to follow convention (seeing as Rubocop presumably hasn't been shouting at us)?

brucebolt: As an aside, Whitehall now uses `erb_lint` to lint erb files.  Maybe we should do the same in all our apps too?

PR #3030, opened 2024-07-19
Fix organisation column in permissions Rake task

yndajas:
[Trello](https://trello.com/c/jPtUcWij/1256-check-whether-any-user-in-departments-check-whether-any-users-in-non-gds-departments-have-recently-used-the-delegatable-permissi)

The output of this Rake task is a CSV-formatted string. Two of the columns contain organisation names, which themselves occasionally include commas. This wraps the organisation name columns in quote marks to prevent each of these being read as multiple columns

---

### DIFF
--- a/lib/tasks/event_log.rake
+++ b/lib/tasks/event_log.rake
@@ -36,11 +36,11 @@ namespace :event_log do
         event_log.trailing_message[1..-2].gsub(',', ';'),
         event_log.initiator_id,
         event_log.initiator.email,
-        event_log.initiator.organisation_name,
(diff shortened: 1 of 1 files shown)


mike21573 (approved):
Nice one 😎 

FWIW, I have a feeling that we could've alternatively just appended `.inspect` to those method calls, but I can't think of any reason off the top of my head to favour that approach

yndajas:
> FWIW, I have a feeling that we could've alternatively just appended `.inspect` to those method calls, but I can't think of any reason off the top of my head to favour that approach

Nice! Didn't know that that wrapped strings in quotes

PR #3080, opened 2024-08-09
Add admin UI to view application access events and monthly stats

richardTowers:
Occasionally, we get asked questions like "how many people actually use this Whitehall?" or "how many people actually use Content Data?".

This feels like information that should be at anyone working on Publishing's fingertips, but it isn't. We can answer things like "how many people have permission to access Content Data?" easily enough, by filtering the Users table. But people who've _actually accessed it_ is a bit of a mystery.

Currently the best bet is to ask a developer to run a SQL query for you, and hope that they don't make one of the many easy-to-make mistakes which would result in you getting the wrong answer.

[In November 2023](https://github.com/alphagov/signon/pull/2471) we added successful authorization events to the event log (which is retained for two years). That means it's reasonably easy to show exactly who's accessed each application, as well as counting the events by month, so long as we don't want to go back by more than 2 years.

This PR adds two views of this data, available only to superadmin users. These are tucked a away in a superadmin-only bit of the interface:

screenshot of the Edit Whitehall page showing two new links - View access log and View monthly access stats

The first - View access log - shows a paginated list of everyone who's accessed a particular app in the last two years, sorted descending on date. Because Smokey has a habit of accessing every app dozens of times a day, I've added a checkbox to explicitly include it and excluded it by default. I've also added a `month` parameter, which can be used to filter the list down to a specific month (pagination by limit / offset is a bit of a clumsy way to find the data you care about otherwise).

screenshot of the Whithall access log, showing 8768 events, such as Rosa signing in to Whitehall

The second - View monthly access stats - shows the count of these events by month. Since we often care about the number of unique users accessing an app, it shows the count distinct by user id as well as the total number of times the application has been accessed each month.

The total monthly access stats will be a bit lower than the number of times users click the app, because the apps have a 20 hour cache, so they won't trigger the event we're measuring more than once per user per day(ish). I think this is probably good enough / not massively confusing though.

Again, I've included a checkbox to include Smokey, as including it by default makes the Total authorization count numbers go crazy.

screenshot of the Whithall monthly event counts

I've only written some pretty light integration tests for these two pages. I'm interested in views on whether it's worth writing some more thorough tests (e.g. my tests don't cover the Include Smokey Users checkbox, or the Month filter, or much in the way of complex data).

Oh, also I had to add a couple database indexes to the event_logs table, because it was impossible to query for this data. See the individual commit for more detail on why they're needed / how well they perform.

### DIFF
--- a/app/controllers/doorkeeper_applications_controller.rb
+++ b/app/controllers/doorkeeper_applications_controller.rb
@@ -25,12 +25,17 @@ class DoorkeeperApplicationsController < ApplicationController
   end
 
   def access_logs
-    smokey_uids = User.where("name LIKE 'Smokey%'").pluck(:uid)
-    @logs = @application.event_logs
+    relation = @application.event_logs
       .includes(:user)
       .where(event_id: 47)
-      .where.not(uid: smokey_uids)
       .order(created_at: :desc)
+
+    unless params[:include_smokey_users] == "true"
+      smokey_uids = User.where("name LIKE 'Smokey%'").pluck(:uid)
+      relation = relation.where.not(uid: smokey_uids)
+    end
+
+    @logs = relation
       .page(params[:page])
       .per(100)
   end
@@ -55,6 +60,7 @@ private
       :home_uri,
       :supports_push_updates,
       :api_only,
+      :include_smokey_users,
     )
   end
 end
--- a/app/controllers/doorkeeper_applications_controller.rb
+++ b/app/controllers/doorkeeper_applications_controller.rb
@@ -24,6 +24,17 @@ class DoorkeeperApplicationsController < ApplicationController
     @users = query.page(params[:page]).per(100)
   end
 
+  def access_logs
+    smokey_uids = User.where("name LIKE 'Smokey%'").pluck(:uid)
+    @logs = @application.event_logs
+      .includes(:user)
+      .where(event_id: 47)
+      .where.not(uid: smokey_uids)
+      .order(created_at: :desc)
+      .page(params[:page])
+      .per(100)
+  end
+
 private
 
   def load_and_authorize_application
--- a/app/controllers/doorkeeper_applications_controller.rb
+++ b/app/controllers/doorkeeper_applications_controller.rb
@@ -27,7 +27,7 @@ class DoorkeeperApplicationsController < ApplicationController
   def access_logs
     relation = @application.event_logs
       .includes(:user)
-      .where(event_id: 47)
+      .where(event_id: EventLog::SUCCESSFUL_USER_APPLICATION_AUTHORIZATION.id)
       .order(created_at: :desc)
 
     unless params[:include_smokey_users] == "true"
@@ -46,7 +46,7 @@ class DoorkeeperApplicationsController < ApplicationController
 
   def monthly_access_stats
     relation = @application.event_logs
-                .where(event_id: 47)
+                .where(event_id: EventLog::SUCCESSFUL_USER_APPLICATION_AUTHORIZATION.id)
                 .group("DATE_FORMAT(created_at, '%Y-%m')")
                 .order(Arel.sql("DATE_FORMAT(created_at, '%Y-%m') DESC"))
 
--- a/app/controllers/doorkeeper_applications_controller.rb
+++ b/app/controllers/doorkeeper_applications_controller.rb
@@ -44,6 +44,25 @@ class DoorkeeperApplicationsController < ApplicationController
       .per(100)
   end
 
+  def monthly_access_stats
+    relation = @application.event_logs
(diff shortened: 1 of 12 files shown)


app/controllers/doorkeeper_applications_controller.rb
```diff
@@ -21,7 +21,18 @@ def update
 
   def users_with_access
     query = policy_scope(User).with_access_to_application(@application)
-    @users = query.page(params[:page]).per(100)
+    @users = query.page(params[:page]).per(10)
```
Gweaton, line 24: Has this slipped in from another commit, or is there a reason we're cutting down on the number of users here? Might be worth updating the commit message if so!

app/controllers/doorkeeper_applications_controller.rb
```diff
@@ -21,7 +21,18 @@ def update
...
+  end
+
+  def access_logs
+    smokey_uids = User.where("name LIKE 'Smokey%'").pluck(:uid)
+    @logs = @application.event_logs
+      .includes(:user)
+      .where(event_id: 47)
```
Gweaton, line 30: Is `47` the successful user application event? Do you think it's worth storing this under a descriptive variable name or some kind of enum to make this as obvious as possible?

app/views/doorkeeper_applications/access_logs.html.erb
```diff
@@ -36,6 +20,32 @@
...
+    {
+      label: "Include Smokey Users",
+      value: "true",
+      checked: params["include_smokey_users"] == "true"
+    }
+  ]
+} %>
+<%= render "govuk_publishing_components/components/input", {
+  label: {
+    text: "Month"
```
Gweaton, line 36: Is "month" clear enough here for users, or do we want to bake "Year" into the text at all if we're requiring that too? This could apply to the params/name etc too but I think that's less important.

app/views/doorkeeper_applications/edit.html.erb
```diff
@@ -23,6 +23,14 @@
   doorkeeper_application_supported_permissions_path(@application),
   class: "govuk-link" %>
 
+<%= link_to "View access log",
```
Gweaton, line 26: Not a big deal but this could probably have come in the last commit that added the access log page.

app/views/doorkeeper_applications/access_logs.html.erb
```diff
@@ -0,0 +1,42 @@
...
+                   },
+                   {
+                     title: @application.name,
+                     url: edit_doorkeeper_application_path(@application),
+                   }
+                 ]
+               })
+%>
+
+<% if @logs.any? %>
```
Gweaton, line 49: It might be useful to provide a bit of an explanation of this data for users. Particularly the fact that this data only goes back 2 years would be a useful caveat to be aware of.

Gweaton (approved):
This looks good to me, nice work. Thanks for documenting that first commit with the gains from an index, I found it a handy explainer.

I've left a couple of comments and suggestions, nothing blocking. I would definitely be in favour of adding some integration tests for the month filter, the Smokey inclusion and potentially what might be shown if there are no results for a month.

One other thing would be validations: are there any we want to add here e.g. searching in future, too far in the past? Again, I don't think these have to block things but would be nice to haves.

app/controllers/doorkeeper_applications_controller.rb
richardTowers: Ah yeah, it's there as `EventLog::SUCCESSFUL_USER_APPLICATION_AUTHORIZATION.id` I just couldn't get it to work originally because my ruby sux. (It's [defined here](https://github.com/alphagov/signon/blob/af0bd6f3bb545ba7c29ca363c8f112d92c1f9570/app/models/event_log.rb#L6-L53), and I couldn't work out the syntax - thought it would be `EventLog::EVENTS::SUCCESSFUL_USER_APPLICATION_AUTHORIZATION` but of course it isn't).

app/views/doorkeeper_applications/access_logs.html.erb
richardTowers: Yeah - "Year and month" is better I think. The hint text has the format, so hopefully together that will be clear enough.

app/views/doorkeeper_applications/edit.html.erb
richardTowers: Yep, I'll rebase that in.

app/views/doorkeeper_applications/access_logs.html.erb
richardTowers: Added in 168dd3f

Gweaton (approved):
This looks good to me, thanks for the tweaks.

PR #3107, opened 2024-08-16
Refactor API Users Permissions Controller test

Gweaton:
The sequel to https://github.com/alphagov/signon/pull/3079, largely following the approach taken in refactoring the controller tests for the Account and User Permissions Controllers to bring them all in line with each others' style.

### DIFF
--- a/test/controllers/api_users/permissions_controller_test.rb
+++ b/test/controllers/api_users/permissions_controller_test.rb
@@ -194,24 +194,42 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
       assert_same_elements [other_permission, not_from_ui_permission], api_user.supported_permissions
     end
 
-    should "prevent permissions being added for other apps" do
-      other_application = create(:application, with_non_delegatable_supported_permissions: %w[other])
-      application = create(:application)
-      api_user = create(:api_user)
-      create(:access_token, application:, resource_owner_id: api_user.id)
+    should "when updating permissions for app A, prevent additionally adding or removing permissions for app B" do
+      application_a = create(:application, with_non_delegatable_supported_permissions: %w[other])
+      application_a_old_permission = create(:supported_permission, application: application_a)
+      application_a_new_permission = create(:supported_permission, application: application_a)
+
+      application_b = create(:application)
+      application_b_old_permission = create(:supported_permission, application: application_b)
+      application_b_new_permission = create(:supported_permission, application: application_b)
+
+      api_user = create(:api_user,
+                        with_signin_permissions_for: [application_a, application_b],
+                        with_permissions: {
+                          application_a => [application_a_old_permission.name],
+                          application_b => [application_b_old_permission.name],
+                        })
+
+      create(:access_token, application: application_a, resource_owner_id: api_user.id)
 
       current_user = create(:superadmin_user)
       sign_in current_user
 
       stub_policy current_user, api_user, update?: true
 
-      other_permission = other_application.supported_permissions.find_by(name: "other")
-
-      patch :update, params: { api_user_id: api_user, application_id: application, application: { supported_permission_ids: [other_permission.id] } }
+      patch :update, params: { api_user_id: api_user, application_id: application_a, application: { supported_permission_ids: [application_a_new_permission.id, application_b_new_permission.id] } }
 
       api_user.reload
 
-      assert_equal [], api_user.supported_permissions
+      assert_same_elements [
+        application_a_new_permission,
+        application_b_old_permission,
+        application_a.signin_permission,
+        application_b.signin_permission,
+      ], api_user.supported_permissions
+
+      assert_not_includes current_user.supported_permissions, application_a_old_permission
+      assert_not_includes current_user.supported_permissions, application_b_new_permission
     end
 
     should "prevent permissions being added that are not grantable from the ui" do
--- a/test/controllers/api_users/permissions_controller_test.rb
+++ b/test/controllers/api_users/permissions_controller_test.rb
@@ -11,7 +11,7 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
       assert_not_authenticated
     end
 
-    should "prevent unauthorized users" do
+    should "prevent unauthorised users" do
       application = create(:application)
       api_user = create(:api_user)
       create(:access_token, application:, resource_owner_id: api_user.id)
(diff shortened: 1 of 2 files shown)


test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -123,21 +123,25 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
-    should "redirect once the permissions have been updated" do
-      application = create(:application, with_non_delegatable_supported_permissions: %w[new old])
-      api_user = create(:api_user, with_permissions: { application => %w[old] })
+    should "update non-signin permissions, retaining the signin permission, then redirect to the API applications path" do
+      application = create(:application)
+      old_permission = create(:supported_permission, application:)
+      new_permission = create(:supported_permission, application:)
+
+      api_user = create(:api_user,
+                        with_signin_permissions_for: [application],
```
yndajas, line 226: Interesting that the grantee didn't previously have the signin permission in this test. Perhaps you can technically update users' non-signin permissions without them having access to the app

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -147,22 +147,26 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
 
-      app2_permission = application2.supported_permissions.find_by!(name: "app2-permission")
+      forbidden_application_permission = forbidden_application.supported_permissions.find_by!(name: "forbidden-permission")
 
-      patch :update, params: { api_user_id: api_user, application_id: application1, application: { supported_permission_ids: [app2_permission.id] } }
+      patch :update, params: {
+        api_user_id: api_user,
+        application_id: allowed_application,
+        application: { supported_permission_ids: [forbidden_application_permission.id] },
+      }
```
yndajas, line 263: So this still appears to be testing two things by mistake:
1. updating permissions for an app other than the one specified by `params[:application_id]`
2. updating permissions for an app the granter (current user) doesn't have access to

When you change the `application_id` to `forbidden_application`, the permissions are added without issue (the `ApiUsersPolicy` only cares that the user is a superadmin, and the `SupportedPermissionsPolicy` allows all permissions to be updated by admins and superadmins), so the fact that the current user/granter doesn't have access to the app (2) is irrelevant

(1) is already covered by another test: "prevent permissions being added for other apps" (updated and renamed in the next commit)

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -194,24 +194,42 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
       assert_same_elements [other_permission, not_from_ui_permission], api_user.supported_permissions
     end
 
-    should "prevent permissions being added for other apps" do
```
yndajas, line 208: Typo in the commit message (copied over from my own typo!): `included` -> `includes`

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -194,24 +194,42 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
-      other_application = create(:application, with_non_delegatable_supported_permissions: %w[other])
-      application = create(:application)
-      api_user = create(:api_user)
-      create(:access_token, application:, resource_owner_id: api_user.id)
+    should "when updating permissions for app A, prevent additionally adding or removing permissions for app B" do
+      application_a = create(:application, with_non_delegatable_supported_permissions: %w[other])
```
yndajas, line 269: The `with_non_delegatable_supported_permissions` can be removed here

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -194,24 +194,42 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
+      application_b_new_permission = create(:supported_permission, application: application_b)
+
+      api_user = create(:api_user,
+                        with_signin_permissions_for: [application_a, application_b],
+                        with_permissions: {
+                          application_a => [application_a_old_permission.name],
+                          application_b => [application_b_old_permission.name],
+                        })
+
+      create(:access_token, application: application_a, resource_owner_id: api_user.id)
```
yndajas, line 284: I was going to ask if we need an access token for `application_b`, but then thought that it's only used when setting the application from `params[:application_id]` in the controller. But I can't see that the resource owner matters in that context, so it wouldn't matter whether the actual API user has an access token. Is there somewhere else that the resource owner becomes relevant in this controller/its dependencies?

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -26,6 +26,69 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
       assert_not_authorised
     end
 
+    should "raise an exception if the user doesn't have access to the application" do
```
yndajas, line 29: I wonder if it's worth grouping this and the access token test into a context, since they could both be considered "user doesn't have access" tests, and then making sure in this test that they do have an access token, just not the signin permission

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -26,6 +26,69 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
+      sign_in create(:admin_user)
```
yndajas, line 30: Only superadmins can access these routes, so we'll need to use a superadmin here

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -26,6 +26,69 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
+    end
+
+    should "raise an exception if the application is retired" do
+      sign_in create(:superadmin_user)
+
+      application = create(:application, retired: true)
+      permission = create(:supported_permission, application:)
+      api_user = create(:api_user,
+                        with_signin_permissions_for: [application],
+                        with_permissions: { application => [permission.name] })
```
yndajas, line 47: Do we need the non-signin permissions here?

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -26,6 +26,69 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
+      application = create(:application)
+      api_user = create(:api_user)
+
+      current_user = create(:superadmin_user)
+      sign_in current_user
+
+      assert_raises(ActiveRecord::RecordNotFound) do
+        get :edit, params: { api_user_id: api_user, application_id: application }
+      end
+    end
```
yndajas: In this test and the following one, should the user have the `signin` permission, to make sure we're just testing the missing access token, especially since they raise the same exception?

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -26,6 +26,69 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
+
+    should "include applications with revoked access tokens when there is at least one non-revoked access token" do
+      sign_in create(:superadmin_user)
+
+      application = create(:application)
+      api_user = create(:api_user)
```
yndajas: So this works even without the API user having the signin permission? Is the signin permission relevant at all?

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -63,64 +126,95 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
+      create(:access_token, application:, resource_owner_id: api_user.id)
 
-      assert_response :success
+      assert_raises(ActiveRecord::RecordNotFound) do
+        patch :update, params: { api_user_id: api_user, application_id: application, application: { supported_permission_ids: [application.id] } }
+      end
     end
 
-    should "raise an exception if the user does not have an access token for the application" do
+    should "raise an exception if the user cannot be found" do
```
yndajas, line 182: I think this test can go? It's not called out in your list in the commit message, we don't otherwise test that non-existent params raise a `RecordNotFound` error (it's sort of testing the default Rails behaviour?), and it's not tested for the other actions. I think I removed this in the equivalent commit on the other PR

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -63,64 +126,95 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
+      create(:access_token, resource_owner_id: api_user.id, application:, revoked_at: Time.current)
 
-      patch :update, params: { api_user_id: api_user, application_id: application }
+      assert_raises(ActiveRecord::RecordNotFound) do
+        patch :update, params: { api_user_id: api_user, application_id: application, application: { supported_permission_ids: [] } }
+      end
+    end
 
-      assert_not_authenticated
+    should "include applications with revoked access tokens when there is at least one non-revoked access token" do
```
yndajas, line 205: `include` -> `allow updating`?

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -63,64 +126,95 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
...
+      sign_in create(:superadmin_user)
+
+      application = create(:application, with_non_delegatable_supported_permissions: %w[permission])
+      api_user = create(:api_user)
+      create(:access_token, resource_owner_id: api_user.id, application:, revoked_at: Time.current)
+      create(:access_token, resource_owner_id: api_user.id, application:)
+
+      permission = application.supported_permissions.find_by(name: "permission")
```
yndajas, line 213: What do you think about moving this (and other tests) to the other style where we assign the permission to a variable on creation, rather than using the transient then finding?

yndajas (commented):
Generally looks good, but a few comments on some details, and some questions we might want to answer about how one or two things work, like whether the signin permission actually matters

test/controllers/api_users/permissions_controller_test.rb
yndajas: I.e. this change: https://github.com/alphagov/signon/pull/3079/commits/21e7a0ee76dbeec4f16a32f31c9c6d073e4a91ec

test/factories/users.rb
```diff
@@ -135,14 +126,13 @@
...
-                  app_or_name
-                end
-          user.grant_application_permissions(app, permission_names)
+          user.grant_application_permissions(find_application(app_or_name), permission_names)
         end
       end
+
+      evaluator.with_signin_permissions_for.each do |app_or_name|
+        user.grant_application_signin_permission(find_application(app_or_name))
+      end
```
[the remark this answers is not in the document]
yndajas: I.e. this change: https://github.com/alphagov/signon/pull/3079/commits/44ac08450ee85abac43ae9229ca5caf684afe039

test/controllers/api_users/permissions_controller_test.rb
Gweaton: We can view the edit page without the permissions, so we can remove the permissions here.

This can be a copy of the access token test below (with retired and access_token) in it

Gweaton:
Closing this in favour of #3110

test/controllers/api_users/permissions_controller_test.rb
Gweaton: We check that the user has a token that matches the params of the `application_id` we're trying to update [here](https://github.com/alphagov/signon/blob/main/app/controllers/api_users/permissions_controller.rb#L37) - `@api_user` there is the resource owner.

PR #3110, opened 2024-08-20
Refactor API Users Permissions Controller #edit tests

Gweaton:
[Trello card](https://trello.com/c/mppostUR/1289-refactor-apiuserspermissionscontrollertest)

# Changes in this PR

Refactors the API Users Permissions Controller tests for the `#edit` action to make things a bit easier to read and ensure we're testing the right things to prevent false positives.

This is a rework of #3107, in which I felt I tried to change too much in one go, was difficult to review and modify. In splitting out PRs into smaller chunks, I'm hoping they'll be easier to review and unblock possible parallel work.

Up next, we'll be refactoring tests for the `#update` action.

## Caveat

We're a little unsure about the need for `signin` permissions with API users at the moment, as tests appear to be passing for the right reason without us assigning them to test users. We'll be investigating them shortly, but to keep changes small and incremental, I've not touched that functionality in these tests yet - we should tidy that up in future work.

### DIFF
--- a/test/controllers/api_users/permissions_controller_test.rb
+++ b/test/controllers/api_users/permissions_controller_test.rb
@@ -49,7 +49,7 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
     end
 
     should "include a hidden field for the signin permission so that it is not removed" do
-      application = create(:application, with_non_delegatable_supported_permissions: ["perm-1", SupportedPermission::SIGNIN_NAME])
+      application = create(:application, with_non_delegatable_supported_permissions: %w[perm-1])
       api_user = create(:api_user, with_permissions: { application => %w[perm-1] })
       create(:access_token, application:, resource_owner_id: api_user.id)
 
--- a/test/controllers/api_users/permissions_controller_test.rb
+++ b/test/controllers/api_users/permissions_controller_test.rb
@@ -95,6 +95,7 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
 
         application = create(:application, retired: true)
         api_user = create(:api_user)
+        create(:access_token, resource_owner_id: api_user.id, application:)
 
         assert_raises(ActiveRecord::RecordNotFound) do
           get :edit, params: { api_user_id: api_user, application_id: application }
--- a/test/controllers/api_users/permissions_controller_test.rb
+++ b/test/controllers/api_users/permissions_controller_test.rb
@@ -2,99 +2,103 @@ require "test_helper"
 
 class ApiUsers::PermissionsControllerTest < ActionController::TestCase
   context "#edit" do
-    should "prevent unauthenticated users" do
-      application = create(:application)
-      api_user = create(:api_user)
+    context "when the granting user is authorised to edit another's permissions" do
+      should "render a page with checkboxes for the grantable permissions and a hidden field for the signin permission so that it is not removed" do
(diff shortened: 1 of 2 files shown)


test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -49,7 +49,7 @@ class ApiUsers::PermissionsControllerTest < ActionController::TestCase
     end
 
     should "include a hidden field for the signin permission so that it is not removed" do
-      application = create(:application, with_non_delegatable_supported_permissions: ["perm-1", SupportedPermission::SIGNIN_NAME])
+      application = create(:application, with_non_delegatable_supported_permissions: %w[perm-1])
```
yndajas: Typo in the commit message: `Signon` -> `signin` 🙃

Also, "A factory `application` is always created with a Signon [sic] permission" -> I think it's more that the factory doesn't override the default non-factory behaviour (`after_create` hook). It's not something specific to factory-made applications. You kinda say this at the end though!

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -2,99 +2,103 @@
 
 class ApiUsers::PermissionsControllerTest < ActionController::TestCase
   context "#edit" do
-    should "prevent unauthenticated users" do
-      application = create(:application)
-      api_user = create(:api_user)
+    context "when the granting user has access to edit another's permissions" do
```
yndajas, line 5: `has access` -> `is authorised`?

test/controllers/api_users/permissions_controller_test.rb
```diff
@@ -2,99 +2,103 @@
 
 class ApiUsers::PermissionsControllerTest < ActionController::TestCase
   context "#edit" do
-    should "prevent unauthenticated users" do
```
yndajas, line 5: The diff here ("Group related tests in contexts ") is a bit horrible I think because of the move + context nesting in one go - would it be possible to reorder then nest in two separate commits, at least temporarily?

yndajas, line 5: From the commit: "This doesn't change the implementation of the tests yet" -> are there likely to be more changes to the `#edit` tests later? It looks like these are fairly small things for now (and two of the commits are factory changes)

yndajas (commented):
Generally looks good, but might be good to break up one of the commits, at least temporarily, to make it easier to review

yndajas (approved):
Actually it's easy enough to compare with the split diff view, so I'm happy that everything is ported over! See the earlier review for a couple of minor things

test/controllers/api_users/permissions_controller_test.rb
Gweaton: One day I'll get the difference between Signon and signin correct, but today obviously wasn't that day.

test/controllers/api_users/permissions_controller_test.rb
Gweaton: Yeah I think I thought there'd be more, but there was only one small thing. I've updated the commit message now.

yndajas: I think we could still do the stuff like relying less on names, but that can come later

yndajas: Oh yeah - not sure what I was seeing yesterday!

PR #3171, opened 2024-09-17
Display success banner when adding or removing access to an application

Gweaton:
[Trello card](https://trello.com/c/TrHzYz6u/1352-add-a-success-banner-when-granting-or-removing-app-access)

## Changes in this PR

We previously displayed a success alert banner when one's own or another user's permissions were updated, but didn't show anything when granting access or removing it. We now display the same banner with different messaging when a user's access is granted or removed.

This was a little more involved than I'd expected due to the challenge in differentiating cleanly between a user having all their permissions removed and seeing one message (i.e. they just have the "Signin" permission) and a user being granted access to an app (i.e. also just having the "Signin" permission) and seeing another message in the banner. Because of this, I've had to make use of some additional flash variables and flags to method arguments.

The API Users journey didn't need updating here, as we already show flash messages when an API token is created or revoked, as well as when permissions are updated.

### Granting own access

image

### Removing own access

image

### Granting another user access

image

### Removing another user's access

image

### DIFF
--- a/app/helpers/success_alert_helper.rb
+++ b/app/helpers/success_alert_helper.rb
@@ -1,0 +1,15 @@
+module SuccessAlertHelper
+  def access_and_permissions_granted_params(application_id, granting_access:, user: current_user)
+    if granting_access
+      {
+        message: "Access granted",
+        description: access_granted_message(application_id, user),
+      }
+    else
+      {
+        message: "Permissions updated",
+        description: message_for_success(application_id, user),
+      }
+    end
+  end
+end
--- a/app/helpers/success_alert_helper.rb
+++ b/app/helpers/success_alert_helper.rb
@@ -12,4 +12,11 @@ module SuccessAlertHelper
       }
     end
   end
+
+  def access_removed_params(application_id, user: current_user)
+    {
+      message: "Access removed",
+      description: access_removed_message(application_id, user),
+    }
+  end
 end
--- a/test/helpers/application_access_helper_test.rb
+++ b/test/helpers/application_access_helper_test.rb
@@ -26,4 +26,25 @@ class ApplicationAccessHelperTest < ActionView::TestCase
       end
     end
   end
+
+  context "#access_removed_message" do
+    context "when the user is removing their own access" do
+      should "return a message informing them that they have access to an application" do
+        assert_equal "Your access to Whitehall has been removed.", access_removed_message(@application)
+      end
+    end
+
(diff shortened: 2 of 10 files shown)


app/helpers/success_alert_helper.rb
```diff
@@ -0,0 +1,15 @@
...
+  def access_and_permissions_granted_params(application_id, granting_access:, user: current_user)
+    if granting_access
+      {
+        message: "Access granted",
+        description: access_granted_message(application_id, user),
+      }
+    else
+      {
+        message: "Permissions updated",
+        description: message_for_success(application_id, user),
```
yndajas, line 11: Could we possibly change the name of `message_for_success` to something like `permissions_updated_message` to make this more readable without much context (and more in line with the better name you chose: `access_granted_message`)?

app/helpers/success_alert_helper.rb
```diff
@@ -12,4 +12,11 @@ def access_and_permissions_granted_params(application_id, granting_access:, user
       }
     end
   end
+
+  def access_removed_params(application_id, user: current_user)
+    {
+      message: "Access removed",
+      description: access_removed_message(application_id, user),
+    }
+  end
```
yndajas, line 21: I'm not sure I see the distinction between the two methods in this file. The first one includes `granted` in the name but it would also apply if permissions have been removed (or even some granted, some removed), so I'm not sure the granted/removed distinction holds?

Could the method added in the previous commit maybe be rejigged here to accommodate the three options (e.g. adding `removing_access:` to the method signature), or a clearer distinction made? This would extend the conditional logic in the helper but allow the view to stay 'simple'

test/helpers/application_access_helper_test.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  context "#access_granted_message" do
+    context "when the user is granting themself access" do
+      should "return a message informing them that they have access to an application" do
+        assert_equal "You have been granted access to Whitehall.", access_granted_message(@application)
+      end
+    end
+
+    context "when the user is granting another access" do
+      should "return a message informing them that the other user have access to an application" do
```
yndajas, line 17: `have access` -> `has access`

yndajas (commented):
Looks good. Just a few suggestions, the main one being about the distinction between methods in the success alert helper

Gweaton:
Closing in favour of #3175

PR #3172, opened 2024-09-17
Make organisations field searchable

yndajas:
This upgrades the organisations field in the page for inviting a new user to use the accessible autocomplete module (with progressive enhancement)

### Before

https://github.com/user-attachments/assets/5975651a-5c36-4e8f-a7aa-be507711746b

### After

https://github.com/user-attachments/assets/fbb662b8-d0c9-4beb-ba65-d6f78a21e2fd


test/integration/inviting_users_test.rb
```diff
@@ -384,4 +384,132 @@ class InvitingUsersTest < ActionDispatch::IntegrationTest
...
+      click_button "Create user and send email"
+
+      new_user = User.find_by(email: "h@from.steps", role: Roles::Superadmin.name)
+      assert_not_nil new_user
+      assert_equal new_user.organisation, @organisation
+    end
+
+    should("not show an 'Add' button") { assert_no_selector "button", text: "Add" }
+
+    should "reset the value of the select element when it no longer matches what's shown in the autocomplete input" do
```
Gweaton, line 429: How would you feel about extracting these tests into more "component" style tests? That way we'd only need to test it once as we're reusing the component across multiple pages.

As it stands we're testing this behaviour twice across the two integration tests. We could then restrict the two test files modified in this PR to only test a user inputting some text into the autocomplete and inviting a user.

test/integration/updating_permissions_for_apps_with_many_permissions_test.rb
```diff
@@ -162,14 +152,13 @@ def assert_permissions_unchanged
       should "reset the value of the select element when it no longer matches what's shown in the autocomplete input" do
         assert_select_permission_to_grant_with_javascript
 
-        @autocomplete_input.fill_in with: "addin"
-        autocomplete_option = find(".autocomplete__option")
-
-        assert_equal "addin", @autocomplete_input.value
-        assert_equal @new_permission_to_grant.name, autocomplete_option.text
-        assert_equal "", @select_element.value
+        assert_resets_select_when_desynced_with_autocomplete(
```
Gweaton, line 155: If we are going to go down a more abstracted approach with our tests, could we make method names for our helpers even more human readable? I know we're following the lead taken by Minitest here, but I'm not the biggest fan of the brevity of Minitest's DSL.

Also, forgive my ignorance but I'm not sure what desynced means!

PR #3175, opened 2024-09-18
Display success banner when adding or removing access to an application (take 2)

Gweaton:
[Trello card](https://trello.com/c/TrHzYz6u/1352-add-a-success-banner-when-granting-or-removing-app-access)

## Changes in this PR

This is (I feel) a neater, second attempt at #3171 with less complexity and is more easily extensible.

We previously displayed a success alert banner when one's own or another user's permissions were updated, but didn't show anything when granting access or removing it. We now display the same banner with different messaging when a user's access is granted or removed.

The API Users journey didn't need updating here, as we already show flash messages when an API token is created or revoked, as well as when permissions are updated.

### Granting own access

image

### Removing own access

image

### Granting another user access

image

### Removing another user's access

image

### DIFF
--- a/app/views/account/applications/index.html.erb
+++ b/app/views/account/applications/index.html.erb
@@ -20,7 +20,7 @@
   <% content_for(:custom_alerts) do %>
     <%= render "govuk_publishing_components/components/success_alert", {
         message: "Permissions updated",
-        description: message_for_success(flash[:application_id]),
+        description: permissions_updated_description(flash[:application_id]),
     } %>
   <% end %>
 <% end %>
--- a/app/views/account/applications/index.html.erb
+++ b/app/views/account/applications/index.html.erb
@@ -16,11 +16,11 @@
    })
 %>
 
-<% if flash[:application_id] %>
+<% if flash[:success_alert] %>
   <% content_for(:custom_alerts) do %>
     <%= render "govuk_publishing_components/components/success_alert", {
-        message: "Permissions updated",
-        description: permissions_updated_description(flash[:application_id]),
+        message: flash[:success_alert]["message"],
+        description: sanitize(flash[:success_alert]["description"]),
     } %>
   <% end %>
 <% end %>
--- a/test/controllers/account/permissions_controller_test.rb
+++ b/test/controllers/account/permissions_controller_test.rb
@@ -269,7 +269,7 @@ class Account::PermissionsControllerTest < ActionController::TestCase
         assert_same_elements [application.signin_permission, new_permission], current_user.supported_permissions
       end
 
-      should "assign the application id to the application_id flash" do
+      should "assign the success alert hash to flash" do
         application = create(:application)
         old_permission = create(:supported_permission, application:)
         new_permission = create(:supported_permission, application:)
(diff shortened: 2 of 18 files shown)


app/views/account/applications/index.html.erb
```diff
@@ -16,11 +16,11 @@
...
 %>
 
-<% if flash[:application_id] %>
+<% if flash[:success_alert] %>
   <% content_for(:custom_alerts) do %>
     <%= render "govuk_publishing_components/components/success_alert", {
-        message: "Permissions updated",
-        description: permissions_updated_description(flash[:application_id]),
+        message: flash[:success_alert]["message"],
+        description:flash[:success_alert]["description"].html_safe,
```
yndajas, line 23: Could we call `.html_safe` in the helper or the controller, which would then allow us to simplify this hash to `flash[:success_alert]`, or do you prefer to have all the options written out in the view?

test/controllers/account/permissions_controller_test.rb
```diff
@@ -269,7 +269,7 @@ class Account::PermissionsControllerTest < ActionController::TestCase
         assert_same_elements [application.signin_permission, new_permission], current_user.supported_permissions
       end
 
-      should "assign the application id to the application_id flash" do
+      should "assign the success alert message to flash" do
```
yndajas, line 272: `success alert hash`? Given message is one of the keys, this might be confusing

test/support/updating_permissions_helpers.rb
```diff
@@ -35,6 +35,7 @@ def assert_update_permissions(application, grantee, grant: [], revoke: [])
 
     click_button "Update permissions"
 
+    assert_flash_content("Permissions updated")
     assert_flash_content(grant.map(&:name))
```
yndajas, line 39: These could be combined (`assert_flash_content(["Permissions updated", *grant.map(&:name)])`) so that the flash element is only grabbed once, but the clarity of this is probably more important!

test/support/granting_access_helpers.rb
```diff
@@ -29,6 +29,10 @@ def assert_grant_access(application, grantee, grantee_is_self: false)
 
     assert app_with_access_table.has_content?(application.name)
     assert grantee.has_access_to?(application)
+
+    success_banner_description = grantee_is_self ? "You have been granted access to #{application.name}." : "#{grantee.name} has been granted access to #{application.name}."
+    assert_flash_content("Access granted")
+    assert_flash_content(success_banner_description)
```
yndajas, line 35: `success_alert_description`?

test/helpers/application_access_helper_test.rb
```diff
@@ -26,4 +26,25 @@ class ApplicationAccessHelperTest < ActionView::TestCase
...
+        assert_equal "Gerald's access to Whitehall has been removed.", access_removed_description(@application, user)
+      end
+    end
+
+    context "when the application does not exist" do
+      should "return nil" do
+        assert_nil access_removed_description(:made_up_id)
+      end
+    end
+  end
```
yndajas, line 49: Should the integration tests be updated in this final commit in the same way as in the penultimate commit?

yndajas (approved):
Looks good!

app/views/account/applications/index.html.erb
Gweaton: I'm afraid we can't call `html_safe` in the helper/controller as the escaping is stripped out when it's serialized as JSON when stored as a browser cookie. [1]

This same serialization is what converts symbols to strings, meaning we also can't pass `flash[:success_alert]` to the Success Alert, as the `message` and `description` keys are converted to strings to store in the cookie, while the component expected an object with symbol keys.[2]

[1]: https://groups.google.com/g/rubyonrails-core/c/z52zgDgUmbs
[2]: bigbinary.com/blog/flash-access-changes-rails-4-1

Gweaton: I've also updated this to `sanitize` as [apparently](https://apidock.com/rails/String/html_safe) it's recommended over `html_safe`

test/support/updating_permissions_helpers.rb
Gweaton: As we've discussed over the past couple of weeks, I'm definitely one for more readability in tests!

test/helpers/application_access_helper_test.rb
Gweaton: Oops, must have missed this in my rush to get this finished before going off. Well spotted.

PR #3178, opened 2024-09-19
Simplify organisation selection

yndajas:


app/helpers/role_organisations_helper.rb
```diff
@@ -1,6 +1,6 @@
 module RoleOrganisationsHelper
   def options_for_your_organisation_select(current_user)
-    organisations = Pundit.policy_scope(current_user, Organisation).not_closed.order(:name)
```
Gweaton, line 3: I'm probably overthinking this, but removing this policy makes me _slightly_ uneasy - because this is a helper method (and `users_helper` too), the caller has to know whether or not it's acceptable to call this method and render all organisations, e.g if this were called outside of the context of the protected controller methods, so maybe having that double authentication is safer than not having it.

That said, we can probably revisit this/put it back in if we were to ever allow non `govuk_admins` to select organisations for any reason.

app/helpers/organisation_helper.rb
```diff
@@ -0,0 +1,9 @@
+module OrganisationHelper
+  def options_for_organisation_select(selected_id: nil)
+    [{ text: "None", value: nil }] + Organisation.not_closed.map do |organisation|
```
Gweaton, line 3: Now we're doing this in one place, can we remove the default sorting scope and sort here (and in the index page), or do think it's better to sort at database level? I suppose I'm wondering if it'll slow things down in situations where we don't need to sort but it's probably not a big deal.

Gweaton (approved):
Looks good, just a couple of questions

app/helpers/role_organisations_helper.rb
yndajas: Yeah I was a little uneasy about that too. I've reinstated the scope for now, but also added a note in the commit message about another method elsewhere in the codebase that makes things even more confusing!

app/helpers/organisation_helper.rb
yndajas: We could, but we already order a few other models using a default scope, and I don't think we have a use case for not wanting this anywhere yet. I've removed the `order(:name)` call from a few other places in the latest push, and moved the changes into the commit where the default scope is added to make the benefits/use cases clearer

app/helpers/role_organisations_helper.rb
yndajas: That's this commit message, where I also explained the choice to retain the policy scope: https://github.com/alphagov/signon/pull/3178/commits/8075a729eeb8fe34da5039e55ffa7eb7914d6d51

Gweaton: Nice one, sounds good to me.

PR #3203, opened 2024-10-01
Fix missing assertions warning

yndajas:
We were getting a lot of noise about tests missing assertions because of using custom assertion methods (or just relying on tests not raising errors rather than explicitly asserting anything). This adds `assert` in multiple places to fix this, and in one case adds a default return of `true` in a method we need to assert on

E.g.

```
Test is missing assertions: `test_: setting a 2SV code signed in with an existing 2SV setup should require the code again on next login. ` /govuk/signon/test/integration/two_step_verification_test.rb:57
```

---


Gweaton (approved):
Nice, well done for tidying this up.

PR #3205, opened 2024-10-01
Rename 'delegatable' permissions to 'delegated'

yndajas:
[Trello](https://trello.com/c/XTbuNRCz/1385-revise-delegated-permissions-terminology-in-codebase)

This came out of work to fix 'delegatable' permissions. We've gradually settled on some new terminology, the main change being that we now refer to them as 'delegated' permissions (all permissions are delegatable). This updates the codebase, including the UI and documentation, to reflect that shift

---


app/views/supported_permissions/index.html.erb
```diff
@@ -26,6 +26,20 @@
   } %>
 </div>
 
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/details", {
+      title: "What is a 'delegated' permission?"
+    } do %>
+      <p class="govuk-body">
+        When a permission is delegated, publishing managers are able to grant the
```
Gweaton, line 35: This is a helpful explainer, but I'm not sure I understand the distinction between "delegated" and "delegatable" here. Isn't "delegatable" the right word here for someone being able to delegate a thing? Using the past participle feels like a thing that's already happened, rather than a thing that can happen

yndajas: I can see how this could be confusing, and I think the choice of distinguishing verbs isn't the most natural, but 'delegating' and 'granting' permissions are two separate concepts. We grant permissions to individuals, while we delegate the authority to grant permissions to publishing managers. I wonder if we need to clarify that here (and maybe also in the access and permissions documentation)?

yndajas: tl;dr 'delegated' == 'grantable by publishing managers'

brucebolt (commented):
This all makes sense to me.  As long as the name is consistent throughout, I'm not really too concerned about the exact term.

callumknights:
Makes sense to me as well. It take me reading your other comment explaining this to George for me to figure out exactly what's going on, so it's not super natural, but I do understand and agree with the terminology.

Gweaton (approved):
This is more consistent and seems to be clearer for others, particularly people who haven't worked much on Signon, so let's do it.

PR #3214, opened 2024-10-07
Configure hosts for application

brucebolt:
Note: the healthcheck endpoints are requested by IP, not domain, so we need to specifically exclude them from the protection.

[Trello card](https://trello.com/c/frqFN7D8)


brucebolt:
> Probably worth testing this one on integration before merging.

I had this deployed to integration earlier today.  Confirmed both Signon works and the issue appears to be resolved.

PR #3367, opened 2024-11-27
Show parent organisation in table

brucebolt:
This adds a column to the organisations index table with the name of the parent organisation.

The information is available from a Rails console, but not in the UI, meaning it can be difficult to see the organisation hierarchy without developer input.

Screenshot:
Screenshot of the Signon organisations index table.  There are five columns: name, type, slug, parent name and 2-step verification mandated.

[Trello card](https://trello.com/c/il3oWbfb)


mike3985 (approved):
We should consider making a #show page for organisations one of these days. But this seems like an improvement for now 👍

PR #3458, opened 2024-12-19
(783) Configute Pact tests provider for Signon

pezholio:
Trello card: https://trello.com/c/ifMWZl9n/783-add-api-adapters-for-signon

This adds the Pact tests provider and the workflow, so we can run Pact tests via `gds-api-adapters` (PR here https://github.com/alphagov/gds-api-adapters/pull/1316).

I've also added some lightweight documentation with details about the endpoint and how to authenticate.


JonathanHallam (approved):
Nice, thanks

PR #3529, opened 2025-01-21
Add rake tasks for dealing with deleted organisations

brucebolt:
The `OrganisationFetcher` does not handle the situation of an organisation having been deleted in Whitehall without first being closed.
    
This means a deleted organisation remains marked as open, and can have users assigned to it in Signon.
    
Adding rake tasks to:
1. allow us to manually mark organisations as closed, so the organisation no longer appears as an option to which users can be assigned in the user interface.
2. allow us to move all users from one organisation to another, without needing to add all their email addresses in a CSV file.

[Trello card](https://trello.com/c/WDv2141z)

### DIFF
--- a/lib/tasks/data_hygiene.rake
+++ b/lib/tasks/data_hygiene.rake
@@ -12,4 +12,15 @@ namespace :data_hygiene do
     organisation.update!(closed: true)
     puts "Marked organisation #{organisation.slug} as closed"
   end
+
+  desc "Move all users from one organisation to another"
+  task :bulk_update_user_organisation, %i[old_content_id new_content_id] => :environment do |_, args|
+    old_organisation = Organisation.find_by(content_id: args[:old_content_id])
+    new_organisation = Organisation.find_by(content_id: args[:new_content_id])
+
+    users = User.where(organisation: old_organisation)
+    users.update_all(organisation_id: new_organisation.id)
+
+    puts "Moved #{users.count} users from #{old_organisation.slug} to #{new_organisation.slug}"
+  end
 end
--- a/lib/tasks/data_hygiene.rake
+++ b/lib/tasks/data_hygiene.rake
@@ -5,4 +5,11 @@ namespace :data_hygiene do
       abort "bulk updating organisations encountered errors"
     end
   end
+
+  desc "Mark an organisation as closed"
+  task :close_organisation, %i[content_id] => :environment do |_, args|
+    organisation = Organisation.find_by(content_id: args[:content_id])
+    organisation.update!(closed: true)
+    puts "Marked organisation #{organisation.slug} as closed"
+  end
 end
--- a/test/lib/tasks/data_hygiene_test.rb
+++ b/test/lib/tasks/data_hygiene_test.rb
@@ -16,4 +16,20 @@ class DataHygieneTaskTest < ActiveSupport::TestCase
       assert organisation.reload.closed?
     end
   end
+
+  context "#bulk_update_user_organisation" do
+    should "update the organisation for matching users" do
+      old_organisation = create(:organisation, slug: "department-of-health-old")
+      new_organisation = create(:organisation, slug: "department-of-health-new")
+      another_organisation = create(:organisation, slug: "department-of-other-stuff")
+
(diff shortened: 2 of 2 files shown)


lib/tasks/data_hygiene.rake
```diff
@@ -10,4 +10,12 @@ namespace :data_hygiene do
...
     Organisation.find_by(content_id: args[:content_id]).update!(closed: true)
   end
+
+  desc "Move all users from one organisation to another."
+  task :bulk_update_user_organisation, %i[old_content_id new_content_id] => :environment do |_, args|
+    old_organisation = Organisation.find_by(content_id: args[:old_content_id])
+    new_organisation = Organisation.find_by(content_id: args[:new_content_id])
+
+    User.where(organisation: old_organisation).update_all(organisation_id: new_organisation.id)
+  end
```
yndajas, line 20: **suggestion:** feedback

Might be useful to have a success message that explicitly mentions the direction of travel here in case the user got the old and new content IDs the wrong way around? A success message in the other task might be nice too, but I suppose you can always use the Rails console to check things

lib/tasks/data_hygiene.rake
```diff
@@ -10,4 +10,12 @@ namespace :data_hygiene do
   task :close_organisation, %i[content_id] => :environment do |_, args|
     Organisation.find_by(content_id: args[:content_id]).update!(closed: true)
   end
+
+  desc "Move all users from one organisation to another."
```
yndajas, line 14: **suggestion:** commit message

> There may be circumstances in which we need to move all the users from one department to another.

In the second commit message, it might be useful to note the specific use case that's motivating introducing this now (since it's not a hypothetical; in case we consider removing it in future)

yndajas (approved):

lib/tasks/data_hygiene.rake
```diff
@@ -5,4 +5,9 @@ namespace :data_hygiene do
       abort "bulk updating organisations encountered errors"
     end
   end
+
+  desc "Mark as organisation as closed."
```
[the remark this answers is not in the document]
brucebolt: Have fixed.

test/lib/tasks/data_hygiene_test.rb
```diff
@@ -0,0 +1,17 @@
+require "test_helper"
+
+class DataHygieneTaskTest < ActiveSupport::TestCase
+  setup do
+    Signon::Application.load_tasks if Rake::Task.tasks.empty?
+  end
+
+  context "#close_organisation" do
+    should "marks the organisation as closed" do
```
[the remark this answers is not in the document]
brucebolt: Oops.  Was writing with `it` instead of `should` syntax.

lib/tasks/data_hygiene.rake
brucebolt: I've added an extra sentence to the commit message explaining the motivation for adding this task.

lib/tasks/data_hygiene.rake
```diff
@@ -5,4 +5,20 @@ namespace :data_hygiene do
       abort "bulk updating organisations encountered errors"
     end
   end
+
+  desc "Mark an organisation as closed."
```
yndajas, line 9: **nitpick:** style

Forgot to note one thing. I think the `desc`s in Rake tasks don't typically have full stops. I saw that the other task in this file had one, but I don't think tasks in other files do

lib/tasks/data_hygiene.rake
yndajas: Is there feedback in the first Rake task? I can't see it. Also wonder if the name is enough for the second one if they have the same name (I think that's the case with the integration MHCLGs?)?

brucebolt: Looks like I put it into the test instead of the rake task 🤦‍♂️.

Maybe the slugs would be better to output?

lib/tasks/data_hygiene.rake
brucebolt: I've removed the full stops from the tasks I've added.  Didn't originally include them, but decided to add it for consistency with the other task in this file.

PR #3579, opened 2025-02-06
Skip API authentication in dev environment

pezholio:
Now we’re using the users API locally in a couple of applications, the apps are falling over when trying to access the endpoint. I looked into seeding an application user, but this is tricky because we can’t guarantee what the token will be very easily. I’ve worked around this by skipping the authentication if we’re in the development environment.

It’s not ideal, but saves a lot of messy config in the Docker containers to get apps that use the Signon API up and running.

### DIFF
--- a/app/controllers/api/users_controller.rb
+++ b/app/controllers/api/users_controller.rb
@@ -1,5 +1,5 @@
 class Api::UsersController < ApplicationController
-  before_action :doorkeeper_authorize!, :check_signon_permissions
+  before_action :authorize_api_access, unless: -> { Rails.env.development? }
   skip_after_action :verify_authorized
 
   def index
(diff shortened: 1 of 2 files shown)


mike3985 (approved):
If I've understood it correctly, this is a problem that we only experience with Signon's API and not our other APIs because those other ones use the GDS-SSO gem for authentication (which has its own way of handling working in a development environment).

It's (at best) tricky to use GDS-SSO for Signon's API because of
1. Signon's existing User model (which is in some ways at odds with the User model that GDS-SSO requires of its apps),
2. Signon's use of Devise (which is built on top of Warden, while Warden also happens to be an integral part of GDS-SSO, and both uses of Warden, again, want to be dealing with their own User model),
3. maybe other issues, too?

So, this endpoint is already a special case. We're adding to the specialness a bit here, but the result is that from the outside it'll behave like the other apps' APIs in development environments?

:+1:

pezholio:
Yeah, that sums it up nicely. It's annoying, but it's the only way I could see clear to getting this working in dev without some even uglier hacks. Thanks for the review! 👍

PR #3983, opened 2025-08-19
[PP-7302] Add decision record for anonymised user ids

richardTowers:
✨ [Rendered](https://github.com/alphagov/signon/blob/anonymised-user-id/docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md) ✨

### DIFF
--- a/docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
+++ b/docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
@@ -38,7 +38,7 @@ Digest::SHA2.hexdigest(JSON.dump([ user.uid, ENV["ANONYMOUS_USER_ID_SECRET"] ]))
 (Note: using JSON.dump to combine values before hashing them is [considered better practice than concatenation](https://jameshfisher.com/2018/01/09/how-to-hash-multiple-values/)
 because the combine function needs to be "injective" to avoid collisions like `hash("ab" + "cd") == hash("a" + "bcd")`).
 
-We chose SHA256 as the hashing algorithm rather than something like BCrypt of PBKDF2 because:
+We chose SHA256 as the hashing algorithm rather than something like BCrypt or PBKDF2 because:
 
 - SHA256 implementations are ubiquitous, including in the ruby standard library without the need for any gems or
   external dependencies such as openssl.
--- a/docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
+++ b/docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
@@ -7,7 +7,7 @@ We want to be able to reason about repeat usage of publishing applications at th
 We don't want to track full user IDs, because that would be bad for privacy.
 
 Instead, we'd like to use "anonymised" IDs, which persist for a given user, but can't easily be traced back to
-as specific person.
+a specific person.
 
 ## Problem
 
--- a/docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
+++ b/docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
@@ -24,7 +24,8 @@ ID to a named individual.
 
 Signon already generates a unique ID for each user, which is a random UUID. We will combine this with a secret known
 only to the applications, and hash the result using SHA256. We'll truncate the hash to 16 hexadecimal characters, which
-is long enough to avoid collisions while not taking up too much space.
+is long enough to avoid collisions ([source](https://github.com/alphagov/signon/pull/3983#discussion_r2284680904))
+while not taking up too much space.
 
 Calculation of the anonymous user ID will look something like this:
 
--- a/docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
+++ b/docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
@@ -1,0 +1,67 @@
+# Decision Record: Pass anonymised user IDs to publishing apps and analytics
+
+## Context
+
+We want to be able to reason about repeat usage of publishing applications at the level of individual users.
+
+We don't want to track full user IDs, because that would be bad for privacy.
+
+Instead, we'd like to use "anonymised" IDs, which persist for a given user, but can't easily be traced back to
+as specific person.
+
+## Problem
+
+The threat model here is primarily internal users, with access to analytics. We do want these people to be able
+to answer questions like "how many unique users have use Whitehall this month?", but we do not want them to be able
+to answer questions like "which named individual on my team has done the most work this month?"
+
+Anonymising the user IDs should make it reasonably hard (although not impossible) to work backwards from an anonymised
+ID to a named individual.
+
+## Solution
+
+### Anonymising the user ID
+
+Signon already generates a unique ID for each user, which is a random UUID. We will combine this with a secret known
+only to the applications, and hash the result using SHA256. We'll truncate the hash to 16 hexadecimal characters, which
+is long enough to avoid collisions while not taking up too much space.
+
+Calculation of the anonymous user ID will look something like this:
+
+```ruby
+Digest::SHA2.hexdigest(user.uid + ENV["ANONYMOUS_USER_ID_SECRET"])[..16]
+```
+
+### Providing the anonymised user ID to applications
+
+We could have Signon generate anonymous user IDs, or we could have each application generate the IDs themselves using a
+shared secret.
+
+The advantage of having Signon generate the ID is that we only need the secret to be available in one application, which
+simplifies the infrastructure and somewhat reduces the risk of the secret leaking. The disadvantages are complexity of
+implementation, and risk that an application may accidentally store the anonymous ID in the database (effectively
+de-anonymising it).
+
+On balance, we prefer to have each application generate the IDs using a shared secret - trading off some additional
+complexity in the infrastructure for simplicity of implementation in the code.
+
+### Sending the anonymised user ID to the data layer (and analytics)
+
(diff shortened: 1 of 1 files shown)


docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
```diff
@@ -0,0 +1,67 @@
...
+
+We don't want to track full user IDs, because that would be bad for privacy.
+
+Instead, we'd like to use "anonymised" IDs, which persist for a given user, but can't easily be traced back to
+as specific person.
+
+## Problem
+
+The threat model here is primarily internal users, with access to analytics. We do want these people to be able
+to answer questions like "how many unique users have use Whitehall this month?", but we do not want them to be able
```
yndajas, line 15: **typo:** "have use**d** Whitehall"

docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
```diff
@@ -0,0 +1,67 @@
...
+Anonymising the user IDs should make it reasonably hard (although not impossible) to work backwards from an anonymised
+ID to a named individual.
+
+## Solution
+
+### Anonymising the user ID
+
+Signon already generates a unique ID for each user, which is a random UUID. We will combine this with a secret known
+only to the applications, and hash the result using SHA256. We'll truncate the hash to 16 hexadecimal characters, which
+is long enough to avoid collisions while not taking up too much space.
```
yndajas, line 27: **suggestion:** I'm curious if there's a particular logic to 16 characters being long enough. It feels plenty long enough given the number of users, but is it definitively long enough for some particular reason?

docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
```diff
@@ -0,0 +1,67 @@
...
+
+Calculation of the anonymous user ID will look something like this:
+
+```ruby
+Digest::SHA2.hexdigest(user.uid + ENV["ANONYMOUS_USER_ID_SECRET"])[..16]
```
yndajas, line 32: **question:** you said something about the order of the ID and the pepper being important at some point, but I think there was a reason it didn't matter so much here? Is this still the case even with truncation?

docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
```diff
@@ -0,0 +1,67 @@
...
+
+Anonymising the user IDs should make it reasonably hard (although not impossible) to work backwards from an anonymised
+ID to a named individual.
+
+## Solution
+
+### Anonymising the user ID
+
+Signon already generates a unique ID for each user, which is a random UUID. We will combine this with a secret known
+only to the applications, and hash the result using SHA256. We'll truncate the hash to 16 hexadecimal characters, which
```
yndajas, line 26: **suggestion:** is it worth saying why we went for SHA256 over the other options, or did we decide to keep this focused?

docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+### Providing the anonymised user ID to applications
+
+We could have Signon generate anonymous user IDs, or we could have each application generate the IDs themselves using a
+shared secret.
+
+The advantage of having Signon generate the ID is that we only need the secret to be available in one application, which
+simplifies the infrastructure and somewhat reduces the risk of the secret leaking. The disadvantages are complexity of
+implementation, and risk that an application may accidentally store the anonymous ID in the database (effectively
```
yndajas, line 42: **suggestion:** is it worth explaining why the implementation would be more complex - presumably the need to pass the anonymised ID through to applications from Signon and those applications needing to store it in the session?

yndajas (commented):
Looks good - just a few minor suggestions

docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
richardTowers: I didn't think it through in much depth actually, and I should have because this is an example where the [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem) is relevant.

We have ~4,000 users. Let's be generous and round it up to 10k. What are the chances that any two of them might have the same hash? (which would be bad / annoying / an undesirable property of the system)

If the hashes were 8 characters long, that would be 16^8 == 2^32 ~= 4 billion possible hashes. Which you might think would be way more than enough to avoid the chance of collision. But actually, the chance of two of them colliding out of a pool of 10k hashes would be 0.01157 (i.e. greater than 1/100). Which is probably unacceptably high.

At 10 characters, there are 16^10 == 2^40 ~= 1 trillion possible hashes. The chance of collision out of a pool of 10k hashes would be 0.00005 (i.e. less than one in 10 thousand), which is probably unlikely enough for us.

16 characters gives plenty of headroom. The chance of collision out of a pool of 10k users would be ~2.7086e-12  (i.e. less than one in a trillion). Plenty good enough.

This wouldn't be good enough if every human being on earth had a signon account though! With 8.2 billion users, there would be a >80% chance of a hash collision.

Source: https://kevingal.com/apps/collision.html

docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
richardTowers: Yeah, the uid is a fixed length, so you can't play silly games with it to work out things about the secret.

An example of the problem of using concatenation would be:

Imagine a system where we do:

```
password_hash = SHA2(username + password)
```

(not a good idea, but roll with it).

I, an attacker, manage to find a way to read everyone's password hashes from the database.

I can see your username is `Ynda`. I want to work out what your password is. I've got a big list of common passwords, and I reckon yours will be one of them because you have rubbish op sec.

I sign up for a user `Ynd`, and then repeatedly change my password to:

* aPassword1
* aPassword2
* aPassword3

... and so on, until my password hash matches the one for `Ynda` in the database. And then boom! I know your password.

This is even more of a problem if you're worried about hash collisions, rather than preimages. But generally "concatenate then hash" is bad practice.

I googled it, and perhaps better practice (which avoids this class of issues) would be:

```
Digest::SHA2.hexdigest([user.uid, ENV["ANONYMOUS_USER_ID_SECRET"]].to_json)[..16]
```

or even:

```
Digest::SHA2.hexdigest([user.uid, ENV["ANONYMOUS_USER_ID_SECRET"]].to_s)[..16]
```

🤔 

Source: https://jameshfisher.com/2018/01/09/how-to-hash-multiple-values/

docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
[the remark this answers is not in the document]
richardTowers: Addressed in [e779858](https://github.com/alphagov/signon/pull/3983/commits/e779858e055c6a674c2a100fd0a8a4af41dbb1b6)

docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
richardTowers: Linked to this discussion in the ADR in [e9738d0](https://github.com/alphagov/signon/pull/3983/commits/e9738d0f915319b6e818cabc2497ff5ddf48c94d) ➿

docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
richardTowers: Justified the choice of SHA256 in [13c6500](https://github.com/alphagov/signon/pull/3983/commits/13c65003a6bfd59e8e9c049cd87b642ac6687dc4)

docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
```diff
@@ -0,0 +1,67 @@
...
+Digest::SHA2.hexdigest(user.uid + ENV["ANONYMOUS_USER_ID_SECRET"])[..16]
+```
+
+### Providing the anonymised user ID to applications
+
+We could have Signon generate anonymous user IDs, or we could have each application generate the IDs themselves using a
+shared secret.
+
+The advantage of having Signon generate the ID is that we only need the secret to be available in one application, which
+simplifies the infrastructure and somewhat reduces the risk of the secret leaking. The disadvantages are complexity of
```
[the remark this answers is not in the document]
richardTowers: Attempted to explain this in [ab5c695](https://github.com/alphagov/signon/pull/3983/commits/ab5c69514d52da093243f02563e55eaced57c76f)

docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md
```diff
@@ -0,0 +1,86 @@
+# Decision Record: Pass anonymised user IDs to publishing apps and analytics
+
+## Context
+
+We want to be able to reason about repeat usage of publishing applications at the level of individual users.
+
+We don't want to track full user IDs, because that would be bad for privacy.
+
+Instead, we'd like to use "anonymised" IDs, which persist for a given user, but can't easily be traced back to
+as specific person.
```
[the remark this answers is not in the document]
richardTowers: Fixed in [deaca33](https://github.com/alphagov/signon/pull/3983/commits/deaca3333598f6c629f3ecb15739647d394866b0)

PR #3992, opened 2025-08-21
[PP-7304] Add anonymised user id method and populate meta tag

richardTowers:
As per [ADR-001](https://github.com/alphagov/signon/blob/main/docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md).

I had to mess about with the google_tag_manager partial because only one thing is allowed to call `content_for :head`, and now we need to put another bit of content in the head.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,4 +1,10 @@
-<%= render "layouts/google_tag_manager" %>
+<% content_for :head do %>
+  <%= render "layouts/google_tag_manager" if ENV["GOOGLE_TAG_MANAGER_ID"] %>
+
+  <% if current_user&.anonymous_user_id %>
+    <%= tag.meta name: "govuk:user-id", content: current_user.anonymous_user_id  %>
+  <% end %>
+<% end %>
 
 <%= render "govuk_publishing_components/components/layout_for_admin",
   product_name: "Signon",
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -418,6 +418,13 @@ def web_user?
     !api_user?
   end
 
+  def anonymous_user_id
+    secret = ENV["ANONYMOUS_USER_ID_SECRET"]
+    return if secret.nil?
+
+    @anonymous_user_id ||= Digest::SHA2.hexdigest([self.uid, secret].to_json)[..16]
```
yndajas, line 425: **thought:** I suppose you could reorganise this to check if the instance variable exists on the first line instead of the last, but it probably wouldn't make much difference performance-wise, so this seems fine

app/views/layouts/application.html.erb
```diff
@@ -1,4 +1,10 @@
-<%= render "layouts/google_tag_manager" %>
+<% content_for :head do %>
+  <%= render "layouts/google_tag_manager" if ENV["GOOGLE_TAG_MANAGER_ID"] %>
+
+  <% if current_user && current_user.anonymous_user_id.present? %>
```
yndajas, line 4: **suggestion:** you could `current_user&.anonymous_user_id.present?` here

yndajas (approved):
Smells good to me

app/models/user.rb
richardTowers: Yeah, I considered it, but I couldn't do it without adding like one more line to the implementation and I didn't like the look of the code. I don't love it as it is either, but 🤷🏻

PR #4006, opened 2025-08-22
[PP-7304] Fix analytics

richardTowers:
We were still including the old universal analytics snippet, which hasn't worked since google retired UA around 1 year ago. All of the analytics tests were checking this old defunct implementation.

Meanwhile, a change was made in our google tag manager property which prevented page view events from firing in the new GA4 implementation, unless the [load-analytics](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/load-analytics.js) script was included (which we weren't doing). This affects several other apps (signon, content data, travel advice, manuals, contacts admin, search admin, content publisher), which will be addressed separately.

While cleaning up the tests for the old universal analytics implementation, I spotted a couple of important looking tests about sanitising certain query parameters (e.g. password reset tokens). It looks like some of these are ignored by our analytics scripts anyway, but to make sure we cover them all, I've added [a meta tag which lets the script know which parameters to skip](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/pii-remover.md#for-query-string-parameters). I've tested the presence of this meta tag, but I'm trusting that publishing components already tests the JS functionality well enough, and we don't need tests in every application that uses it.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,9 +1,14 @@
 //= require govuk_publishing_components/dependencies
+//= require govuk_publishing_components/lib/cookie-functions
 //= require govuk_publishing_components/components/copy-to-clipboard
 //= require govuk_publishing_components/components/option-select
 //= require govuk_publishing_components/components/password-input
 //= require govuk_publishing_components/components/table
 
+//= require ./domain-config
 //= require_tree ./modules
 //= require_tree ./components
 //= require rails-ujs
+
+window.GOVUK.approveAllCookieTypes()
+window.GOVUK.cookie('cookies_preferences_set', 'true', { days: 365 })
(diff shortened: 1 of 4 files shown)


patrickpatrickpatrick (approved):

app/views/layouts/_analytics_meta_tags.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= tag.meta name: "govuk:components_gem_version", content: GovukPublishingComponents::VERSION %>
+
+<% if user_signed_in? %>
```
brucebolt, line 4: Non-blocking comment: would it be better to add `&& current_user` to this conditional rather than using the safe navigation operator on each tag below?

brucebolt (approved):

PR #4565, opened 2026-04-14
Fix URI substitution bug

callumknights:
Fixes a bug where the URI sub would repeat even if the original URI matched the pattern. Sometimes resulting in
staging.staging.publishing.service.gov.uk etc


brucebolt (commented):
The code change looks fine, but could you also update the tests to include this scenario.

brucebolt (commented):
The code change and tests look good to me.

But I think the same test will need to be added [for `redirect_uri`](https://github.com/alphagov/signon/blob/d63f1d75d78fd4de184ab31c3e2400b786667ad7/test/models/doorkeeper/application_test.rb#L27-L51), which appears to have almost identical tests and functionality.

brucebolt (approved):
Thanks for making the change, all looks good to me.

PR #4592, opened 2026-04-22
Fix flaky test

nacnudus:
This allows a particular call in a particular test to wait longer for an element to be rendered. Otherwise we occasionally see the following error.

```
EmailChangeTest#test_: by a user themselves should be cancellable. :
Capybara::ElementNotFound: Unable to find field "Email" that is not disabled
    test/integration/email_change_test.rb:163:in 'block (2 levels) in <class:EmailChangeTest>'
    test/integration/email_change_test.rb:164:in 'BasicObject#instance_exec'
    test/integration/email_change_test.rb:164:in 'block in EmailChangeTest#create_test_from_should_hash'
bin/rails test test/integration/email_change_test.rb:156
```

If similar problems occur elsewhere, then Capybara's `default_max_wait_time` (default: 2) could be configured globally.

A gem version also had to be tweaked, as described in the commit message.

### DIFF
--- a/test/integration/email_change_test.rb
+++ b/test/integration/email_change_test.rb
@@ -160,7 +160,7 @@ class EmailChangeTest < ActionDispatch::IntegrationTest
       signin_with(@user)
 
       click_link "Change your email"
-      fill_in "Email", with: "new@email.com"
+      fill_in "Email", with: "new@email.com", wait: 3
       click_button "Change email"
 
       confirmation_token = token_sent_to(@user)
--- a/Gemfile
+++ b/Gemfile
@@ -62,7 +62,7 @@ group :test do
   gem "minitest"
   gem "mocha", require: false
   gem "rails-controller-testing"
-  gem "shoulda-context", require: false
+  gem "shoulda-context", "~> 3.0.0.rc1", require: false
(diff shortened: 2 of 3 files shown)


test/integration/email_change_test.rb
```diff
@@ -160,7 +160,7 @@ class EmailChangeTest < ActionDispatch::IntegrationTest
       signin_with(@user)
 
       click_link "Change your email"
-      fill_in "Email", with: "new@email.com"
+      fill_in "Email", with: "new@email.com", wait: 3
```
yndajas, line 163: **suggestion:** I'd maybe note in the commit message that the test is failing _randomly_ (not consistently), even when using a consistent seed, and that we (you) can replicate it locally (not just in CI; also randomly)

yndajas: There's a 'can' missing in the new message. Perhaps, for extra clarity on the randomness:

> A test fails sometimes (at random) in CI because an element hasn't yet been rendered. This ~random~ failure _can_ be replicated locally, _but it remains random_ even when using a _specific_ seed.

brucebolt (approved):

PR #4601, opened 2026-04-29
Add workflow to build Docker image

brucebolt:
This calls the [build-image-from-PR](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/build-image-from-pr.yml) workflow which builds the image from the pull request ensuring that an image can successfully be built the PR's branch.

### DIFF
--- a/.github/workflows/build-image-from-pr.yml
+++ b/.github/workflows/build-image-from-pr.yml
@@ -1,0 +1,20 @@
+name: Build image from PR
+
+on:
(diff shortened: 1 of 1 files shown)


AgaDufrat (approved):
LGTM. 
This is the link to the reusable workflow (as the one in PR description seems to be broken): https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/build-image-from-pr.yml

PR #4637, opened 2026-06-03
[PP-7756] Resolve email address XSS vulnerability

brucebolt:
When a user changes their email address, we store the new email in the `unconfirmed_email` field until they have verified the new email.

However this field does not have the same validations as the `email` field.

This has user impact, because it means that users could enter an invalid email address, but only find out once they attempt to verify it. This wastes their time, as they have to wait for an email and click the link, only to find the email is invalid.

However, perhaps more importantly, it creates a XSS vulnerability, where a compromised user account could attempt to change their email address to something like ascript. This would then execute in the browser of anyone who views that
user's account access log.

By restricting the value to only valid email addresses, we minimise the risk of this occurring. This PR has sanitises the value when rendered from the event log, in the case any malicious entries slip past the validation.

### DIFF
--- a/app/helpers/event_log_helper.rb
+++ b/app/helpers/event_log_helper.rb
@@ -31,7 +31,7 @@ module EventLogHelper
   end
 
   def formatted_trailing_message(log)
-    log.trailing_message
+    sanitize(log.trailing_message)
   end
 
   def formatted_ip_address(log)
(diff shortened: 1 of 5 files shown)


callumknights (approved):
LGTM. Good thinking with the event log change.

PR #4638, opened 2026-06-08
Update the GTM container attributes [WHIT-3365]

patrickpatrickpatrick:
## What

Update the GTM container ID

## Why

Request from PAs


AgaDufrat (approved):
The change looks good to me. It maybe good to include a bit more context on this GTM instance in the commit message. Is it Signon specific now? I belive it was shared with other apps before.

patrickpatrickpatrick:
> The change looks good to me. It maybe good to include a bit more context on this GTM instance in the commit message. Is it Signon specific now? I belive it was shared with other apps before.

It uses the more approach of other applications that have moved to the Design System, which specify the GTM Container ID in `domain-config.js`. This changed when support was added for passing a user ID in the analytics it seems.

PR #4647, opened 2026-06-12
Optimise breadcrumbs component calls

andysellick:

## What
Reduce the number of calls to the breadcrumb component by calling it once in the layout and passing that instance the various breadcrumbs from each view, rather than having an instance of the breadcrumb component per view.

## Why
There are 46 views where the breadcrumb component is called, and they're all exactly the same apart from the actual breadcrumbs that get passed to the component.

## Visual changes
None.

### DIFF
--- a/app/views/account/activities/show.html.erb
+++ b/app/views/account/activities/show.html.erb
@@ -1,16 +1,5 @@
 <% content_for :title, "Account access log" %>
 
-<%
-  @breadcrumbs = [
-    {
-      title: "Dashboard",
-      url: root_path,
-    },
-    {
-      title: "Settings",
-      url: account_path,
-    }
-  ]
-%>
+<% @breadcrumbs = account_breadcrumbs %>
 
 <%= render "shared/event_logs_table", logs: @logs %>
--- a/app/views/account/activities/show.html.erb
+++ b/app/views/account/activities/show.html.erb
@@ -1,19 +1,16 @@
 <% content_for :title, "Account access log" %>
 
-<% content_for :breadcrumbs,
-   render("govuk_publishing_components/components/breadcrumbs", {
-     collapse_on_mobile: true,
-     breadcrumbs: [
-       {
-         title: "Dashboard",
-         url: root_path,
-       },
-       {
-         title: "Settings",
-         url: account_path,
-       }
-     ]
-   })
+<%
+  @breadcrumbs = [
+    {
+      title: "Dashboard",
+      url: root_path,
+    },
+    {
+      title: "Settings",
+      url: account_path,
+    }
+  ]
 %>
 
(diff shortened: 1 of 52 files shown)


KludgeKML (approved):
LGTM

app/views/account/activities/show.html.erb
```diff
@@ -1,19 +1,16 @@
 <% content_for :title, "Account access log" %>
 
-<% content_for :breadcrumbs,
```
yndajas, line 3: nitpick: technically there is no `content_for` anymore

> and pass only the breadcrumb contents in content_for

test/integration/applications/access_logs_test.rb
```diff
@@ -24,6 +24,8 @@ class ApplicationsAccessLogsIntegrationTest < ActionDispatch::IntegrationTest
     should "have permission to view account access log" do
       visit access_logs_doorkeeper_application_path(@application)
       assert_equal page.title, "app-name access log - GOV.UK Signon"
+      assert has_link?("Dashboard", href: "/", class: "govuk-breadcrumbs__link")
+      assert has_link?("Applications", href: "/doorkeeper_applications", class: "govuk-breadcrumbs__link")
```
yndajas, line 28: though (non-blocking): I'm a little unsure about tagging these assertions onto the end of permissions tests. I prefer the previous set of more focused breadcrumbs tests

test/integration/users/account_access_logs_test.rb
```diff
@@ -121,5 +121,7 @@ def assert_account_access_log_page_content(user)
     assert assert_text "Event"
     assert assert_text "account locked"
     assert_selector "a", text: user.name
+    assert has_link?("Dashboard", href: "/", class: "govuk-breadcrumbs__link")
+    assert has_link?("Users", href: "/users", class: "govuk-breadcrumbs__link")
```
yndajas, line 125: ... although this spec is already in the pattern of making multiple assertions about page content in tests whose titles focus on permissions

yndajas (approved):
Thanks - much improved!

app/views/account/activities/show.html.erb
andysellick: Reworded, thanks 👍

test/integration/users/account_access_logs_test.rb
andysellick: Yes, I was in two minds about this as well. To be honest the coverage of the tests that I've added isn't as comprehensive as I'd like, mainly because I'm unfamiliar with minitest and signon, but having some tests in place felt like an improvement and the approach I chose felt reasonable enough without being too disruptive and stepping outside of my boundaries. I'm more than happy for someone more familiar with this app to step in and improve this at a later date.

PR #4666, opened 2026-06-30
Upgrade `standardx` to address a security vulnerability

brucebolt:
This upgrades `js-yaml` to version 3.15.0, which includes a security vulnerability fix.

Generated with `yarn upgrade standardx`.

Resolves https://github.com/alphagov/signon/security/dependabot/200 and https://github.com/alphagov/signon/security/dependabot/199.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -722,11 +722,12 @@ es-shim-unscopables@^1.0.2:
     hasown "^2.0.2"
 
 es-to-primitive@^1.3.0:
-  version "1.3.1"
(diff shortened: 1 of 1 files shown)


yndajas (approved):
Would it be clearer to say "Upgrade `standardx` dependencies"? While technically the command run was to upgrade standardx, its version didn't change, so the commit title might be a little confusing

yndajas:
Updated and merged in Bruce's absence

PR #4707, opened 2026-08-11
Bump js-yaml packages

callumknights:
Resolves two security alerts: https://github.com/alphagov/signon/security/dependabot/217 https://github.com/alphagov/signon/security/dependabot/216


package.json
```diff
@@ -46,6 +46,7 @@
     "stylelint/string-width": "4.2.3"
   },
   "dependencies": {
-    "accessible-autocomplete": "^3.0.1"
+    "accessible-autocomplete": "^3.0.1",
+    "js-yaml": "^4.1.0"
```
yndajas: **question:** do we need to add this as an explicit dependency to bump the lockfile? I think we should be able to bump it without, thereby retaining its transient status

callumknights: Oh that might have slipped in from previous attempts at me trying to make sense of classic yarn. Let me edit that.

callumknights: Much better

yndajas (approved):

PR #4712, opened 2026-08-24
Bump govuk_publishing_components to v68.3.0

geek4good:
- Update govuk_publishing_components to v68.3.0, because it contains govuk-frontend v6 (latest major version of the design system)
- Update calls to govuk-colour() to get rid of deprecation warnings

**Why a draft PR?**

I haven't optimised the `import` and `require` statements yet, because I couldn't get the component guide to work in signon. It would just end up displaying the following error message:

LoadError in GovukPublishingComponents::ComponentGuideController#index
cannot load such file -- sassc
```ruby
          Kernel.warn message, uplevel: ::Gem::BUNDLED_GEMS.uplevel
        end
        kernel_class.send(:no_warning_require, name)
      end
      if kernel_class == ::Kernel
        kernel_class.send(:private, :require)
```
Rails.root: /govuk/signon

`/root/.rbenv/versions/4.0.6/lib/ruby/4.0.0/bundled_gems.rb:60:in 'Kernel.require'` (first line of the stack trace)

Apparently, the error is happening during the `send` call.

### DIFF
--- a/app/assets/javascripts/es6-components.js
+++ b/app/assets/javascripts/es6-components.js
@@ -10,6 +10,5 @@
 //= require govuk_publishing_components/components/button
 //= require govuk_publishing_components/components/checkboxes
 //= require govuk_publishing_components/components/error-summary
-//= require govuk_publishing_components/components/layout-header
 //= require govuk_publishing_components/components/skip-link
 //= require govuk_publishing_components/components/tabs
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/es6-components.js
```diff
@@ -10,6 +10,5 @@
 //= require govuk_publishing_components/components/button
 //= require govuk_publishing_components/components/checkboxes
 //= require govuk_publishing_components/components/error-summary
-//= require govuk_publishing_components/components/layout-header
```
yndajas: **question:** I'm guessing this is unused?

geek4good:
Accidentally used the wrong ticket number in the branch name.

PR #4719, opened 2026-08-28
Require ES6 components through layout_for_admin

geek4good:
The `layout_for_admin` helper allows to easily separate regular and ES6-compatible JS. But for the ES6 file to be included in the HTML it outputs it needs to be enabled via `use_es6_components = true` (see this [link](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/install-and-use.md)).

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -99,6 +99,4 @@
       }
     ]
   } %>
-
-  <%= javascript_include_tag "es6-components", type: "module" %>
 <% end %>
--- a/config/initializers/govuk_publishing_components.rb
+++ b/config/initializers/govuk_publishing_components.rb
@@ -1,0 +1,3 @@
+GovukPublishingComponents.configure do |config|
+  config.use_es6_components = true
+end


callumknights (dismissed):

app/views/layouts/application.html.erb
```diff
@@ -99,6 +99,4 @@
       }
     ]
   } %>
-
-  <%= javascript_include_tag "es6-components", type: "module" %>
```
yndajas, line 103: **question:** I can't see that we're setting `GovukPublishingComponents::Config.use_es6_components` to `true` in this codebase, and it defaults to `false`, so do we need an accompanying config change?

Related: https://github.com/alphagov/govspeak-preview/pull/936#discussion_r3873614597

geek4good: I've added the initialiser. Thanks for pointing this out.

yndajas: I wonder if the commit message needs updating, since it turns out it wasn't redundant before adding the intialiser

geek4good: Your attention to detail is impressive. Commit message rewritten.

yndajas (approved):

PR #4722, opened 2026-09-07
[PP-7717] Pact v2 migration

geek4good:
Migrate pact provider verification to pact v2 ([Jira ticket](https://gov-uk.atlassian.net/browse/PP-7717))

Upgrade pact from v1.x to v2.x and rewrite the provider verification setup for the new Rust-core based API.

- include Pact tests in the CI run
- bump pact to ~> 2.0 and add pact-ffi (v2 verification runs on the Pact Rust core via FFI) and the rspec meta-gem (pact 2 no longer declares it as a runtime dependency)
- replace the v1 DSL in spec/service_consumers/pact_helper.rb with the v2 RSpec DSL in spec/pact/consumers/gds_api_adapters_spec.rb
- move per-interaction setup into Pact.configuration global hooks so auth stubbing and database cleaning run for every interaction
- define the `pact:verify` task explicitly in the Rakefile (pact v2 no longer ships lib/pact/tasks.rb)
- mark `pact_broker-client` as `require: false` (only used by rake task)


callumknights (dismissed):

spec/support/pact/database_cleaner.rb
```diff
@@ -0,0 +1,9 @@
+require "database_cleaner/active_record"
+
+if ENV["DATABASE_URL"]
+  DatabaseCleaner.url_allowlist = [
+    %r{\Amysql2://root:root@mysql-8/signon_(?:development|test)\z},
+  ]
+end
```
callumknights, line 7: Commenting this for posterity so other readers can understand the reason for this. Signon is the only app where this config is not set in docker https://github.com/alphagov/govuk-docker/blob/main/projects/link-checker-api/docker-compose.yml#L26, therefore this specific app requires this allowlist change, and is probably more secure than having the blanket allow for remote_urls like the other apps

callumknights: Given the discussion we had in dev chat, should we perhaps change approach here and actually change the config in govuk-docker to match the other 12 apps that allow remote database cleaner?

geek4good: I think the conclusion of our discussion was that the other repos are okay the way they are, because it's virtually impossible for anybody to accidentally use production database credentials and wipe everything. Apart from that, this repo / app is the only one that's perfectly safe due to the code above.

callumknights: Ahh, I obviously didn't communicate this well enough. I wondered if given the risks of that config are mitigated separately, maybe it would be better to remove this code, and modify the project in `govuk-docker`, just for parity with other apps?

test/service_consumers/pact_helper.rb
```diff
@@ -45,6 +45,11 @@ def stub_access_token_creation!
 
 Pact.provider_states_for "GDS API Adapters" do
   set_up do
+    if ENV["DATABASE_URL"]
+      DatabaseCleaner.url_allowlist = [
+        %r{\Amysql2://root:root@mysql-8/signon_(?:development|test)\z},
```
yndajas, line 50: **question:** would we be alerted if we upgrade to MySQL 9 (or 26), or would this fail silently?

I wonder if we could grab the values from the environment variables set by govuk-docker instead (and just include them as two array entries): https://github.com/alphagov/govuk-docker/blob/42a0958320172aebf77827d2bc2cef698a3add96/projects/signon/docker-compose.yml#L27-L28

spec/pact/consumers/gds_api_adapters_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    config.after_provider_state_teardown do
+      DatabaseCleaner.clean
+    end
+  end
+
+  http_pact_provider "Signon API", opts: {
+    http_port: 9292,
+    pact_uri: ENV["PACT_URI"],
```
yndajas, line 22: **question:** as per https://github.com/alphagov/publishing-api/pull/4193#discussion_r4047190639, do we need to account for ENV["PACT_URI"] potentially being nil?

yndajas (commented):
Nothing major again. Generally makes sense

spec/pact/consumers/gds_api_adapters_spec.rb
geek4good: If `nil` it falls back to the pact broker. This is expected behaviour.

test/service_consumers/pact_helper.rb
geek4good: I've implemented your suggested solution. Wasn't too complicated, and I quite like it.

Gemfile
```diff
@@ -62,8 +62,10 @@ group :test do
   gem "minitest"
   gem "mocha", require: false
   gem "rails-controller-testing"
+  gem "selenium-webdriver", "~> 4.43.0" # newer versions have dropped Chrome 128
   gem "shoulda-context", "~> 3.0.0.rc1", require: false
   gem "simplecov"
   gem "timecop"
   gem "webmock"
+  gem "zxcvbn-ruby", "~> 1.4" # v2 has updated its pasword strength expectation
```
geek4good, line 70: Pinning this and selenium-webdriver might warrant follow-up tasks to properly update both. After resolving some merge conflicts, I ended up with selenium-webdriver v4.49 and zxcvbn-ruby v2.0.1, which both led to tests failing (see https://github.com/alphagov/signon/actions/runs/35839957177).

test/service_consumers/pact_helper.rb
yndajas: TIL about the second argument to `split`

yndajas: So it looks like:
- the test directory helper uses the old approach in commit two
- the spec directory helper uses the old approach in commit four
- the test directory helper is removed in commit five
- the spec directory is updated to the new approach in commit six

Is there a reason for now using the new approach straight away for the respective files in commits two and four?

Gemfile.lock
```diff
@@ -874,6 +890,308 @@ DEPENDENCIES
   timecop
   uuid
   webmock
+  zxcvbn-ruby (~> 1.4)
+
+CHECKSUMS
```
yndajas: **question:** what command did you run when replacing `pact_broker-client` with `pact-ffi`? I've tried:
- resetting the lockfile
- removing the `selenium-webdriver` and `zxcvbn-ruby` additions from the Gemfile
- running `govuk-docker-run bundle install`

I end up with a much smaller diff, not touching those dependencies that you needed to pin, nor adding the checksums section or changing the platforms

yndajas:
Other than [comment A](https://github.com/alphagov/signon/pull/4722/#discussion_r4084967593) and [comment B](https://github.com/alphagov/signon/pull/4722/#discussion_r4085068919), I think this is good to go (labelling them so GitHub doesn't override the links with just the PR!)

test/service_consumers/pact_helper.rb
geek4good: I'm guessing you meant "not" instead of "now", in which case the answer to your question would be something along the lines of convenience / laziness. ;)

As per usual, your assessment is correct, and I've now gone back and changed those earlier commits to also ensure DatabaseCleaner's allow list pulls in the database URL from the docker compose config.

Gemfile.lock
geek4good: Because the prospect of having to sort out a merge conflict in Gemfile.lock is horrifying to me, I've run `rm Gemfile.lock && bundle install`, which has led to more stuff getting updated.

Your way of doing it was a lot less disruptive. Went back and have undone my unnecessary changes.

Gemfile
geek4good: This is now gone again.

